PR #11, opened 2026-08-07
perf(consensus): fix mainnet script verification and make apply 2.89x faster

metaphorics:
## Summary

Mainnet full-validation sync no longer wedges, and processing-bound apply is **2.89× faster** than at the branch point. Two consensus defects blocked real mainnet chains before this branch: taproot key-path spends were verified without the full prevout set, and the default verifier could not validate the script classes mainnet actually contains. Both are fixed, `bitcoinkernel` is now the default verifier, and mainnet nodes ship a hash-pinned assume-valid anchor so historical script checks are skipped only against a verified chain.

The performance work is measured, not asserted: every candidate was benchmarked on a pinned 3× median, and the five that did not clear a 1.05× noise floor were reverted rather than kept.

## Correctness

| Fix | Before | After |
|---|---|---|
| `fix(script): pass all prevouts to taproot key spends` | key-path sighash computed against an incomplete prevout set | full prevout set passed; taproot key spends validate correctly |
| `fix(consensus): make bitcoinkernel the default verifier` | default builds could not verify every mainnet script class | `libbitcoinkernel` (Core's own engine) is the default; the Rust interpreter remains for differential testing |
| `feat(node): hash-pinned assume-valid default on mainnet` | assume-valid trusted a height with no chain binding | height 938343 is skipped **only** after the active header chain validates the pinned anchor hash; sub-anchor and diverged chains fall back to full verification |
| `fix(node): send getheaders after getdata within a tick` | an earlier commit moved the header request to the top of `tick`, inverting message order and breaking 43 sync tests | order restored while keeping best-ready-peer selection and the post-drain locator |
| `fix(index): dedup block rows before counting them` | write batching counted rows *generated*, not rows *written* (block 481824: 4113 vs 3740 funding) | dedup runs before counting, as it did pre-batching |
| `fix(bench): model the Electrum node as a single process` | the fixture served Electrum from the test process while naming a different `--pid`, so it could not satisfy the script's new tip and socket-ownership checks | one process under argv0 `bitcoin-rs` serves Electrum and owns the socket |

## Performance

Processing-bound replay, mainnet 0→150,000, `assume_valid_height=0` (full script verification), fjall, `taskset -c 0-31`, 3× medians:

| Stage | Elapsed | blk/s | Change |
|---|---|---|---|
| branch point (`4700c25`) | 389.7s | 385 | baseline |
| shared prevout resolution + txid parallelism | 173.1s | 867 | 2.25× |
| script-verify pool 16 → 32 threads | 157.8s | 950 | 2.47× |
| parallel-check threshold 16 → 4 | 135.0s | 1111 | 2.89× |
| serial UTXO prevout resolve | **~127s** | — | **~3.0× cumulative** |

Supporting wins: deferred redb body-index durability, active-prefix ancestry and locator indexes in `chain`, MTP heap-allocation removal, batched txindex writes across `apply_buffered_blocks`, and inbound block byte reuse in `sync`.

Bitcoin Core 31.0 runs the same window in 67s, so **bitcoin-rs remains ~2.01× slower** in this regime. That gap is documented rather than papered over, including the caveat that ~22s of the run is REST fetch the Core harness never pays, which puts the defensible apply-only ratio nearer 1.7×.

## Design decisions

- **Reverts are part of the record.** Parallel `prepare_block_input_checks` (1.00×), a thread-local serialize buffer (1.00×), per-block profiling histograms (+23s overhead), and a 64-thread pool (177.9s, worse than 16) were each measured and backed out. The branch keeps the measurements and drops the code.
- **Pool width has an optimum, not a monotonic curve.** 16 → 173.1s, 32 → 157.8s, 64 → 177.9s. The constant stays a cap rather than `available_parallelism` so verification cannot starve the rest of the apply pipeline.
- **Width is useless to a block that never reaches the pool.** The bigger win was granularity: `MIN_PARALLEL_SCRIPT_CHECKS` routed every block under 16 input checks to the serial branch, which on 0→150k is most of them. Lowering it to 4 bought 1.15× on its own. That threshold sweep is also an interior minimum (2 → 143.8s, 4 → 139.4s, 8 → 145.1s), so it is pinned by measurement in both directions.
- **Attribution before optimization.** Whether `script_parallel` was real secp256k1 work or rayon overhead was settled by disabling the stage (all-serial: 313.3s) rather than guessing — which is what identified pool width as the binding constraint.
- **Assume-valid is hash-gated, not height-gated.** A height alone would let a diverged chain inherit the skip; requiring the anchor hash in the active chain keeps the fast path tied to the chain it was measured on.
- **Benchmark harness hardened alongside the code.** G14 evidence scripts now bind elapsed seconds to exact Criterion sections and attest the Core process before networking, so a mismatched or unlabeled timing cannot silently become an artifact.

## Test plan

- **`cargo test --workspace --release --no-fail-fast` — 0 failures.** This is the gate that matters and it was not run earlier in the branch's life: subset runs hid 47 real failures (43 sync, 3 Electrum evidence, 1 index), each fixed above and each verified against `origin/main` before being called a regression.
- `cargo check --workspace --all-targets` — clean, both with and without the `kernel` feature.
- `cargo test -p bitcoin-rs-node --lib --release` — 341 pass (was 298 passed / 43 failed).
- `cargo test -p bitcoin-rs-consensus --release` — 60 lib + 5 vector + 3 kernel tests pass, including `script_verdict_parity` and the taproot script-path vectors.
- `cargo test -p bitcoin-rs-node --lib apply::consensus_rule_tests --release` — 72 pass, covering BIP30/BIP34, DAA retarget boundaries, BIP68, coinbase maturity, and the assume-valid skip/no-skip paths.
- `crates/storage/tests/backend_equivalence.rs` extended so the redb durability change is checked against the other backends rather than in isolation.
- Performance claims come from `mainnet_prefix_replay` against a local `bitcoind -rest`, 3× medians on pinned CPUs; the JSON artifacts are referenced from the solution note.

Durable findings are recorded under `docs/solutions/` (script-verify pool sweep, redb durability boundary, at-scale benchmark fidelity) and the assume-valid vocabulary is reconciled in `CONCEPTS.md`, per `AGENTS.md`.


metaphorics:
@coderabbitai ultrareview

PR #49, opened 2026-08-11
rpc: Core-compatible REST gateway so a remote bip300301_enforcer can use bitcoin-rs as its mainchain node

metaphorics:
## Summary

Adds the minimal Bitcoin Core-compatible REST surface an unmodified [`bip300301_enforcer`](https://github.com/LayerTwo-Labs/bip300301_enforcer) consumes, so it can be pointed at bitcoin-rs instead of bitcoind. No BIP300/301 consensus logic is ported.

The enforcer's REST client dictates the shape: it builds its base URL as `http://{--node-rpc-addr}` (`app/main.rs`), sends no `Authorization` header, and only calls `GET /rest/chaininfo.json` and `GET /rest/headers/{hash}.json?count=N` (`lib/validator/main_rest_client.rs`); it reads **any** 404 on `/rest/*` as "REST not enabled". So REST is served on the *existing* JSON-RPC listener/port, unauthenticated, off by default behind Core's `rest=1` — not a second listener — and 404 is reserved for exactly the cases that detection is meant to catch (gateway disabled, unknown route, missing output format).

Endpoints: `chaininfo.json` (reuses `getblockchaininfo`), `headers/{hash}.{json,hex,bin}`. Status semantics follow Core: `count` defaults to 5 and must be in `1..=2000` (out of range → 400, no silent clamping), unknown query parameters and valueless keys are ignored, a well-formed hash that is not on the active chain → `200` with an empty body, a bad hash or unsupported extension → `400 Invalid hash: <hash>` (hash only, no extension), a missing extension → `404 output format not found`. Header JSON fields match the enforcer's deserializer, including `bits` as unprefixed hex so `CompactTarget::from_unprefixed_hex` round-trips. `/rest/block/*`, `/rest/tx/*`, `/rest/mempool/*`, `/rest/blockhashbyheight/*` are deliberately not implemented — the enforcer gets raw blocks via `getblock` verbosity 0 over JSON-RPC.

Server change is confined to `serve_connection`, which previously hard-rejected any non-`POST` request line:

```
request = read_request(..)                  // now parses method + path; GET may omit Content-Length,
if request.method == "GET" {                // POST still errors without it (unchanged)
    response = if path.starts_with("/rest/") { rest::route(ctx, path, query, rest_enabled) }
               else { 404 }                 // never falls through into the auth/JSON-RPC path
    write_response(..); continue            // keep-alive preserved, no auth check
}
if !auth.validate_header(..) { 401 }        // POST path byte-identical to before
```

Header traversal mirrors Core's `rest_headers`, whose loop condition is `ActiveChain().Contains(pindex)`: the requested hash must be *positively* on the applied chain, otherwise the result is the empty `200`. That covers side branches, orphans, and headers already known to the tree but not yet applied — during header-first sync a header above the applied tip yields `[]`, and a walk that reaches the applied tip stops there rather than continuing up the header chain. One `block_tree` read snapshot per request serves the whole walk via parent pointers (resolve the hash once, resolve the terminal height once, walk ancestors, reverse), so a large `count` no longer costs a full `Context::blocks` scan per header — `count=2000` on a 6100-block chain went from seconds to ~5 ms — and the walk cannot splice records from two branches mid-reorg. The linkage check is kept as a defense against an internally inconsistent tree and truncates the tail at the first broken link.

Two known gaps for enforcer operators, documented in `docs/rest-interface.md` rather than papered over: `getnetworkinfo` reports version `10000` (major 1) which the enforcer's version gate rejects, so it must run with `--bitcoin-core-skip-version-check`; and bitcoin-rs publishes no ZMQ `pubsequence` topic, so the enforcer's mempool sync needs an explicit `--node-zmq-addr-sequence` or a no-mempool/bounded mode. `pubsequence` can follow separately.

Coverage: unit tests for routing/count boundaries/error precedence (`count` is validated before the hash, as Core does)/active-chain rules/JSON field shapes, socket-level tests in `crates/rpc/tests/auth.rs` (REST 200 without credentials, 404 when disabled, POST JSON-RPC still 401/200 with and without auth, non-`/rest/` GET → 404, keep-alive serving two REST requests on one connection), plus two rounds of runtime verification against a live regtest node peered to Bitcoin Core 27.0 (see PR comments) which is what surfaced the `BITCOIN_RS_REST=1` env-parsing crash, the header-cache scan cost, the applied-vs-header tip boundary, and the Core status/message divergences fixed here.

Requested by: @metaphorics

### DIFF
--- a/CONCEPTS.md
+++ b/CONCEPTS.md
@@ -2,6 +2,13 @@
 
 Shared domain vocabulary for this project: entities, named processes, and status concepts with project-specific meaning. Seeded with core domain vocabulary, then accretes as ce-compound processes learnings; direct edits are fine. Glossary only, not a spec or catch-all.
 
+## Node interfaces
+
+### REST gateway
+The optional, unauthenticated Bitcoin Core-compatible HTTP surface served on
+the existing JSON-RPC listener. It is enabled with `rest=1`; JSON-RPC requests
+on the same listener retain their configured authentication.
+
 ## Initial Block Download
 
 ### Initial Block Download (IBD)
--- a/crates/node/src/bitcoin_conf_compat.rs
+++ b/crates/node/src/bitcoin_conf_compat.rs
@@ -54,6 +54,7 @@ fn apply_key(layer: &mut ConfigLayer, key: &str, value: &str) {
         "rpcuser" => layer.rpc_user = Some(value.to_owned()),
         "rpcpassword" => layer.rpc_password = Some(value.to_owned()),
         "rpccookiefile" => layer.rpc_cookie = Some(value.into()),
+        "rest" => layer.rest = parse_core_bool(value),
         "listen" if parse_core_bool(value).is_some_and(|listen| !listen) => {
             layer.p2p_listen = Some(Vec::new());
         }
@@ -138,6 +139,9 @@ impl ConfigLayerMerge for ConfigLayer {
         }
         if other.rpc_bind.is_some() {
             self.rpc_bind = other.rpc_bind;
+        }
+        if other.rest.is_some() {
+            self.rest = other.rest;
         }
         if other.rpc_auth.is_some() {
             self.rpc_auth.clone_from(&other.rpc_auth);
--- a/crates/node/src/config.rs
+++ b/crates/node/src/config.rs
@@ -431,6 +431,7 @@ impl Config {
         Ok(config)
     }
 
+    #[allow(clippy::too_many_lines)]
     fn apply_layer(&mut self, layer: &ConfigLayer) {
         if let Some(network) = layer.network {
             self.network = network;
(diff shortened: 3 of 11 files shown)


metaphorics:
@codex review

metaphorics:
@devin fix that as another pr

metaphorics:
@codex review

metaphorics:
@coderabbitai ultrareview

metaphorics:
@devin Should support ZMQ too. See https://github.com/LayerTwo-Labs/bip300301_enforcer and get full compatibility. Do your best.

metaphorics:
@devin Be modular. Don't naively copy bitcoin-core's full surface.

metaphorics:
@codex review

PR #50, opened 2026-08-11
rpc: report real difficulty, time and mediantime from getblockchaininfo

metaphorics:
## Summary

`getblockchaininfo` reported three values that were wrong rather than merely imprecise. Diffing the handler's output against Bitcoin Core 27.0 on a regtest chain of 2107 blocks:

| field | bitcoin-rs (before) | Core 27.0 |
| --- | --- | --- |
| `difficulty` | `1.0` | `4.656542373906925e-10` |
| `time` | `0` | tip header timestamp |
| `mediantime` | `0` | tip median-time-past |

`time`/`mediantime` were hardcoded zeros. `difficulty` was computed as a ratio against *the network's own* PoW limit, which makes every chain report `1.0` at its easiest target. Core's `GetDifficulty` is network-independent — the ratio to the difficulty-1 target — and is computed straight from the nBits exponent/mantissa:

```diff
-let params = Params::new(bitcoin_network(self.chain_network));   // regtest powLimit => always 1.0
-target_to_f64(params.max_attainable_target) / target_to_f64(Target::from_compact(bits))
+let mut shift = (consensus_bits >> 24) & 0xff;                   // Core's GetDifficulty
+let mut difficulty = f64::from(0x0000_ffff_u32) / f64::from(consensus_bits & 0x00ff_ffff);
+while shift < 29 { difficulty *= 256.0; shift += 1; }
+while shift > 29 { difficulty /= 256.0; shift -= 1; }
```

Core's exact operation order is reproduced (rather than the algebraically equivalent target division) so the result is bit-identical to Core's double, not merely close: the tests assert on `f64::to_bits()`. Note Core's UniValue prints doubles with `%.16g` while our RPC path serializes via sonic-rs' shortest-round-trip formatting, so the *same* double is rendered `4.656542373906925e-10` there and `4.6565423739069247e-10` here. That trap is recorded in `docs/solutions/logic-errors/core-float-parity-is-value-parity-not-json-text-parity.md`, since the tempting "fix" (nudging the f64 by a ULP) changes the value in order to make the text match — and Core's text, parsed back, really is the adjacent double.

Since `difficulty_for_bits` is shared, this also corrects `difficulty` in `getblockheader`, `getblock`, `getdifficulty` and `getmininginfo` — consistent with Core, where all of them go through `GetDifficulty`.

`time` is the applied tip header's timestamp and `mediantime` reuses the existing `BlockTree::median_time_past_at(tip, 11)`, so the 11-block window (and its short form near genesis) is not reimplemented here.

**Consistency, from review:** the handler was making four independent tip reads (`applied_height()`, `applied_tip`, `applied_hash()`, `chainwork_hex()`), so mid-sync a single response could mix `blocks` from tip N with `time`/`mediantime`/`difficulty`/`bestblockhash` from tip N+1. It now loads `applied_tip` once and derives every applied-chain field from that one snapshot; `headers` still comes from the header tip, which is the one field that legitimately differs. One behavior change falls out of this: `chainwork` previously came from the *header* tip via `chainwork_hex()` and is now the applied tip's work, matching Core, which reports `ActiveChain().Tip()->nChainWork`. The two only differed while blocks lagged headers.

`verificationprogress` and `warnings` remain stubs. `size_on_disk` is also untouched and still inaccurate — it accumulates the sizes of blocks received by the current process, so it reads `0` after a restart; worth its own fix.

Follow-up to #49, where the discrepancy surfaced while verifying the REST gateway (`/rest/chaininfo.json` returns this handler's output verbatim, so a REST consumer saw the same wrong values).

Requested by: @metaphorics

### DIFF
--- a/CONCEPTS.md
+++ b/CONCEPTS.md
@@ -39,6 +39,20 @@ Bitcoin Core's C++ consensus engine (`libbitcoinkernel`), compiled into `bitcoin
 ### bitcoinconsensus
 Removed historical script verification backend. Previously linked as an extracted C library for non-taproot script checks before being deleted in favor of `bitcoinkernel`. The library lacked complete-prevout and Taproot script-path verification capabilities required for current mainnet script validation (exposed by block 938344 during mainnet IBD).
 
+### Difficulty-1 target
+The network-independent reference target used by Bitcoin Core's difficulty
+calculation: compact nBits `0x1d00ffff`, rather than the selected network's
+PoW limit. Confusing the two makes every network report difficulty `1.0` at
+its easiest target. See
+`docs/solutions/logic-errors/core-float-parity-is-value-parity-not-json-text-parity.md`.
+
+### Float value/text parity
+The distinction between equal IEEE-754 values and equal serialized spellings.
+Core's UniValue uses `%.16g`, while the live RPC path's sonic-rs serializer
+uses shortest-round-trip formatting, so compatibility means preserving the
+value and operation order, not forcing JSON text to match. See
+`docs/solutions/logic-errors/core-float-parity-is-value-parity-not-json-text-parity.md`.
+
 ### Rust interpreter (portable posture)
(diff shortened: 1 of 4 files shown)


metaphorics:
@coderabbitai ultrareview

metaphorics:
@codex review

metaphorics:
@codex review

PR #52, opened 2026-08-11
node: publish Core's `sequence` ZMQ topic for block connects and disconnects

metaphorics:
## Summary

Adds Bitcoin Core's `sequence` ZMQ topic, carrying **block** events only (`C` on connect, `D` on disconnect). This is the last piece a remote [`bip300301_enforcer`](https://github.com/LayerTwo-Labs/bip300301_enforcer) needs to drive its validator off bitcoin-rs: it subscribes to exactly one topic (`sequence`, `lib/cli.rs`), and when `--node-zmq-addr-sequence` is omitted it *discovers* the endpoint by looking for `pubsequence` in `getzmqnotifications` and hard-errors if it is absent. Complements #49 (REST) — no BIP300/301 logic is ported, and none of Core's other notification surface is added.

Frame layout is the three-frame shape its parser requires (`cusf-enforcer-mempool/lib/zmq.rs`), and the topic gets its own counter, which is what makes the counter *unified across event kinds* — one stream, one monotonic `u32`:

```
frame 0: b"sequence"
frame 1: 32-byte block hash (Core wire orientation, same helper as `hashblock`) ++ b'C' | b'D'
frame 2: u32 LE counter, shared by C and D
```

The publisher API takes a `SequenceEvent` enum rather than a method per label, so adding the mempool variants later doesn't reshape `ZmqPublisher`. The four existing topics and their independent counters are untouched.

### Why `A`/`R` are deliberately absent

Core also multiplexes mempool add/remove onto this topic. Emitting them now would be *worse than silence*: the consumer treats a mempool-sequence gap as fatal, but never checks tx-sequence continuity if no tx events arrive — so C/D-only leaves the validator path fully working and degrades the enforcer's mempool mode to an empty mempool instead of a hard error. Our mempool cannot currently produce a gapless per-transaction sequence: `remove_entries` bumps once per *call* rather than per transaction, `prioritise` bumps with no membership change, `insert_entry` discards the IDs of the evictions it triggers, and reorg reconsideration is still unfinished (`docs/solutions/architecture-patterns/node-reorg-execution-design.md`), so no legitimate `A` can follow a `D` yet. That needs a per-transaction event-sequence and a removal-reason model in `crates/mempool` — a separate change, documented rather than faked.

### Emission points

Both labels are published where the chainstate mutation *lands*, which is what makes the stream safe to react to:

```
connect:    apply chainstate -> applied_tip.store(new) -> publish C
disconnect: roll back        -> (tip updated)          -> publish D
reorg:      D(old tip) -> D(old tip-1) -> ... -> C(fork+1) -> ... -> C(new tip)
            counter contiguous across the whole interleaved stream
```

Two ordering bugs are fixed by that placement, both of which a subscriber can actually observe:

- Emitting `C` before `applied_tip.store` let a subscriber act on a connect while RPC still reported the *parent* as the tip — and the enforcer's `C` handler calls straight back into the node.
- Collecting disconnect hashes and publishing them after the connect loop put the new branch's `C`s *before* the old branch's `D`s, so the enforcer would have called `connect_block` for a block whose parent was not its tip. Emitting at the success path of `disconnect_block_admitted` makes publication order equal occurrence order for every caller and removes the deferred list.

Both are pinned by tests: an executor-driven reorg asserts the exact `D…C…` sequence with a contiguous counter, and a publisher that reads `applied_tip` during emission asserts the announced block is already the visible tip.

Configured like the existing topics (`--zmqpubsequence`, `BITCOIN_RS_ZMQPUBSEQUENCE`, `zmqpubsequence` in `bitcoin.conf`, HWM sibling, layered config), and `getzmqnotifications` reports `pubsequence` so endpoint auto-discovery works.

Docs that claimed disconnects are silent (`README.md`, `docs/getting-started.md`, `docs/README.md`, the reorg-design table) are reconciled, and `CONCEPTS.md` gains the sequence-stream term.

Verified live against Bitcoin Core 27.0 on regtest — frames diffed against Core's own `sequence` stream over 200 blocks and a reorg ([evidence](https://github.com/gosuda/bitcoin-rs/pull/52#issuecomment-5254632611)).

Requested by: @metaphorics

### DIFF
--- a/CONCEPTS.md
+++ b/CONCEPTS.md
@@ -137,9 +137,21 @@ keeps its ownership for retry.
 
 Still open around it: returning a disconnected block's transactions through one
 production admission pipeline shared by Electrum, P2P relay, and reorg handling;
-publishing a disconnect notification; and backfilling the filter index after a
-gap. Raw mempool insertion is not reconsideration because it cannot reconstruct
-fee, policy, conflict, and ancestry metadata.
+and backfilling the filter index after a gap. The `pubsequence` stream publishes
+block connect/disconnect notifications, but intentionally does not publish
+mempool `A`/`R` events: the current mempool counter and mutation reasons cannot
+yet guarantee the enforcer's required contiguous transaction event sequence.
+Raw mempool insertion is not reconsideration because it cannot reconstruct fee,
+policy, conflict, and ancestry metadata.
+
+### Sequence stream
+
+The Core-compatible `pubsequence` ZMQ stream is a unified block-event stream.
+Each event carries the block hash, one label (`C` for connect or `D` for
+disconnect), and a topic-local little-endian `u32` sequence counter. Reorg
+disconnects are emitted tip-first before connects on the replacement branch.
+This implementation deliberately omits mempool `A`/`R` events until the
+mempool has per-transaction sequence assignment and explicit removal reasons.
 
 ### Dispatch-bound parallelism
 
--- a/README.md
+++ b/README.md
@@ -150,8 +150,9 @@ header tip against the applied tip each tick and switches branches when the
 applied chain is outweighed.
 
 It is still not the node to depend on. A disconnected block's transactions do
-not return to the mempool, no disconnect notification is published, and the
-filter index is not backfilled across a gap. `docs/README.md` lists the rest.
+not return to the mempool, and the filter index is not backfilled across a gap.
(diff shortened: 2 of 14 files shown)


metaphorics:
@devin Resolve all the feedbacks if relevent.

metaphorics:
@codex review

metaphorics:
@codex review

Issue #75, opened 2026-08-13
Add resumable post-IBD backfill for txindex, Electrum, and block filters


gosunuts:
## Problem

Initial block download currently pays optional indexing work on the synchronous block-apply path whenever an index is enabled:

- `txindex` ingests `TxConfirmed`, `Funding`, `Spending`, and `BlockHeaders` rows in `apply_block`.
- Electrum requires `txindex`, so an Electrum deployment must currently carry that ingest cost during IBD.
- `blockfilterindex` builds and stores BIP157/158 filters during `apply_block`.

With all optional indexes disabled, the node can prioritize body download, consensus validation, UTXO mutation, body persistence, and undo persistence. But enabling an index after synchronization opens an empty index database; there is no production backfill/reindex worker that scans stored active-chain block bodies and catches it up. Filter-index recovery after a gap is explicitly still open.

This forces operators to choose between slower IBD with indexes enabled from genesis and a fast IBD that cannot later serve Electrum/address-UTXO or filter-index queries without resyncing.

## Goal

Allow a node to complete IBD with optional indexes disabled, then build each requested index from locally stored active-chain bodies without re-downloading or revalidating consensus. The worker must be resumable, reorg-safe, bounded, and must converge with blocks arriving at the tip.

## Scope

Start with independently selectable workers for:

1. `txindex` / Electrum confirmed-history data
   - `TxConfirmed`
   - `Funding`
   - `Spending`
   - `BlockHeaders`

2. `blockfilterindex`
   - compact filter bytes
   - filter-header chain

The design should be reusable for future derived indexes, but this issue does not require a generic plugin framework.

### Index lifecycle and metadata

Persist per-index metadata containing at least:

- index schema/version
- lifecycle state: absent, building, caught-up, failed, or stale
- last committed active-chain height and block hash
- active-chain anchor used to start the build
- durable work/checkpoint boundaries

Expose this state in RPC/status surfaces. An index must not claim readiness until its indexed tip equals the node's applied tip on the same active chain.

### Historical scan

- Read only locally persisted active-chain bodies in height order.
- Reuse decoded/verified transaction identifiers where practical, but do not route data through the consensus apply path or mutate UTXO state.
- Batch index writes with bounded memory and configurable I/O/CPU budget.
- Checkpoint progress after durable index batches, so restart resumes from the last committed height/hash.
- Fail clearly if an expected active-chain body is unavailable because of pruning or storage corruption.

### Tip catch-up and reorgs

[...]

Issue #51, opened 2026-08-11
Research: P2P staging-window and database-cache optimization headroom


gosunuts:
## Problem

The local-file processing path is already competitive with Bitcoin Core, but the loopback P2P path still trails Core by approximately 1.37x in the historical matched 0→150k measurement. The current implementation also exposes storage-cache configuration that does not appear to reach the selected backend.

This issue tracks two concrete sources of remaining headroom:

1. P2P download/staging depth and memory representation.
2. Effective database-cache configuration and measured storage behavior.

This is a scoped investigation under #33 and #39. The controlled P2P fixture in #35 should be used for acceptance measurements.

### Fixed 128-block P2P window

`crates/node/src/sync.rs` currently couples:

- `PENDING_BUDGET = 128`
- `RECEIVED_BLOCK_BUDGET = 128`
- a 2 MiB estimated block size
- an approximately 256 MiB pending/staging byte budget
- 16 in-flight blocks per peer while fan-out is active

The count bound is reasonable for modern 1–2 MiB blocks, but it binds far before the byte budget for small historical blocks. This limits both request depth and the script-verification windows formed by P2P sync. The README already identifies the 128-block staging depth as the reason peer sync does not reach local-replay throughput.

A constant-only increase is not acceptable. The count is coupled to staller arming, pending/staged parity, inline capacities, channel drain sizing, peer fan-out, and memory assertions.

### Staging retains decoded and serialized bodies

`crates/node/src/sync/stage.rs` retains both:

- a decoded `bitcoin::Block`
- the original `bytes::Bytes` P2P payload

The byte budget accounts for the serialized payload, while the decoded block adds substantial unaccounted memory. A fully out-of-order window can therefore retain roughly twice the nominal staged bytes.

### `dbcache_mb` is not plumbed into backend construction

`Config::dbcache_mb` is parsed and defaults to 450 MiB, but `NodeStorage::open` calls backend constructors without passing it.

Current backend behavior appears to be:

- Fjall: builder defaults
- RocksDB: fixed 256 MiB block cache
- redb/MDBX: backend defaults

The configured cache target therefore does not currently define a process-level database cache posture. This also makes cache-parity claims difficult to interpret.

### 1. Add attribution before tuning

Measure at least:

- time apply is idle waiting for blocks
- time download/staging is blocked by apply
- per-peer block RTT and bytes/sec EWMA
- inbound queue wait and block decode time
- pending/staged block and byte high-water marks
[...]

Issue #43, opened 2026-08-10
Capture fresh 36-cell baseline matrix


metaphorics:
## Question

What are the current provenance-bound Core and bitcoin-rs medians, guardrails, correctness results, and exclusive diagnostics across three domains, two corpora, two architectures, and three backends?

## Blocked by

- [Build strict offline full-validation comparator](https://github.com/gosuda/bitcoin-rs/issues/34)
- [Build controlled loopback P2P comparator](https://github.com/gosuda/bitcoin-rs/issues/35)
- [Build production MuHash query comparator](https://github.com/gosuda/bitcoin-rs/issues/41)

Issue #39, opened 2026-08-10
Attribute every measured product hot path exhaustively


metaphorics:
## Question

Across all three #45 product workloads and all 36 required cells, what is the overlap-aware whole-workload wall-time contribution of every measured hot path and cost center, and what is the evidence-backed disposition of every repeatable contribution above the predeclared noise floor?

## Scope

This ticket covers every measurable product hot path in all 36 cells, including cells that already exceed 2.0x. SIMD, allocators, and mechanisms named in advance are examples, not limits. Measure and reconcile:

- exclusive and inclusive CPU cost;
- allocation count, bytes, lifetime, fragmentation, and retained memory;
- storage reads, writes, compaction, flush, durability, and page-cache behavior;
- I/O wait, syscalls, copying, parsing, serialization, and network transfer;
- lock contention, atomics, queueing, scheduler delay, backpressure, and idle gaps;
- branch, instruction-cache, data-cache, TLB, vectorization, and memory-bandwidth behavior;
- validation, hashing, Merkle, MuHash, script, UTXO, reorg-readiness, and body-persistence work;
- backend-specific and architecture-specific costs.

Do not stop after finding the first hotspot. Continue until the overlap-aware measured components reconcile with wall time and the unclassified residual is within the predeclared run-to-run noise floor. When profiler evidence is ambiguous, disable or neutralize one stage and measure the whole-workload delta. Treat a mechanism as a campaign candidate only when the attributed cost is large enough to affect a product-cell result.

## Deliverables

- A coverage ledger for every measured product hot path across every applicable workload and cell. Each row records applicability, raw custody, baseline wall contribution, disable or neutralize delta on total product wall time, overlap or concurrency with other stages, affected cells, and disposition.
- A cross-cell ledger that separates shared hot paths from backend-, corpus-, domain-, and architecture-specific paths.
- One ordered candidate list based on whole-product impact, dependency order, reversibility, and correctness risk.
- Explicit dispositions for every repeatable above-noise product bottleneck: optimize with whole-workload remeasurement, already bounded, blocked with evidence, or rejected with evidence.

## Acceptance

- Freeze the attribution method, trial count, and per-workload and per-host noise rule before inspecting attribution results. Do not choose a materiality threshold after seeing the data.
- Every one of the 36 cells has an attributed wall-time account whose explicit residual is no larger than its predeclared run-to-run noise floor.
[...]

PR #80, opened 2026-08-13
perf(index): serve Electrum reads from transaction positions instead of scanning blocks

rabbitson87:
## Why

`crates/index` and `crates/electrum` had no benchmarks, so the G14 Electrum
`get_history` p95 <= 30 ms budget had no number attached to it. Prior campaigns
covered the sync and apply path only.

Measuring it first showed why. `resolve_script_history` loaded and fully decoded
the block once per funding row, then SHA256-hashed **every output script in
it** — cost `O(funding rows x block size)`, both terms linear and multiplying
(63.9x for 64x the rows, 3.6x for 4x the bytes). An address funded at 64 heights
cost **86.53 ms** end to end.

## What changed

Funding and `TxConfirmed` row values were empty. They now carry a packed
`TxPosition[n]` — the byte range of each transaction that produced the row. The
resolvers read only those ranges, through a new `load_range` on the flat-file
store.

**The block-size term is gone.** At 8 funding heights `resolve_script_history`
costs 106.38 µs over 250 KB blocks and 106.26 µs over 1 MB blocks, against
8.182 ms and 33.275 ms scanning.

| Fixture | Resolver | before | after | ratio |
|---|---|---:|---:|---:|
| 1 height, 250 KB | `resolve_script_history` | 1.020 ms | 14.39 µs | 70.9x |
| 64 heights, 250 KB | `resolve_script_history` | 65.644 ms | 836.17 µs | 78.5x |
| 8 heights, 1 MB | `resolve_script_history` | 33.275 ms | 106.26 µs | 313x |
| 8 heights, 1 MB | `resolve_unspent_..._with_height` | 67.713 ms | 106.50 µs | 636x |

End to end, `get_history` at 64 heights goes from ~65.7 ms to **881 µs**.

## The rule that makes it safe

Funding and txid keys carry no block identity, so a superseded block at the same
height leaves rows pointing into a different block's bytes. Rather than pay 8
more bytes per row for a block tag, a resolver **falls back to a full block scan
the moment any single position fails to resolve**, and never skips a failed
position while keeping the rest — that is how a partial result gets reported as
a complete one. See the *All-or-scan position fallback* concept.

## Compatibility

Keys, key ordering and row counts are unchanged. An existing index keeps working:
an empty row value takes the whole-block scan path, retained as `*_scan`. Nothing
forces a reindex; clearing the index directory earns the fast path. A format
marker in `UtxoMeta` is adopted only for an empty index, and a populated index
without one logs a startup warning naming the directory to delete rather than
refusing to start.

**Costs 1.67x row storage** on Funding and `TxConfirmed` (measured: 12 -> 20
bytes/row). Spending rows are untouched — nothing resolves them back to
transactions today.

## Verification

- 99 tests across index/electrum/storage, 0 failures. Workspace suite passes
  except `g14_perf_evidence_script`, which fails 79/81 on this host with the
  changes **stashed** as well: those scripts need Python 3.10+ and read `/proc`.
- 11 equivalence tests, each run **twice** — against a source that serves ranges
  and one that declines — so the position path and the scan fallback are both
  checked against the same oracle. Plus proptests.
- Mutation-verified. The all-or-scan rule initially survived mutation: the first
  stale-position test only produced undecodable bytes, so any implementation
  bailed for the wrong reason. The case that pins it needed both blocks laying
  transactions at identical offsets, so a stale offset lands on a real
  transaction boundary and yields a valid transaction funding something else.

## Two candidates built and rejected, recorded with their numbers

- **Decoded-block cache**: 1.027x *slower* on hits. `block_at_height` returns an
  owned `Block`, so a hit still deep-clones ~2,200 transactions. Caching a value
  the API forces you to copy saves nothing.
- **A block-identity tag in the row value**: correct, but cost another 0.66x
  storage on top of the 1.67x. Replaced by the all-or-scan rule above.

## Not claimed

The G14 gate. `scripts/measure-g14-electrum-rss.sh` remains the evidence of
record and has not been run: it needs Python 3.10+, `/proc`, and a mainnet-tip
node. Everything here is synthetic fixtures on a laptop, and
`docs/benchmarks/index-read-path.md` says so, including a "harness correction"
note about an earlier revision that measured no file I/O and inflated the ratios
by roughly an order of magnitude.

## Known interaction, pre-existing

`prune` + `txindex` is not rejected by this node the way Core rejects it. Index
rows survive pruning while the bodies they point into are deleted, so lookups
for pruned heights return empty rather than an error. Positions inherit this;
they do not create it.


rabbitson87:
Adjacent open issues this touches, for coordination rather than because they block it.

**#75 (post-IBD backfill for txindex/Electrum/filters)** — two of its acceptance criteria move because of this PR:

- *"Txindex backfill produces byte-equivalent rows to an index built inline from genesis"*. Row **values** are no longer empty; they carry `TxPosition[n]`. A backfill worker scanning local bodies goes through `pending_rows_for_block`, which is the same path that populates positions, so byte-equivalence should hold for free — but it is now a property the backfill has to preserve rather than one it gets by writing nothing.
- *"Index metadata records an atomic committed cursor of height, hash, state, and schema version"*. This PR adds a narrower `index:format_version` marker in `UtxoMeta` plus `IndexFormat::{Current, Legacy}`. That is a subset of what #75 wants and should be absorbed into its lifecycle metadata rather than left as a second, parallel notion of index state.

#75 also raises pruning constraints, which is where the pre-existing `prune` + `txindex` interaction noted in the PR description lands.

**#51 (P2P staging window and database-cache headroom)** — its observation that staging retains both the decoded `bitcoin::Block` and the original `bytes::Bytes` is the reason RSS sampled during IBD is unusable for attribution: it swung between 1.1 and 3.2 GB on a run whose UTXO set only grows. Unrelated to this PR's changes, but it is the same measurement trap, and the read-path benchmarks here avoid it by measuring a quiet process.

No open issue covers Electrum/index **read** performance — every performance issue (#16, #32, #33, #39, #43) targets the sync/apply path or SIMD. That gap is why these crates had no benchmarks at all.

rabbitson87:
Addressed in ffe2c4b and 9d40d29. Both `🟠 Major` findings were real, and both invalidated evidence rather than code — they are the two that mattered.

**Electrum benchmark arms were the same call.** Confirmed: one `IndexHandle` over one range-capable source, invoked twice. Every Electrum figure published between the resolver rewrite and now was an absolute cost wearing a paired-arm label. Each arm now has its own handle over the same rows and the same block files, differing only in whether the source serves ranged reads. Re-measured, paired, one run:

| Method | Heights | `before_scan` | `after_fast` | Ratio |
|---|---:|---:|---:|---:|
| `get_history` | 1 | 1.0198 ms | 15.99 µs | 63.8x |
| `get_history` | 8 | 8.3475 ms | 115.94 µs | 72.0x |
| `get_history` | 64 | 67.621 ms | 919.56 µs | 73.5x |
| `subscribe` | 64 | 67.337 ms | 880.63 µs | 76.5x |
| `get_balance` | 64 | 67.505 ms | 942.47 µs | 71.6x |
| `listunspent` | 64 | 68.872 ms | 997.20 µs | 69.1x |

The fixture now also asserts both arms resolve identical history before either is timed — compared field by field rather than as rendered JSON, because `sonic_rs` does not emit object keys in a stable order. That tripped on the first run and briefly looked like a real divergence.

**Nothing proved the fast path was taken.** Also correct, and the sharper form of it is that equivalence is measured *against scanning*, so deleting the position path leaves all eleven tests green. Added `the_position_path_reads_ranges_and_never_whole_blocks`, which counts what the source is asked for and asserts zero whole-block loads plus at least one ranged read on the position path, and the mirror image when ranges are declined.

All five nitpicks taken as well:

- `IndexError::UnaddressablePosition` replaces the three `InvalidHeaderLength { len: usize::MAX }` uses.
- `IndexFormat::UnreadableMarker` separates a malformed marker from a genuine version 0. Went with a distinct variant rather than the suggested `found: None`, since `None` already means "no marker at all" and the operator response differs — the new warning deliberately names no directory to delete, because a re-sync would erase the evidence of whatever wrote the bad bytes.
- `ensure_format_version` probes for one header row instead of counting all of them.
- `block_body_range` logs the discarded `StorageError` at debug; return type unchanged.
- Extracted `FlatFilePruneBodyStore::body_position`.

Verification: 33 tests in `bitcoin-rs-index`, 435 in `bitcoin-rs-node`, clippy clean on `index` and `electrum` with `--all-targets`.

Issue #77, opened 2026-08-13
Architectural boundary: durable chain events and independent derived-state consumers


gosunuts:
## Problem

Optional products currently perform work directly in synchronous block apply: txindex ingestion, BIP157/158 filters, mempool eviction, coinstats updates, and ZMQ publication. Electrum depends on txindex storage; Utreexo has a separate state model; the BIP300/301 enforcer consumes post-apply interfaces. Each new product requirement can therefore add locks, persistence ordering, and reorg handling to consensus-critical code.

## Goal

Keep full-node chainstate transitions authoritative and small, while non-consensus derived state becomes independently recoverable consumers of one durable, ordered chain-event boundary.

    P2P -> headers/bodies -> consensus -> UTXO + body/undo + applied tip
                                                       |
                                          durable ChainEvent source
                                                       |
        txindex | Electrum | filter index | enforcer adapter | Utreexo mode

The default full-node path must not wait for, trust, or become unavailable because a derived consumer is slow, disabled, failed, or rebuilding.

### Authoritative core

The core apply/disconnect path owns consensus validation, default full-UTXO mutation, required body/undo persistence, the authoritative applied tip, and durable emission of ordered state-transition records. Optional indexes, Electrum serving, external notifications, and experimental Utreexo state are not consensus prerequisites.

### Durable event contract

Define a versioned, ordered record containing at least:

- ConnectedBlock: height, hash, parent hash, body locator, and data needed for derived state.
- DisconnectedBlock: height, hash, prior/replacement context, and a durable undo/body locator as needed.
- A monotonic event cursor independent of the ZMQ transport sequence.
- Schema/version and retention guarantees.

The contract must allow a consumer to restart from a stored cursor without relying on best-effort ZMQ. It may be an append-only log or an authoritative-storage journal with equivalent replay guarantees.

Mempool acceptance/removal should later use the same abstraction, but must not block the block-event boundary.

### Consumer lifecycle

Every optional consumer owns:

- an independent durable cursor expressed as event sequence plus height/hash;
- state: disabled, building, caught-up, paused, failed, or stale;
- backfill from a named active-chain anchor;
- idempotent connect/disconnect processing;
- interruption/event-retention/reorg recovery; and
- an explicit query-readiness contract.

A consumer failure must not invalidate chainstate. Its API becomes unavailable or explicitly partial until it catches up.

### Ordering, reorgs, and external notifications

- Events reflect only committed authoritative transitions.
[...]

PR #81, opened 2026-08-13
Feat/77 durable chain events txindex

gosunuts:


crates/node/src/txindex_worker.rs
```diff
@@ -0,0 +1,495 @@
...
+            }
+            if current.height == 0 {
+                return Err(fatal(anyhow!(
+                    "TxIndex and applied chain have no common genesis"
+                )));
+            }
+            let block = self
+                .load_body(current.height, current.hash)
+                .map_err(fatal)?
+                .ok_or_else(|| {
```
metaphorics, line 140: A stale reconciliation target can turn a recoverable missing rollback body into a permanent worker failure here. Suppose the watermark is on branch B, this attempt captured branch A and decided to roll the watermark back, but `applied_tip` switches back to B before this lookup. If the old body is unavailable, this path returns `Fatal` and permanently fails TxIndex even though retrying against the fresh B target would discover that the watermark is already on the authoritative ancestry and no rollback is needed.

Before treating a missing rollback body as fatal, re-check whether the captured target is still authoritative; return `Retry` when it changed, analogous to the forward lookup path. Only report the missing retained body as fatal when the target is still current.

metaphorics (commented):
One correctness issue found in the asynchronous TxIndex reconciliation path.

crates/electrum/src/methods.rs
```diff
@@ -143,15 +199,15 @@ where
...
-        };
-
-        entries
+    fn confirmed_history(
+        &self,
+        scripthash: ScriptHash,
+    ) -> Result<Vec<HistoryRecord>, ElectrumError> {
+        let entries = self.complete_query(|| {
+            self.indexer
+                .resolve_script_history(scripthash, &self.block_source)
```
metaphorics: **The completeness check is anchored to `applied_tip`, but the resolver is not.** `complete_query` proves `watermark == start_tip`, then `resolve_script_history` asks `BlockSource::block_at_height(height)`. In production, `NodeBlockSource::block_at_height` resolves that height from `BlockTree::active_node_at_height`, which walks from the tree's published best-header tip, not from `start_tip.tip_id` / `applied_tip`.

That means a better header branch B can be published while the authoritative applied chain and TxIndex are still on A. The gate passes (`watermark == applied_tip == A`), but this resolver can inspect B's block at an indexed height. The final `applied_tip` recheck still sees A, so the mixed-chain result is accepted. This can produce false negatives or data from an unapplied branch.

Please resolve indexed heights against the captured `start_tip.tip_id` (height -> exact hash -> body), or otherwise make the `BlockSource` used inside `complete_query` explicitly anchored to the same applied-tip snapshot. The same issue applies to the other height-based resolvers (`resolve_transaction`, unspent/outpoint resolution).

metaphorics (commented):
One correctness finding on the completeness boundary.

metaphorics:
Forward TxIndex batching currently limits estimated index-row bytes but retains full decoded block bodies in the batch. Please also bound retained block-body/decoded bytes, or release block bodies after preparing their rows, so the batching limit is a real memory bound.

metaphorics:
**Complete transaction queries can still return authoritative false negatives.**

`getrawtransaction` and Electrum now run `resolve_transaction` inside their complete TxIndex query boundaries, but `resolve_transaction` still silently skips an indexed candidate when `BlockSource::block_at_height(height)` returns `None` and can then return `Ok(None)`. That converts failure to exact-resolve an indexed candidate into `transaction not found` even though the TxIndex row exists.

The new bounded history/unspent resolvers already treat this case correctly by returning `QueryBlockUnavailable`. Please add/use an equivalent strict transaction resolver for complete RPC/Electrum queries, and keep the current skip-on-missing-body behavior only for explicitly best-effort callers.

metaphorics:
**Major: the lock-free completeness boundary can still return an authoritative false negative while a stale worker attempt mutates TxIndex.**

`complete_tx_index_query` checks `watermark == start_tip` once, runs the query, then only verifies that the applied-tip publication identity did not change. That misses a worker attempt which captured an older target before the query started and continues mutating TxIndex afterward.

Example: TxIndex and `applied_tip` are A105. Core briefly publishes B100 and the worker captures B100. Core then publishes a fresh A105 before the worker performs its rollback. A query starting now sees `watermark == start_tip == A105` and passes the initial check, but the still-running B100 attempt can roll the index back and delete A105 rows while the query executes. The final `Arc::ptr_eq(&start_tip, &end_tip)` still succeeds because that fresh A105 publication never changed during the query, so a transient negative can be returned as authoritative.

Please add synchronization that detects or excludes any TxIndex mutation during a complete logical read: restore an equivalent query/mutation gate, use a mutation generation/seqlock around every TxIndex transition and complete read, or execute the watermark check and row reads from one point-in-time DB snapshot. Rechecking only the watermark value is weaker because TxIndex itself can undergo A→B→A while `applied_tip` stays fixed.

metaphorics:
Electrum header RPCs should not read chain headers from the async TxIndex DB. `IndexHandle::headers()` uses TxIndex `BlockHeaders`, while this PR now permits normal `TxIndex watermark < applied_tip` lag. During that lag, `headers.subscribe` reports an old tip and `block.header(s)` can reject already-applied heights; a failed optional worker can freeze header service indefinitely while the authoritative chain advances. Please serve these header methods from the applied-chain view (`applied_tip` + `BlockTree`) rather than TxIndex progress.

metaphorics:
**Keeping TxIndex while deleting `chainstate` makes the prescribed recovery path unable to reconcile it.** The startup error now tells the operator to remove `chainstate`/checkpoints/filters but retain `txindex`. However `FlatFilePruneBodyStore` stores each `(height, hash) -> flat-file position` row in the chainstate KV store, and `load_block_body` returns `None` when that row is gone. After the operator follows this instruction, the retained TxIndex may still have a high watermark but the new chainstate has no body-position rows for it. The worker starts before resync; on the first newly applied tip it tries to roll that watermark back, cannot load its body, and permanently fails. Either keep/reconstruct the block-body index when rebuilding authoritative chainstate, defer reconciliation until those source rows exist, or include `txindex` in the stores that must be removed for this recovery path.

metaphorics:
**`confirmed_history` is not complete: it emits funding transactions but never the transactions that spend this scripthash's outputs.** The new complete path calls `resolve_script_history_limited`, which scans only `ColumnFamily::Funding` and, at each candidate height, emits transactions whose outputs match the scripthash. It never follows those exact funding outpoints through the Spending index and exact-resolves the spending transaction. A wallet can therefore pass the new `watermark == applied_tip` completeness boundary yet receive history/status that omits confirmed spends. Build complete history from both funding transactions and exact-resolved spend candidates (deduping txids), and apply the work bounds to both sides.

metaphorics:
**Normal TxIndex lag can disconnect subscribed Electrum clients.** `Session::serve` propagates `poll_serialized()?`; `poll()` now calls complete history, so `watermark < applied_tip` returns `TxIndexUnavailable` and exits the session. Treat that polling error as transient/no notification and keep cached statuses, rather than closing the connection during ordinary worker catch-up.

metaphorics:
**Public `NodeState` users can enable TxIndex but have no way to start its writer anymore.** `NodeState::open`, `NodeState::apply_block`, and `import_block` are public, while `start_tx_index_worker` is `pub(crate)` and only `run()` calls it. Before this PR, applying through `NodeState` synchronously updated TxIndex; now an external embedder can open `txindex=true` and apply blocks forever while `tx_index_reader()` remains empty/lagging, with no public API to reconcile it. Either make worker startup part of `NodeState` lifecycle, expose an explicit public start/stop lifecycle, or make the standalone state/apply APIs unable to promise TxIndex.

metaphorics:
**This TxIndex slice does not satisfy #77’s published acceptance criteria.** #77 requires a durable ordered event/cursor contract and migration of TxIndex/filter consumers; this PR’s plan explicitly makes the event log/cursor/schema/retention and filter-index migration non-goals while still saying “Issue #77 is complete.” Keep #77 open or scope this PR as a TxIndex subtask rather than closing the parent issue.

metaphorics:
`filter_unspent_outputs_with_height` bounds funding rows, but not exact-resolution work: it can load one full block per candidate spending height and retain every input from those blocks. Bound candidate blocks/body bytes and keep only queried outpoints in `spent`.

PR #83, opened 2026-08-16
perf(utxo): shrink the UTXO record payload 21.7%

rabbitson87:
Measures where the UTXO set's memory actually goes, then shrinks the record payload by 21.7% — and ends up **faster** on the hot lookup than the layout it replaces.

Independent of #80 (no overlapping files); branched from `main`.

## Why

`UtxoSet` is fully memory-resident across 256 shards with no eviction tier, and the published tip-RSS evidence reached 13.83 GiB at height 645,804 against a G14 budget of 16 GiB. Nothing had ever attributed that figure, so the first commit measures before the second changes anything.

## What the measurement found

From a pruned mainnet sync to height 412,732 (38,145,360 outputs across 10,519,335 records), sampled on the checkpoint path:

- The UTXO set is **77.4% of process RSS**, not the ~44% an in-flight sample suggested.
- **Allocator fragmentation is not the problem.** Churning twice the whole set costs 5% more RSS and the curve is flattening. That was the leading hypothesis and it is refuted at this scale on this allocator, which retires the arena work as a candidate.
- **Outputs per record was the assumption that mattered, and it was wrong** — 1.5 assumed, 3.626 measured. The 32-byte txid amortizes over it directly.

## What the change does

```
txid(32) || output_count(4) || legacy_inline_len(1) || widths(1)
|| vout_dir  : one fixed-width little-endian entry per output
|| len_dir   : one fixed-width payload length per output
|| payloads  : varint(amount) [|| raw amount] || varint(height<<1|coinbase) || script
```

Core's `CTxOutCompressor` amount transform, `height`+`coinbase` packed into one varint, and directory widths that are the narrowest the record needs. The script length is not stored — the script is whatever remains of its payload.

**11.75 B/output, 21.7% of the payload**, ~1.97 GiB at tip.

Not done, on purpose: hoisting `height` into the record header would save 3 bytes more but needs "every output of a record shares one height", and BIP30's duplicate coinbase txids are exactly where that might not hold.

## The part worth reviewing

The first draft was a flat varint frame per output. It hit the size target and lost badly on speed:

| operation | v4 | flat v5 | **directory v5** |
|---|---:|---:|---:|
| `get_miss` (real shard lookup) | 705 ns | 3.42 µs | **300 ns** |
| `get_last` | 728 ns | 3.43 µs | **617 ns** |
| `get_middle` | 384 ns | 1.67 µs | **342 ns** |
| `spend_fanout_64` | 18.5 µs | 39.9 µs | 21.3 µs |
| `spend_fanout_64_noop_listener` | 86.7 µs | 115.2 µs | **77.1 µs** |

Two mistakes produced that, and neither was visible until the benchmark was reshaped:

1. **The benchmark measured the wrong operation** — whole-record encode/decode. The hot read is `find_output(vout)`: every spent input resolves through `Shard::get`/`get_entry`/`get_meta` and all three land there.
2. **v4 gets lazy field skipping for free and a flat varint layout cannot.** Every v4 field sits at a constant offset, so when only `vout` is read the optimizer deletes the loads for the rest. In a flat layout each varint's length locates the next field, so the reads are a serial dependency chain and finding output `i` walks the bytes of outputs `0..i`, scripts included.

The directories remove exactly that: a lookup scans one dense fixed-width array and sums a second, touching ~2 bytes per output instead of ~35. Nine of twelve `utxo_commit` lookup arms now beat v4; the three that do not are the `_first` cases, 10 ns apart.

## What it costs

Encoding is 1.6–2.4× slower — the directory widths are a property of the whole record, so nothing can be written until every payload length is known. Commit p95 +3% (`existing`), +8% (`uniform`), +21% (`concentrated`). G14 budget is 50 ms; worst case measured is 2.57 ms.

One encode optimization was tried and **rejected by measurement** (staging both directories in a `SmallVec` and copying once: 505.7 ns vs 428.5 ns). It is reverted, with the number in a comment so nobody retries it.

## Checks

- `tests/record_codec_equivalence.rs` — v4 retained as the oracle, equality **per field** over every decoded `OneUtxoOut` in order. Comparing encoded bytes would be meaningless when the layouts differ by design. Size asserted as a property.
- `non_canonical_v5_spellings_are_rejected` — every second spelling the two layouts introduce: non-minimal varint, the amount escape used for a value the compact form already covers, and a directory wider than the record needs. `UtxoRecord` compares by bytes, so two spellings of one record is a correctness bug.
- `find_output_decompresses_at_most_the_amount_it_returns` — asserts the *work*, not the time. One decompression for a hit, none for a miss, none for `max_vout`. A wall-clock assertion in a test suite is a flake generator.

110 tests in `bitcoin-rs-utxo`; `cargo fmt --check` clean; clippy on `-p bitcoin-rs --all-targets --no-default-features --features "rocksdb,fjall,redb,mdbx,kernel"` reports the same 11 pre-existing errors as `main` (all in `bin/bitcoin-rs/tests/g14_perf_evidence_script.rs`, which this branch does not touch).

## Not in this PR

- Snapshot v5 (write v5, read v4), whose hard assertion is a byte-identical `hash_serialized_3` and MuHash trailer — those are computed over consensus values, not the in-memory encoding, so a v4 file loaded into a v5 build must produce the same trailer.
- G14 tip-RSS evidence, which needs a full-tip sync with `txindex` and `blockfilterindex`.

## Reproduce

```
cargo bench -p bitcoin-rs-utxo --bench record_codec
cargo bench -p bitcoin-rs-utxo --bench utxo_commit -- "lookup|spend_fanout_64"
```

The `utxo_commit` arms cannot be paired in one run — only one codec is compiled in — so the v4 comparison was taken A-B-A across a stash, with the two v5 runs agreeing to 0.1–2.9%. That drift bounds the rebuild effect well below every ratio quoted here.


rabbitson87:
## Correction: the lookup claim was measured on the wrong record size

Pushed in 7d3c948, and retitled the PR. The original framing — "makes lookups faster than v4" — was read off the `utxo_commit` lookup arms. **Those arms hold 256 outputs in one record.** The measured mainnet average is **3.626**, and there v5 is slower. True of the fixture, false of the workload.

Benchmarking only 1/4/16 outputs hid the crossover in one direction and quoting only `utxo_commit` hid it in the other. Both sides are now in the harness:

| outputs | `miss` v4 | v5 | ratio | `hit_last` v4 | v5 | ratio |
|---:|---:|---:|---:|---:|---:|---:|
| 1 | 1.7 ns | 3.6 ns | 0.48x | 2.1 ns | 12.5 ns | 0.16x |
| **4** (≈ mainnet average) | 4.0 ns | 7.0 ns | **0.57x** | 3.7 ns | 18.4 ns | 0.20x |
| 16 | 16.6 ns | 20.6 ns | 0.80x | 16.8 ns | 40.6 ns | 0.41x |
| 64 | 109.3 ns | 79.5 ns | **1.37x** | 126.3 ns | 136.6 ns | 0.92x |
| 256 | 471.8 ns | 298.7 ns | **1.58x** | 481.6 ns | 499.7 ns | 0.96x |

v5 pays a fixed ~10 ns for the directory header and the matched payload, then scans at a fraction of v4's per-output cost. Below ~64 outputs the fixed cost dominates.

**The trade is unchanged and still worth making**, but state it correctly: 3 ns per lookup at the mainnet average is 12 µs per block against a 50 ms commit budget — 0.02% — bought with 21.7% of the record payload. The win concentrates on batch payouts, which is where a lookup was expensive to begin with.

Also worth flagging: the two harnesses disagree on the 256-output case (2.35x through `utxo_commit`, 1.58x here). `utxo_commit` drives the real `UtxoRecord::find_output`; this harness reimplements the v4 search as a direct loop the optimizer handles better, so **its v4 arm is faster than the shipped v4**. The microbenchmark is the conservative number and is the one now quoted.

## Also in this push

`test(node)` (f34e371) — `process_rss_bytes` had **no test at all**, and its Linux branch had never executed: `/proc/self/status` parsing is `cfg`-compiled out on macOS, so it would have run for the first time in production, on the number the memory attribution divides against. Both parsers are now split out of the platform `cfg` so they compile and are tested everywhere, plus an end-to-end assertion a constant-returning stub cannot pass (allocate 128 MB, touch every page, require the reading to move by half of it).

CI confirms the Linux branch now runs:

```
test metrics::tests::proc_status_rss_is_parsed_from_the_kernel_format ... ok
test metrics::tests::process_rss_bytes_tracks_a_real_allocation ... ok
test metrics::tests::ps_rss_output_is_parsed_in_kibibytes ... ok
```

## Linux cross-check

All 113 utxo tests pass under `rust:1.95-bookworm` on aarch64, and the encoded sizes are **byte-identical** to macOS (12.00 / 11.75 / 11.62 B/output at 1 / 4 / 16 outputs), as a deterministic encoding should be.

metaphorics:
@coderabbitai resolve conflicts

PR #85, opened 2026-08-19
perf(rpc): answer gettxoutproof from the txindex instead of scanning the chain

rabbitson87:
Closes nothing; this is candidate **A** from the optimization-candidates survey (#84).

## The problem

Called without a block hash, `gettxoutproof` deep-copies every `BlockRecord`, then loads, deserializes and hashes **every block on the chain** to answer one call. At tip that is roughly a million block loads, almost all discarded.

This is unbounded work for one authenticated RPC call, not a remote DoS. It still stalls the node for the duration and evicts everything else from cache.

## The fix

`Context.indexer` already knows which block confirms a txid. Resolve the confirming height, build the proof from that block alone. Bitcoin Core takes the same route: its `gettxoutproof` requires a block hash *unless* txindex is enabled.

`IndexerLike` gains `resolve_transaction_height`, defaulting to `Ok(None)`, so every existing implementation keeps compiling and no caller may quietly drop its fallback.

## What is preserved

The scan is kept whole as `proof_from_records`. It is the fallback whenever the index cannot answer, and the oracle the equivalence tests compare against:

- no indexer attached
- an unresolved or stale row, or an 8-byte txid-prefix collision
- a pruned body
- a candidate block that does not hold **every** wanted txid

That last one is not belt and braces. BIP30 duplicate coinbase txids mean a txid can confirm in more than one block, so a block chosen from a single txid is a candidate, never a verdict.

The explicit-`blockhash` path and every error code and message are unchanged.

## Measurement

Both arms in one Criterion run over one fixture, on an idle machine (the mainnet IBD sharing the host was stopped first and writeback allowed to drain).

| Arm | `first_block` | `last_block` |
|---|---:|---:|
| `before_scan` | 29.248 us | **21.118 ms** |
| `after_index` | 28.222 us | **32.983 us** |
| ratio | 1.04x | **640x** |

Two positions, because the scan is position-dependent and the index is not. In the scan's best case the arms are within noise — the index costs nothing it does not save. The index arm is flat at 28-33 us across both.

Extrapolating the scan's 10.56 us per record to a mainnet-sized chain gives **about 10 seconds per call**, and that is a **floor**: fixture blocks hold 8 small transactions where real ones hold thousands. Full method and caveats in `docs/benchmarks/gettxoutproof.md`.

`crates/rpc` had no benchmarks at all before this PR.

## Tests, and how they were checked

Twelve tests — nine in `crates/rpc`, three in `crates/index`. They were then audited by mutation, because a green suite proves nothing until it is shown to fail when the behaviour it pins is removed:

| Mutation | Expected | Result |
|---|---|---|
| `proof_via_index` always returns `None` | red | 1 failed |
| the all-wanted-txids guard never fires | red | the 2 predicted tests failed |
| `resolve_transaction_height` always returns `None` | red | 2 failed; the test pinning `None` correctly stayed green |

The audit found two real defects, both fixed here:

- the three `crates/index` tests **were not running at all** — their module is `#[cfg(all(test, feature = "rocksdb"))]` and the crate defaults to no features, so they were filtered out while appearing to pass
- `resolve_transaction_height_agrees_with_the_transaction_resolver` passed vacuously: both resolvers returning `None` satisfied the equality

## Not claimed

`gettxoutproof` is not a G14 budget item. The case is that one RPC call should not do O(chain) work. The fixture is synthetic and establishes the shape of the cost, not its absolute value on mainnet.


rabbitson87:
## Follow-up review found three defects in the path this PR adds

Pushed as `fix(rpc): make the gettxoutproof index path deterministic and non-fatal`.

**An index error failed the whole call.** `proof_via_index` propagated it as `RpcError::Internal`, so a broken txindex turned a call that used to succeed into an error — before this path existed the index was never consulted and the scan answered. A behaviour regression introduced by an optimization, which is what the refactor-set contract exists to prevent. The error is now logged and treated as a miss, and the function returns `Option` rather than `Result` so it cannot come back by accident.

**Only one wanted txid was probed**, and `wanted` is a `HashSet`, so which one was whatever the hasher yielded. One unresolvable txid therefore dropped the call into the full chain scan **non-deterministically** — the precise cost this PR exists to remove. Every wanted txid is now probed before giving up.

**A duplicated docstring** on `proof_from_records` shipped in the first commit. Removed.

## Five more tests

- several txids resolving through the index (only the fallback case was covered)
- an unresolvable probe not abandoning the index path
- an erroring index falling back to the scan
- the explicit-blockhash path never consulting the index, proved with an indexer that panics if used
- a counting indexer pinning that **every** wanted txid is probed

The last one exists because the outcome-based test is only a probabilistic detector: with a `HashSet`, a one-probe implementation picks the resolvable txid about half the time. Counting probes is deterministic.

## Mutation audit, second round

| Mutation | Expected | Result |
|---|---|---|
| probe only one txid (the pre-fix behaviour) | red | `asks_the_index_about_every_wanted_txid` failed |
| one unresolved probe abandons the index path | red | same test failed |
| restored | green | 14 passed |

"An index error must not fail the call" is **not** mutation-detectable any more, because the `Option` return type enforces it. Recording that rather than dressing it up as a passing mutation. The test stays as a guard against anyone reintroducing the `Result`.

Full suite: `bitcoin-rs-rpc` lib, **184 passed**, unfiltered. `cargo fmt --check` clean.

rabbitson87:
## Two more follow-ups, and a claim this PR had wrong

### The fallback still copied the whole log

`ctx.blocks.read().clone()` — at tip roughly 963k records × 168 B ≈ **160 MB allocated and memcpy'd** to answer one call, on the exact path taken when the index cannot answer.

Passing the read guard through instead would be worse: the scan loads a block body from disk per record, and holding that lock across the loop stalls block application for the whole scan. So the length is snapshotted once and each record is copied out under a momentary lock, released before its body is read. Memory is O(1) and the lock never spans I/O.

A test pins it directly — the body source tries to take the write lock while being called and fails if it cannot. Mutation-checked: restoring the held guard turns that test red and nothing else.

`proof_from_records` was then left with one caller passing a one-element slice, so it is now `proof_from_single_record` with both error messages intact.

### The benchmark had only one block shape, and it misled

| Shape | Arm | `first_block` | `last_block` |
|---|---|---:|---:|
| 2,000 × 8 tx | `before_scan` | 16.366 µs | **20.824 ms** |
| | `after_index` | 26.987 µs | **31.312 µs** |
| 200 × 500 tx | `before_scan` | 542.54 µs | **57.228 ms** |
| | `after_index` | 701.30 µs | **700.34 µs** |

Per record: **10.41 µs** at 8 tx, **286.1 µs** at 500. Against 963,124 records that brackets one call at **10 seconds to 4.6 minutes**, where before there was only a floor.

**Correction:** this PR previously reported the scan's best case as `1.04x`, "within noise". That was one fixture shape and it was wrong. The index arm is consistently *slower* there — by 10.6 µs at 8 tx and 158.8 µs at 500 tx. That is what consulting an index costs (row lookup, ranged read, txid comparison), and it does not go away, because the proof needs the block loaded either way.

It is acceptable only because of what "the scan's best case" means: the scan walks from height 0, so it wins only when the wanted transaction is in the **first block on the chain**. The doc now says this rather than claiming parity.

### Noted, not fixed here

At 200–2,000 records `Context::block_by_height` resolves a record by linear scan in negligible time. At 963k records that scan is itself O(chain) — and the index arm pays it too. Separate PR.

PR #98, opened 2026-08-22
fix(rpc): report usage and strippedsize as the quantities they name

rabbitson87:
Item 4 of the Core-compatibility audit work order. Two fields that were **aliases
of the field next to them** — both present, both plausible, both meaning
something else.

## `getmempoolinfo.usage`

Was literally `stats.bytes`.

Core reports `DynamicMemoryUsage()`: how much memory holding the pool costs. That
is a different quantity from the sum of virtual sizes — `bytes` measures the
transactions, `usage` measures the cost of keeping them — and they differ by a
large constant factor.

`Mempool::dynamic_memory_usage()` estimates **this node's** structures: the
per-entry accounting, each transaction's own heap (input and output vectors at
capacity, script and witness bytes), and the four indexes keyed off those
entries. It is an estimate, exactly as Core's is — Core's own carries the comment
*"Estimate the overhead of mapTx to be 9 pointers … as no exact formula for
boost::multi_index_container is implemented"*. The per-entry term follows Core's
shape: **per live entry**, matching `… * mapTx.size()`, not per allocated slot.

It is not a reproduction of Core's number, and cannot be: the structures are not
Core's.

## `getblock.strippedsize`

Was literally `size`.

Core serializes the block **without witness data**, so on any block carrying
witnesses `strippedsize < size`. Reporting `size` made the two fields agree on
every block and hid the witness discount entirely.

Derived from BIP141's `weight = base * 3 + total`, because rust-bitcoin keeps
`Block::base_size` private while exposing both of the other two. The division is
exact by construction.

## The test that could not have caught it

The existing assertion was:

```rust
assert_eq!(block_json.get("strippedsize").as_u64(), Some(block_size));
```

on the **genesis block**, which carries no witnesses — so `strippedsize == size`
is correct there, and the assertion passed whether the field was right or wrong.
The new tests use a witness-bearing block and check against a re-serialization
with every witness emptied, which goes through the encoder rather than through
the weight identity the implementation uses.

## Tests

- `stripped_size_matches_a_witness_free_reserialization` — with an assertion that
  the fixture actually carries witness bytes, so it cannot pass vacuously.
- `stripped_size_equals_the_total_for_a_block_without_witnesses` — the genesis
  case, kept as a boundary rather than as the only case.
- `getblock_reports_a_strippedsize_below_size_for_a_witness_block` — through
  `dispatch`.
- `usage_is_a_different_quantity_from_the_vsize_sum`, `usage_grows_with_the_transactions_it_holds`, `usage_falls_when_the_pool_is_cleared`, `an_empty_pool_reports_no_transaction_memory`, `usage_counts_the_entry_itself_not_only_the_transaction`.
- `getmempoolinfo_usage_is_the_pools_memory_not_its_vsize_sum` — through the handler.

## Mutation audit

7 mutations, **7 killed**, baseline green, run serially.

| mutation | result |
| --- | --- |
| `strippedsize` goes back to the total size | killed |
| the weight identity divides by four | killed |
| stripped size is the weight itself | killed |
| `usage` goes back to the vsize sum | killed |
| `usage` ignores the transactions it holds | killed |
| `usage` ignores the per-entry term | killed |
| transaction heap ignores script bytes | killed |

The per-entry one survived the first pass — a **fixture fault**, the fifth in
five changes. Every usage test asserted only that the number grows and exceeds
`bytes`, and the transaction term alone satisfies all of that; the per-entry term
could be deleted with nothing going red. A pool of transactions with **no inputs
and no outputs** isolates it: those own no heap beyond their struct, and the
funding and spending indexes stay empty, so nothing else can make up the
difference. Chasing that also turned up that the term should follow Core and
count live entries rather than allocated slots, which is now what it does.

## Not changed

`maxmempool` still bounds the **virtual-size sum**, not `usage`. In Core the two
are the same currency and `usage <= maxmempool` holds by construction; here it
does not, and `usage` may exceed `maxmempool`. Changing what the limit bounds is
a policy decision, not a reporting fix, so it is recorded in `DEVIATIONS.md`
rather than made here.

## Verification

- `cargo fmt --check` — clean.
- `cargo clippy --all-targets -- -D warnings` — no finding in any file this PR touches.
- `cargo test -p bitcoin-rs-rpc -p bitcoin-rs-mempool -p bitcoin-rs-node --no-default-features --features bitcoin-rs-node/fjall --no-fail-fast` — green, unfiltered.
- `cargo check --workspace --all-targets` — clean apart from the electrum bench, which does not compile on `main` either.

## Summary by Sourcery

Correct RPC reporting for mempool memory usage and block stripped size, and add regression coverage for both quantities.

Bug Fixes:
- Report mempool memory usage as an estimate of the pool's retained structures and transaction allocations instead of the virtual-size total.
- Report block stripped size without witness data, allowing it to differ from total serialized size for witness-bearing blocks.

Enhancements:
- Add memory accounting for mempool arenas, transaction heaps, transaction indexes, and both collections used by the priority index.
- Align retained-allocation accounting with live pool contents while preserving capacity-based accounting for arenas and hash maps.

Tests:
- Add coverage for witness-bearing and witness-free stripped-size behavior, including RPC wiring.
- Add coverage for mempool usage growth, clearing, retained allocations, transaction data, entry overhead, and priority-index accounting.


metaphorics (commented):

Blocking findings after resolving the current-main conflicts locally:

1. `crates/mempool/src/pool.rs:350`: `dynamic_memory_usage` charges the slab from `entries.len()`, but `slab::Slab` retains backing allocation after removals and `clear`. A grow-then-clear workload therefore reports that retained arena memory as zero. Charge the slab's allocated capacity, while leaving transaction and index payload terms based on live entries. Add a regression that grows and clears the pool and still observes the retained capacity contribution.

2. `crates/mempool/src/pool.rs:373`: the Pareto index term charges one `EntryId` per live transaction. `ParetoFront` stores each entry in both `order: BTreeSet<ParetoKey>` and `keys: BTreeMap<EntryId, ParetoKey>`, so normal non-empty pools materially underreport index payload memory. Give `ParetoFront` a memory estimate that includes both key collections and use it here; pin the lower bound in a test.

The stripped-size calculation and its current block-body/tree path passed review. The fixes above change accounting behavior and require `pareto.rs`, which is outside this PR's changed-file allowlist, so they are blocking rather than queue touch-ups. No local resolution or repair commit was pushed.

rabbitson87:
Both are repaired, rebased on current `main`. 4 mutations, all killed.

**On the allowlist: expanded to `pareto.rs`, deliberately.** Finding 2 asks for an estimate on `ParetoFront`, which cannot be written anywhere else, and splitting it out would leave this PR blocked on a second one for a defect that is its own.

## 1. The arena at capacity

`slab::Slab` keeps its backing allocation across removals and across `clear`, so a pool that peaked at a million transactions and then drained reported that retained memory as **zero** — at exactly the moment an operator reads `usage` to find out where it went. Charged from `capacity()` now.

The txid map turned out to be the same defect one line down: `hashbrown` retains capacity across `clear` too, so it is charged the same way. The transaction and index terms stay keyed to live entries, as you asked — a removed entry's payload really is gone.

**The regression grows and clears, and checks against a pool that never grew.** That comparison is the difference the old accounting erased: both have no live entries, and only one is holding memory. The assertion is `cleared > arena` rather than `>=`, which is what makes the retained map capacity part of the claim instead of an accident.

## 2. The index answers for itself

`ParetoFront::dynamic_memory_usage` counts both collections; the pool delegates.

Two tests, because there are two claims:

- **The floor**, in `pareto.rs` beside the private `ParetoKey` it is stated in terms of: at least two key copies per entry. Deliberately below what the formula computes — the map also carries the id it keys on — so it is a claim rather than a restatement of the implementation.
- **The delegation**, in `pool.rs`: empty the index, leave everything else standing, and the total must fall by exactly what the index said it cost.

That second one exists because my first attempt asserted only that the total was *at least* the index term, and it **passed against the unrepaired code** — the pool was still computing its own figure from the entry count and ignoring the index entirely. Three of the four mutations survived their first run this way; the tests above were written because they did.

## Both figures are lower bounds, and say so

A B-tree leaves its nodes partly filled, so the real footprint is above these and depends on insertion order. Core's own `DynamicMemoryUsage` is an estimate for the same reason — *"no exact formula for `boost::multi_index_container` is implemented"*. What matters is that each term scales with what is stored, which neither of the old ones did.

## One thing the rebase needed

`main` stopped storing headers in `BlockRecord`, so `stripped_size_tests`' `getblock` fixture was falling through to the synthetic path and reading every header-derived field as zero. It seeds the block tree now, and its record height matches the tree node's — the two are joined by height, and a mismatch sends the lookup down the header-only path with no body behind it. The stripped-size logic itself is unchanged from the version you passed.

## Verification

`cargo fmt --all --check` clean. `cargo clippy -p bitcoin-rs-mempool -p bitcoin-rs-rpc -p bitcoin-rs-mining --lib --tests` clean under both `--no-default-features` and default features. `mempool`, `rpc`, `mining` and `node` suites green.

Mutations: the arena charged from length — killed. The index charged one id per entry — killed. Only the ordered set counted — killed. The txid map charged from length — killed.

gosunuts:
Current-main review: this PR is now only partially current. The `getblock.strippedsize` half has already landed in `main` through the canonical block renderer, so keeping that part here would duplicate behavior. The `getmempoolinfo.usage` half is still relevant: `main` still reports `usage` from `stats.bytes` rather than a distinct memory-usage estimate.

I would not merge this PR as-is. Prefer either rebasing it down to the remaining `usage` fix only, or closing it and resubmitting that smaller change against current `main`.

metaphorics (commented):
Gate-and-merge hold at Gate 2 (checks), head `5e106ea`.

This branch is conflict-resolved against `main` 46c1973 (merge 9524669) plus one follow-up fix (5e106ea: missing `TxIn`/`TxOut` imports, unused import, formatting in the four ported files). `bench-smoke` failed on the merge head and passes on this head; `comparator-tests`, `Analyze (actions)`, `Analyze (python)`, `CodeQL` pass.

Blocking: four checks fail identically on `main` HEAD 46c1973 (run 33709944082), so they are pre-existing main breakage, not introduced here. None touches this PR's file set (`crates/mempool/src/pool.rs`, `crates/mempool/src/pareto.rs`, `crates/rpc/src/handlers/chain.rs`, `crates/rpc/src/handlers/mempool.rs`):

- `fmt` — 52 files fail `cargo fmt --all -- --check` on main, none in this PR's set (verified: this PR's four files are `rustfmt --check` clean). https://github.com/gosuda/bitcoin-rs/actions/runs/33721088376/job/100540201778
- `clippy` — `bitcoin-rs-utxo` lib: first doc comment paragraph too long, function too many args (9/8). Same errors on main. https://github.com/gosuda/bitcoin-rs/actions/runs/33721088376/job/100540201501
- `deny` — license requirements unsatisfied plus duplicate-crate entries (`hex-conservative`, `rand`, `syn`, `windows-sys`, …). This PR adds no dependencies. https://github.com/gosuda/bitcoin-rs/actions/runs/33721088376/job/100540201800
- `test` — `crates/rpc/tests/handler_smoke.rs`: `Candidate`, `TemplateId`, `MempoolEntry` not in scope. Same errors on main; that file is outside this PR's set. https://github.com/gosuda/bitcoin-rs/actions/runs/33721088376/job/100540201647

Holding without merge. Will re-gate once main's four jobs go green; no merge on this oid until then.

metaphorics (commented):
Gate-and-merge hold: blocking Gate 4 finding, head `5e106ea` (conflict-resolved, MERGEABLE). Verified against the branch tree.

Blocking, Gate 4 class 1, `crates/mempool/src/pool.rs:607-654` (`dynamic_memory_usage`, backing `getmempoolinfo.usage`): the sum covers arena, payloads, `by_txid`, funding/spending indexes, and Pareto state, but omits other heap-owning pool fields — including `fee_deltas` (pool.rs:131). `prioritise` (pool.rs:871-879) inserts a delta for a txid without requiring it to be present in the pool, so on a fresh pool a single `prioritisetransaction` allocates mempool-owned heap while every term of the usage sum stays zero: `usage` reports 0 for a reachable non-empty memory state. That contradicts this PR's own contract ("report usage ... as the quantity it names"). Include `fee_deltas`, `fee_rate_counts`, and estimator-owned storage under the same accounting policy. The `strippedsize` half checks out (header + CompactSize count + non-witness serialization per tx, marker/flag excluded, matching Core).

Test debt: usage tests assert relative growth only and lack the absent-txid-prioritised state; stripped-size tests are sound (fail the old total-size behavior, non-circular at the helper boundary).

Gate 2: `fmt`, `clippy`, `test`, `deny` fail on this head — the same set `main` itself produces (proven by the byte-identical-to-main PR 176 run). Diagnose outside the queue. No merge on this oid until then.

metaphorics (commented):
## Blocking finding: `dynamic_memory_usage` omits heap-owning fields

`Mempool::dynamic_memory_usage` (`crates/mempool/src/pool.rs:607`) sums the arena, transactions, `by_txid`, `funding`, `spending`, and `pareto` — but three heap-owning fields are missing from the total:

1. **`fee_deltas`** (`:131`, `HashMap<Txid, i64>`) — never counted. `prioritise` (`:871`) inserts into this map *before* checking pool membership (`:879` then `:881`), so a delta for a txid that is not in the pool still allocates an entry. The old usage function reported **0** for a pool that held nothing but a pending fee delta — the exact moment an operator reads `usage` to find where the memory went.

2. **`fee_rate_counts`** (`:122`, `BTreeMap<u64, u64>`) — never counted. Grows with every admission, never shrinks below live entries, but its memory was invisible.

3. **`estimator`** (`:135`, `FeeEstimator`) — never counted. The estimator pre-allocates its bucket vector in `new()` and never shrinks it; the pending map is a `HashMap` that retains capacity across removals. On a fresh pool this is already ~104 KB of allocated heap that `usage` reported as zero.

### Accounting policy

The existing function already establishes two policies:
- **Hash maps** (`by_txid`): charged at **capacity** (retains allocation across removals/clear).
- **B-tree sets** (`funding`, `spending`): charged at **length** (fixed-size keys, payload gone on removal).

The patch follows these same policies:
- `fee_deltas` → capacity (hash map, same as `by_txid`)
- `fee_rate_counts` → length (B-tree map, same as `funding`/`spending`)
- `estimator` → delegates to `FeeEstimator::dynamic_memory_usage`, which charges `buckets` at capacity (pre-allocated Vec) and `pending` at capacity (hash map)

### Patch

```diff
diff --git a/crates/mempool/src/fee_estimator.rs b/crates/mempool/src/fee_estimator.rs
--- a/crates/mempool/src/fee_estimator.rs
+++ b/crates/mempool/src/fee_estimator.rs
@@ -132,6 +132,25 @@
         }
     }
 
+    /// Heap memory this estimator owns: the bucket vector at capacity (it is
+    /// pre-allocated in [`new`](Self::new) and never shrinks) and the pending
+    /// map at capacity (a hash map that retains its allocation across removals,
+    /// same policy as the pool's `by_txid`).
+    #[must_use]
+    pub(crate) fn dynamic_memory_usage(&self) -> u64 {
+        use core::mem::size_of;
+
+        let buckets = u64::try_from(self.buckets.capacity())
+            .unwrap_or(u64::MAX)
+            .saturating_mul(u64::try_from(size_of::<Bucket>()).unwrap_or(0));
+
+        let pending = u64::try_from(self.pending.capacity())
+            .unwrap_or(u64::MAX)
+            .saturating_mul(u64::try_from(size_of::<(Txid, PendingEntry)>()).unwrap_or(0));
+
+        buckets.saturating_add(pending)
+    }
+
     /// Records that a transaction entered the mempool.
     ///
     /// Call this when a transaction is accepted into the mempool.
diff --git a/crates/mempool/src/pool.rs b/crates/mempool/src/pool.rs
--- a/crates/mempool/src/pool.rs
+++ b/crates/mempool/src/pool.rs
@@ -645,12 +645,31 @@
         // `EntryId` per transaction here.
         let pareto = self.pareto.dynamic_memory_usage();
 
+        // `fee_deltas` is a hash map, so it carries slack and retains its
+        // allocation across removals and `clear` — same policy as `by_txid`.
+        // A delta may be stored before its transaction is admitted (see
+        // `prioritise`), so this map can be non-empty on a pool that holds no
+        // transactions at all.
+        let fee_deltas = u64::try_from(self.fee_deltas.capacity())
+            .unwrap_or(u64::MAX)
+            .saturating_mul(u64::try_from(size_of::<(Txid, i64)>()).unwrap_or(0));
+        // `fee_rate_counts` is a B-tree map of fixed-size keys — charged at
+        // length, same policy as `funding` and `spending`.
+        let fee_rate_counts = u64::try_from(self.fee_rate_counts.len())
+            .unwrap_or(u64::MAX)
+            .saturating_mul(u64::try_from(size_of::<(u64, u64)>()).unwrap_or(0));
+        // The estimator owns its bucket vector and pending map.
+        let estimator = self.estimator.dynamic_memory_usage();
+
         arena
             .saturating_add(transactions)
             .saturating_add(by_txid)
             .saturating_add(funding)
             .saturating_add(spending)
             .saturating_add(pareto)
+            .saturating_add(fee_deltas)
+            .saturating_add(fee_rate_counts)
+            .saturating_add(estimator)
     }
 
     /// Estimates the fee rate that historically confirmed within
@@ -3594,9 +3613,20 @@
     #[test]
     fn an_empty_pool_reports_no_transaction_memory() {
         let pool = Mempool::new(MempoolLimits::default());
-        // Only whatever the empty containers hold, which for unallocated ones is
-        // nothing. The point is that it does not report a phantom footprint.
-        assert_eq!(pool.dynamic_memory_usage(), 0);
+        // The estimator's buckets are pre-allocated and never shrink, so a
+        // fresh pool reports that constant overhead — not zero, but not
+        // transaction memory either. The point is that it does not report a
+        // phantom footprint for transactions it does not hold.
+        let estimator_overhead = pool.estimator.dynamic_memory_usage();
+        assert!(
+            estimator_overhead > 0,
+            "the estimator always allocates buckets, so this must be non-zero"
+        );
+        assert_eq!(
+            pool.dynamic_memory_usage(),
+            estimator_overhead,
+            "a fresh pool holds only the estimator's bucket overhead"
+        );
     }
 
     #[test]
@@ -3683,7 +3713,13 @@
             ..MempoolLimits::default()
         })
         .dynamic_memory_usage();
-        assert_eq!(fresh, 0, "a pool that never grew holds nothing");
+        // A fresh pool holds no transactions, but the estimator pre-allocates
+        // its buckets, so `fresh` is the estimator overhead — not zero. The
+        // point is that a drained pool still reports more than a fresh one.
+        assert!(
+            fresh > 0,
+            "the estimator always allocates buckets, so a fresh pool is non-zero"
+        );
         assert!(
             cleared > fresh,
             "a drained pool and a fresh one must not report the same footprint"
@@ -3734,6 +3770,27 @@
             "the pool must charge the index what the index says it costs"
         );
     }
+
+    /// `prioritise` stores a fee delta before checking pool membership, so a
+    /// delta for a txid that is not in the pool still allocates a `fee_deltas`
+    /// entry. The old `dynamic_memory_usage` did not count `fee_deltas`, so it
+    /// reported zero for a pool that held nothing but a pending delta — the
+    /// exact moment an operator asks where the memory went.
+    #[test]
+    fn prioritise_on_an_absent_txid_must_show_in_usage() {
+        let mut pool = Mempool::new(MempoolLimits::default());
+        let before = pool.dynamic_memory_usage();
+
+        let absent_txid = Txid::from(Hash256::from_le_bytes(&[0xAB; 32]));
+        pool.prioritise(absent_txid, 1000).expect("no overflow");
+
+        let after = pool.dynamic_memory_usage();
+        assert!(
+            after > before,
+            "a fee delta for an absent txid allocates in fee_deltas; \
+             usage must reflect it: {after} vs {before}"
+        );
+    }
 }
 
 #[cfg(test)]
```

### Apply instructions

```bash
git apply <patch-file>
cargo test -p bitcoin-rs-mempool dynamic_memory_usage
```

### Verification

- `git apply --check` on a scratch worktree at the PR head: **clean** (exit 0).
- `cargo test -p bitcoin-rs-mempool dynamic_memory_usage`: **7 passed**, including the new regression test `prioritise_on_an_absent_txid_must_show_in_usage`.
- Two existing tests updated: `an_empty_pool_reports_no_transaction_memory` and `the_arena_is_charged_after_the_pool_is_cleared` previously asserted `fresh == 0`, which is no longer true because the estimator's pre-allocated buckets are now correctly counted. Both tests' intent (no phantom transaction memory; drained > fresh) is preserved.

### Note on `fee_rate_floor`

`fee_rate_floor` (`:125`, `Option<u64>`) is a single `u64` on the stack — not a heap allocation — so it is correctly not charged. It is mentioned here only to document that it was considered and excluded deliberately.

PR #100, opened 2026-08-22
fix(rpc): honour gettxout's include_mempool argument

rabbitson87:
Item 5 of the Core-compatibility audit work order — the audit's only Class C
finding: a parameter that was read and discarded.

## The defect

Core's signature is `gettxout "txid" n ( include_mempool )`, with
**`include_mempool` defaulting to `true`**. The handler read params 0 and 1 only,
so this RPC answered from the confirmed set and could not see unconfirmed
activity at all. **A wallet checking an output it had just broadcast got `null`.**

Invisible to `core_compat.rs` for a different reason from the rest of the audit's
findings: the extra argument is simply dropped, so every key is still present and
every value is still of the right type.

## What consulting the mempool means

Two behaviours, from `rpc/blockchain.cpp`:

```cpp
if (fMempool) {
    CCoinsViewMemPool view(coins_view, mempool);
    if (!mempool.isSpent(out)) coin = view.GetCoin(out);
} else {
    coin = coins_view->GetCoin(out);
}
...
if (coin->nHeight == MEMPOOL_HEIGHT) ret.pushKV("confirmations", 0);
```

- An output belonging to an **unconfirmed** transaction is returned, with
  `confirmations: 0` — Core's `MEMPOOL_HEIGHT` renders as zero — and
  `coinbase: false`, since a mempool transaction is never a coinbase.
- An output that a mempool transaction **spends** stops being reported, *even
  though it is still unspent in the confirmed set*. Core documents this on the
  argument itself: *"an unspent output that is spent in the mempool won't
  appear"*. Reporting it spendable hands the caller a conflict.

The second is the one worth reviewing: it makes `gettxout` return `null` for a
coin that is unambiguously in the UTXO set.

## One thing fixed along the way

`Mempool::find_by_outpoint` walked every entry's inputs. `gettxout` asks it on
**every call**, under the mempool read lock that transaction acceptance needs, so
shipping this on top of a linear scan would put an O(n) walk on a
caller-triggered path — the same defect class as #92. It now answers from the
`spending` index, which is keyed by exactly this question.

#92 deliberately left this function alone ("they are a different call path with
different callers, and they belong in their own change"); adding a hot caller is
what makes it this change's business. The two PRs both touch `pool.rs` but not
the same lines.

## Mutation audit

7 mutations, **7 killed**, baseline green, run serially.

| mutation | result |
| --- | --- |
| `include_mempool` defaults to false | killed |
| the parameter is ignored again | killed |
| an outpoint spent in the mempool is still reported | killed |
| a mempool coin reports confirmations as if confirmed | killed |
| a mempool coin is reported as a coinbase | killed |
| the vout bound is not checked | killed |
| `find_by_outpoint` answers from the wrong end of the index | killed |

No fixture fault survived to the audit, because one was headed off while writing
the tests: the spent-in-mempool case seeds the output into the **UTXO set**
first. Without a confirmed coin behind it, *"hidden because the mempool spent
it"* and *"absent because nothing confirmed it"* are the same observation, and
the test would have proven neither.

## Verification

- `cargo fmt --check` — clean.
- `cargo clippy --all-targets -- -D warnings` — no finding in any file this PR touches.
- `cargo test -p bitcoin-rs-rpc -p bitcoin-rs-mempool -p bitcoin-rs-node --no-default-features --features bitcoin-rs-node/fjall --no-fail-fast` — green, unfiltered.
- `cargo check --workspace --all-targets` — clean apart from the electrum bench, which does not compile on `main` either.


metaphorics (commented):

Blocking concurrency finding in `crates/rpc/src/handlers/tx.rs:127-131`: the mempool read guard is dropped after `is_outpoint_spent` returns false and before the confirmed UTXO fallback is read. A concurrent accepted spend can enter that gap, so `gettxout(..., include_mempool=true)` returns a confirmed output that is already spent in the mempool.

The spend check and confirmed fallback need one mempool-consistent observation. Do not blindly hold mempool then acquire UTXO without first proving the node's lock order; if validation takes UTXO before mempool, that would trade the race for a deadlock. Use the established lock order or a sequence/retry snapshot and add a deterministic concurrency regression test.

The remaining include_mempool behavior, spend index integration, and parameter semantics passed review after local conflict resolution. No local resolution commit was pushed.

rabbitson87:
Repaired, rebased on current `main`. 5 mutations: 4 killed, 1 rejected by the compiler.

## The lock order, proved before relying on it

You were right to make this conditional, so here is the argument rather than the assumption.

**`bitcoin-rs-utxo` does not depend on `bitcoin-rs-mempool`** — it is not in its `Cargo.toml` and the name appears nowhere in the crate. So no code path can hold a UTXO shard guard and then ask for the mempool: the shard lock is a leaf, and mempool → UTXO has no cycle to close. `UtxoSet::get_entry` also returns an owned `LiveOutput` rather than a guard, so nothing of the UTXO set outlives the call. The inversion you warned about would need the opposite order, which is why the guard is held rather than the other one taken.

## Holding it is a borrow, not a convention

`confirmed_entry` takes a `&Mempool` borrowed from the read guard. The compiler is what keeps the guard alive across the confirmed read — inserting `drop(pool)` before it gives **E0382**, which is mutation one, recorded as invalid rather than counted as covered. A comment saying "do not release this here" is precisely what failed the first time.

`include_mempool = false` now takes no mempool lock at all, rather than one it has no use for.

## The test, and what it does not cover

A mempool write is held open, `gettxout` is asked, and the answer must not arrive until the write completes. Deterministic in the direction that matters: the read guard is taken before anything is decided, so a handler that answered without consulting the mempool returns immediately and the first assertion fails **every** run, not probabilistically.

**What it cannot show is the gap itself.** Observing that needs the pool to change *between* two operations inside a single call, which needs a scheduling hook this handler does not have — there is no seam between `is_outpoint_spent` and the confirmed read that a test can wedge open. I would rather say that than ship a racing test that looks like it covers the defect and actually only re-checks the semantics. The borrow above is what closes it, and E0382 is the evidence.

## The equivalent mutation

Replacing `confirmed_entry(ctx, &outpoint, &pool)` with `ctx.utxo.get_entry(&outpoint)` under `NotInMempool(_pool)` survives, and it is genuinely equivalent: the guard lives inside the match scrutinee, which is a temporary that outlives the arm either way. What the parameter buys is that the *next* refactor — pulling the arm into an `if let`, hoisting the read — cannot silently drop it. Recorded as equivalent, not as covered.

## Verification

`cargo fmt --all --check` clean. `cargo clippy -p bitcoin-rs-rpc -p bitcoin-rs-mempool --lib --tests` clean under both `--no-default-features` and default features. `rpc`, `mempool` and `node` suites green.

Mutations: the guard released before the confirmed read — **E0382**, invalid. A mempool-spent outpoint falling through to the confirmed set — killed. The mempool never consulted — killed. The spending index returning the wrong txid — killed. The proof parameter dropped — equivalent, as above.

## Conflict note

`main` still has the linear scan in `find_by_outpoint` while carrying the `spending` index for other callers; the merge keeps this branch's indexed lookup, which is the version you passed.

gosunuts:
Current-main review: this PR appears superseded by `main`. The live `gettxout` path now parses `include_mempool`, hides confirmed outputs spent by the mempool, and returns outputs created by unconfirmed transactions with zero confirmations.

Since the core behavior this PR introduced is already present on `main`, I think this can be closed as superseded rather than rebased.

PR #101, opened 2026-08-22
feat(mempool): add the transaction acceptance path

rabbitson87:
Groundwork for the `getblocktemplate` work order. It turned out to be a
prerequisite rather than a detour, and the reason is worth stating first.

## The finding

**This node has no mempool acceptance path.** `sendrawtransaction` put the
decoded transaction into a side `HashMap<Txid, Transaction>` that nothing
treats as the mempool ([`context.rs`][add_tx]). The only non-test caller of
`insert_entry` anywhere is Electrum's `insert_transaction`, and that takes the
**fee as a parameter** rather than deriving it. So in production the mempool is
always empty, no policy check runs, and `testmempoolaccept` answered
`allowed: true` for anything that merely decoded.

Meanwhile the mempool crate is a fully assembled engine with no ignition:

| module | what it has | who called it |
| --- | --- | --- |
| `standardness.rs` | `is_standard_tx` | nothing |
| `rbf.rs` | `check_replacement`, `replace_transaction` | nothing |
| `orphan.rs` | `OrphanPool` — add, expire, per-peer eviction, `take_ready` | nothing |
| `eviction.rs` | `evict_lowest_fee_packages` | nothing |
| `fee_estimator.rs` | `tx_entered` / `tx_left` / `block_connected` | nothing |

The consensus side already fits: `verify_transaction_borrowed_with_mtp` takes
`&impl UtxoView`. So this is wiring, not construction.

## Why this comes before getblocktemplate

Two of the agreed `getblocktemplate` decisions only make sense with an
acceptance path:

- **sigop cost.** `TemplateTransaction.sigops` is hardcoded `0`. The right
  place to count it is where the prevouts are already resolved — P2SH and
  witness sigops are attributed through the spent `scriptPubKey` and are
  invisible to anyone holding only the transaction. That place did not exist.
- **transaction selection.** Selection policy is untestable end to end against
  a pool nothing fills.

## What this adds

- **`accept_to_mempool`** — resolves prevouts against the chain UTXO set with
  the mempool's unconfirmed outputs layered over it (Core's
  `CCoinsViewMemPool`), derives fee, vsize and sigop cost from them, runs
  standardness and full script verification under `VerifyFlags::STANDARD`, and
  inserts through the existing BIP125 path.
- **`check_acceptance`** — the same work without the insertion, which is what
  `testmempoolaccept` needs. It takes `&Mempool`, so admitting a transaction
  from that path is *unrepresentable* rather than merely untested.
- **`MempoolEntry::sigop_cost`**, carried from acceptance. Core stores
  `sigOpCost` on `CTxMemPoolEntry` for the same reason.
- **`MAX_STANDARD_TX_SIGOPS_COST`** enforcement, which was missing entirely.
  Counted before script verification, as Core does in `PreChecks`.

## Core-compatibility fixes found while wiring

- **`sendrawtransaction` on a transaction already in the mempool returns the
  txid** instead of failing. Core's `BroadcastTransaction` rebroadcasts and
  returns OK; callers retry on a dropped connection expecting that to be safe.
  The existing `handler_smoke` test caught this.
- **`testmempoolaccept` reports the real `wtxid`.** It was a copy of the txid.
  Package relay identifies transactions by the witness id.
- **Rejections carry Core's `RPC_VERIFY_REJECTED` (-26)** and a reason string.
- **Datacarrier policy uses `MAX_OP_RETURN_RELAY`**
  (`MAX_STANDARD_TX_WEIGHT / WITNESS_SCALE_FACTOR` = 100 000), not the
  historical 83 bytes. Core removed that limit; using it would have rejected
  traffic Core relays. Read out of the vendored Core 31.99 tree rather than
  from memory.
- A plain insertion limit is no longer reported as a **replacement** failure.
  `replace_transaction` wraps everything in `RbfError`, so "below min relay
  fee" surfaced as "replacement rejected", sending the caller to look for a
  conflict that never existed. A test caught this too.

## Mutation audit

17 mutations, **16 killed, 1 invalid**, baseline green before and after, run
serially.

| mutation | result |
| --- | --- |
| the view stops layering the mempool over the chain | killed |
| sigops are counted without the prevouts | killed |
| the sigop policy limit is not enforced | killed |
| the sigop cost is dropped on the way into the pool | killed |
| a coinbase is admitted to the mempool | killed |
| standardness is never enforced | killed |
| the unresolved outpoints are not reported | killed |
| the fee ignores the outputs | killed |
| a plain insertion limit is reported as a replacement failure | killed |
| the replacement does not report what it evicted | killed |
| a duplicate is not detected | killed |
| a resubmission is an error again | killed |
| testmempoolaccept reports the txid as the wtxid | killed |
| testmempoolaccept reports a placeholder vsize | killed |
| the chain UTXO adapter never resolves anything | killed |
| standardness is enforced on regtest too | killed |
| the adapter reads the txid in the wrong byte order | INVALID (did not build) |

**Two earlier runs of this audit were themselves invalid, and both failure
modes are worth recording:**

1. The restore step used `git checkout --`, which reverts *uncommitted* work.
   It deleted the change under audit and reported every later mutation as a
   broken build. Commit before auditing; restore from a file snapshot.
2. `/mnt/d` is a DrvFs mount with **1-second mtime granularity**. Restoring and
   re-mutating inside the same second leaves cargo's fingerprint unchanged, so
   it reuses the previous binary and the mutation reads as **SURVIVED while
   never having been compiled**. Three did. The fix is a settle delay after
   every write; the tell was that the same mutation, applied by hand, died
   immediately.

### Fixture faults headed off

- The sigop test asserts the **policy limit**, not the number: with the
  prevouts consulted the fixture is over `MAX_STANDARD_TX_SIGOPS_COST`;
  counting blind gives 4. So a mutation that stops reading the prevouts turns a
  rejection into an acceptance. Asserting the count alone would also have
  required the script interpreter, and the portable backend cannot verify a
  P2SH spend — the test would have been silently kernel-only.
- `check_acceptance_leaves_the_pool_untouched` re-accepts the transaction
  afterwards, so "did not insert" is distinguished from "consumed it".
- The child-spends-parent test asserts the **fee**, which can only come from
  the parent's unconfirmed output.

- **p2p transaction relay.** Nothing here accepts transactions from the
  network — `sendrawtransaction` is the only door. That is a separate change
  because it opens a DoS surface, and `getblocktemplate` does not need it.
- **Block-connect mempool maintenance** and `FeeEstimator::block_connected`.
- **Orphan-pool routing.** `AcceptError::MissingInputs` carries the unresolved
  outpoints precisely so that caller can be added later.

## Known untested wiring

Stated rather than papered over. `accept_context` derives two values that no
test pins, because both need a seeded block-tree fixture:

- `locktime_cutoff` from the tip's median time past — every fixture here uses
  `LockTime::ZERO`, so a mutation to `0` survives.
- `height` as `applied_height() + 1`.

Both matter (a wrong cutoff rejects valid timelocked transactions) and both
belong with a block-tree fixture rather than a fake one here.

## Pre-existing, unrelated, reported not fixed

`cargo clippy --all-targets -- -D warnings` is **already red on `main`** for
both crates this PR touches, in files it does not touch:
`crates/mempool/src/orphan.rs` and `standardness.rs` (24 `expect()` findings in
test code), `crates/rpc/src/server.rs` and `crates/rpc/tests/auth.rs`. Same
class as #94. No finding in any file this PR touches.

## Verification

- `cargo clippy -p bitcoin-rs-mempool -p bitcoin-rs-rpc --all-targets -- -D warnings` — no finding in any file this PR touches.
- `cargo test -p bitcoin-rs-mempool -p bitcoin-rs-rpc --no-default-features --no-fail-fast` — green, unfiltered.

Tests run under `--no-default-features`, so script verification used the
portable backend. The kernel backend was not built in this environment.

[add_tx]: https://github.com/gosuda/bitcoin-rs/blob/main/crates/rpc/src/context.rs

---

## Merge order

This is the bottom of the `getblocktemplate` stack: **#101 → #102 → #103 → #105**. Each one needs the one below it, so they merge in that order. Nothing else depends on this PR.


metaphorics (commented):
Gate hold: the transaction acceptance path is not safe to merge.

Blocking findings:

1. `crates/mempool/src/rbf.rs:102-105`: BIP125 rule 1 passes when any direct conflict opts in. A replacement that conflicts with one opt-in transaction and one final transaction can evict both. Each direct conflict must be replaceable through itself or its ancestors.

2. `crates/mempool/src/accept.rs:299-306`: admission reports success from `replace_transaction`, but `insert_entry` can insert, trim for `max_total_bytes`, evict the candidate itself, and still return its id. The RPC can return success for a transaction not in the pool. In replacement cases the originals are removed first, so failure can leave neither originals nor replacement. Stage or preflight the capacity decision and make the whole replacement atomic, restoring every index/counter/original on failure; return success only if the candidate remains indexed.

3. `crates/rpc/src/handlers/tx.rs:343-346`: `sendrawtransaction` ignores its optional `maxfeerate` and the Core default. A change-output error with an extreme fee is admitted despite the RPC's fund-loss guard. Parse Core-compatible units/default, check the acceptance-derived fee rate before mutation under the same atomic acceptance operation, and return the proper rejection.

These are reachable policy, atomicity, and fund-safety defects, not queue touch-ups. No local merge commit was pushed.

rabbitson87:
All three findings are repaired, and current `main` is merged in (one conflict in `crates/rpc/src/context.rs`, both sides' constants kept). 9 mutations, 9 killed.

## 1. BIP125 rule 1

`.any` → `.all`. Rule 1 is a condition on every original a replacement would evict, not on the set containing one willing member.

The test drives one fixture twice — both originals opting in, then one of them final — so the accept and the rejection differ by exactly the thing under test.

## 2. Admission atomicity

`insert_entry` now checks that the arrival survived the trim and reports a new `MempoolError::Full` when it did not, so an id is never returned for a transaction that is not in the pool.

**One deliberate difference from what you asked for, and I would like your call on it.** You asked for the originals to be restored on failure. Bitcoin Core does not do that: `validation.cpp:1393-1401` removes the conflicts, adds the replacement, calls `LimitMempoolSize`, then

```cpp
if (!m_pool.exists(ws.m_hash)) {
    ws.m_state.Invalid(TxValidationResult::TX_RECONSIDERABLE, "mempool full");
```

and the removed conflicts stay removed. I followed Core, because restoring them would be a behavioural deviation in a compatibility target and I did not want to introduce one silently. The reported defect — success reported for a transaction not in the pool — is gone either way.

If you want the stronger restore-the-originals semantics, say so and I will implement it as a deliberate, documented deviation rather than as a repair.

## 3. `maxfeerate`

Implemented with Core's units, default and rounding:

- default `DEFAULT_MAX_RAW_TX_FEE_RATE` = 0.1 BTC/kvB;
- a rate at or above 1 BTC/kvB is refused as a **parameter** (-8), matching `ParseFeeRate`;
- `0` disables the guard;
- the rate becomes an absolute fee for the transaction's vsize with Core's rounding-up `CFeeRate::GetFee`;
- a capped fee answers **-25**, not -26 — `MAX_FEE_EXCEEDED` falls through to `RPC_TRANSACTION_ERROR` in `RPCErrorFromTransactionError`, and the transaction is acceptable, the sender just asked not to send it.

On "before mutation under the same atomic acceptance operation": the ceiling is a field on `AcceptContext` and is applied inside `accept_to_mempool`, between `check_acceptance` and `replace_transaction` — one mempool write lock covers computing the fee, comparing it, and touching the pool. Checking it in the handler would have been two lock acquisitions with a gap between them.

One test pins the unit rather than the number: 0.99 BTC/kvB on a ~110-vbyte transaction is a ceiling near 0.109 BTC, so a 0.99 BTC fee is still refused. Reading the argument as an absolute cap would have sent it.

## Not addressed

`maxburnamount` — Core's other `sendrawtransaction` guard — is also unimplemented. It is a separate guard and was not in the findings, so it is not in this diff.

## Verification

`cargo fmt --all --check` clean. Workspace tests green apart from the `bin/bitcoin-rs` g14 guarded-run harness, which fails on this machine for environmental reasons (it drives `/usr/bin/python3` under a cgroup RSS guard) and passes in CI. `cargo clippy -p bitcoin-rs-mempool -p bitcoin-rs-rpc --all-targets` shows no finding in any file this PR touches; the `benches/pareto.rs` and `orphan.rs` findings are on `main` already.

rabbitson87:
Rebased on current `main`, which brought the Esplora surface in and turned up one real defect plus two fixtures that were resting on the stub this branch replaces.

## `POST /tx` answered a rejected transaction with 503

Esplora's `dispatch_error` maps every unhandled RPC error to `unavailable` — 503, which tells a broadcaster to retry. The one thing a rejected transaction will not do is succeed on a retry. Esplora answers 400 with the reject reason, and a wallet reads that as "fix the transaction" rather than "come back later". `TxRejected` and `TxVerifyError` now map to 400.

This only became reachable here: before this branch, `sendrawtransaction` was a stub and nothing ever hit that arm.

## Two fixtures relied on the stub

- `broadcast_transaction_is_immediately_visible_as_unconfirmed` broadcast a transaction **with no inputs at all**. Under real acceptance that is refused, correctly.
- The surface test rebroadcast its own **confirmed** transaction, whose inputs are already spent.

Both are rebuilt on what acceptance actually admits. Worth recording what that took: Esplora's `prevout` reads the mempool, then the transaction cache, then the index — and **never the UTXO set**, because it has to answer for spent outputs too. So seeding a UTXO row is not enough; the fixture needs a real funding transaction behind the coin.

## On my own verification

The CI failure here was `clippy::match_same_arms` — my new 400 arm had the same body as the existing one. It did not show up locally because I had been running `cargo clippy -p <crate>` and counting `^error` lines, without `-- -D warnings`. CI passes it; I was not. Fixed, and I have re-run CI's exact clippy lanes (`-p bitcoin-rs --all-targets --features rocksdb,fjall,redb,mdbx,kernel -- -D warnings`, and the kernel-parity lane) against this branch — both clean, along with `cargo test -p bitcoin-rs-consensus -- --include-ignored`.

The other twelve PRs in the queue report no CI failures, so the weaker local check did not let anything else through — but the claim I made on them was weaker than it sounded, and this is the correction.

## Verification

`cargo fmt --all --check` clean. CI's clippy and kernel-parity lanes clean. `rpc`, `mempool` and `node` suites green.

gosunuts:
## Review: three blocking findings on the mempool acceptance path

All line references verified against the PR head (`217179f`).

---

### Finding 1 — `admit_transaction` skips consensus verification (reject)

**Site:** `crates/mempool/src/gateway.rs:564–595`

`MempoolGateway::admit_transaction` runs `standardness::evaluate_one` (line 573) then `pool.replace_transaction` (line 595), but never calls `bitcoin_rs_consensus::verify_transaction`. The only caller of `verify_transaction` in the mempool crate is `accept::check_acceptance` (`accept.rs:267`), and nothing in the gateway path invokes `check_acceptance`. A grep across `crates/rpc` for `check_acceptance` and `verify_transaction` returns zero hits — the RPC handler does not verify upstream either.

`evaluate_one` (`standardness.rs:347–392`) checks: already-in-pool, coinbase/missing-inputs, `is_standard_tx`, min-relay-fee, max-fee, replacement, package limits. It does **not** check:

- **Finality** (`LockTime` against `ctx.height` / `ctx.locktime_cutoff`)
- **Overspend** (`value_in < value_out`)
- **Script validity** (signature verification under `VerifyFlags::STANDARD`)
- **Input value overflow** (checked in `check_acceptance` at `accept.rs:240`)

A non-final, overspending, or invalid-signature transaction passes the policy gate and is committed to the pool.

**Why this is not a comment-scope patch:** The fix requires running `verify_transaction` inside the `pool.write()` guard in `admit_transaction`, between the duplicate check (step 3) and `evaluate_one` (step 4). But `verify_transaction` needs a `UtxoView` — the `MempoolUtxoView` that layers mempool parents over chain UTXOs. `AdmissionRequest` currently carries only `PackageTxContext` (fee, vsize, sigop_cost, missing_inputs), not the resolved prevout view. The RPC handler resolves prevouts outside the lock (`resolve_full_context` at `tx.rs:895`) and discards the `Vec<(OutPoint, TxOut)>` — it only keeps the scalar summary.

The smallest correct change is:

1. `AdmissionRequest` gains a field carrying the resolved prevouts (or a `MempoolUtxoView` handle) so `verify_transaction` can run under the write lock.
2. Inside `admit_transaction`, after the duplicate check and before `evaluate_one`, construct the view and call `verify_transaction(tx, &view, request.height, locktime_cutoff, VerifyFlags::STANDARD)`.
3. Map `ConsensusError` to a new `AdmitError` variant (or to `AcceptanceRejectReason` — but that enum has no consensus variant today).
4. The `locktime_cutoff` must come from somewhere — `AdmissionRequest` does not carry it, and `check_acceptance` receives it via `AcceptContext`. Either `AdmissionRequest` grows a `locktime_cutoff: u32` field or the gateway derives it from chain state it does not currently hold.

This is lock-scoped plumbing that changes a public struct and the RPC handler's context resolution, not a mechanical mapper fix. **Rejecting as a comment-scope patch.**

---

### Finding 2 — post-trim survival check missing on the replace path (patch)

**Site:** `crates/mempool/src/pool.rs:327–338` vs `crates/mempool/src/rbf.rs:177–208`

`insert_entry` (pool.rs:328) calls `commit_insert` then checks `contains_txid` (line 337) — if the newly inserted entry was evicted by the size-limit trim inside `commit_insert`, it returns `Err(MempoolError::Full)`. The comment at line 332 explains why: reporting success for a transaction that is not in the pool hands the caller a receipt for a tx that `sendrawtransaction` will treat as accepted.

`replace_transaction` (rbf.rs:177) calls `commit_insert` at line 204 and returns `Ok(self.finish_mutation(changes))` unconditionally — **no survival check**. If the replacement entry is trimmed by the size-limit eviction inside `commit_insert`, `replace_transaction` still returns `Ok`, and `admit_transaction` (gateway.rs:628) returns `Ok(AdmitOutcome::Committed(result))`. The RPC handler then reports success to the sender and calls `ctx.add_transaction(tx)`.

The gateway's `map_err` at line 607 already has a catch-all arm: `other => AdmitError::Policy(Replacement(other))`, so `RbfError::Mempool(MempoolError::Full)` rides that arm without needing a new variant or a new match arm.

**Patch** (verified with `git apply --check` in worktree `~/.omp/wt/p101`):

```diff
--- a/crates/mempool/src/rbf.rs
+++ b/crates/mempool/src/rbf.rs
@@ -202,6 +202,10 @@
         self.remove_entries_with_reasons(&removals, &mut changes);
+        let txid = prepared.entry.txid;
         let replacement = self.commit_insert(prepared);
+        if !self.contains_txid(&txid) {
+            return Err(RbfError::Mempool(MempoolError::Full));
+        }
         changes.extend(replacement.changes);
         Ok(self.finish_mutation(changes))
```

This mirrors `insert_entry` (pool.rs:329–338): capture `txid` from `prepared.entry.txid` before `commit_insert` consumes `prepared`, then check survival after. The `RbfError::Mempool(MempoolError::Full)` surfaces through the gateway's existing catch-all as `AdmitError::Policy(Replacement(RbfError::Mempool(MempoolError::Full)))` — a wart (the reason label says "Replacement" for a size-limit eviction), but correct: the caller gets a rejection instead of a false success.

---

### Finding 3 — `reject_reason_to_rpc_error` folds non-MaxFee rejections into `Internal` (patch)

**Site:** `crates/rpc/src/handlers/tx.rs:848–856`

`reject_reason_to_rpc_error` maps every non-`MaxFeeExceeded` reason to `RpcError::Internal`, which carries code `-32603` (`INTERNAL_ERROR`). But `RpcError::TxRejected` (error.rs:33) exists, `code()` already maps it to `CORE_VERIFY_REJECTED` (`-26`, error.rs:89), and the test at tx.rs:2586 asserts `matches!(error, RpcError::TxRejected(_))` with `error.code() == RpcError::CORE_VERIFY_REJECTED`. The variant is never constructed in production code — only the test expects it.

`TxRejected` holds `String` (error.rs:33) and `reject_reason_to_frozen_string` returns `String` (tx.rs:862) — types match. No existing test asserts `-32603` for these rejection reasons.

**Patch** (verified with `git apply --check` in worktree `~/.omp/wt/p101`):

```diff
--- a/crates/rpc/src/handlers/tx.rs
+++ b/crates/rpc/src/handlers/tx.rs
@@ -848,8 +848,8 @@
 /// Maps an [`AcceptanceRejectReason`] to the frozen RPC error code and
 /// string. `MaxFeeExceeded` is a parameter error (-32602); all others are
-/// internal errors (-32603). The string for `MinRelayFeeNotMet` uses the
+/// transaction rejections (-26). The string for `MinRelayFeeNotMet` uses the
 /// frozen hyphenated form `min-relay-fee-not-met`, not the mempool's Display.
 fn reject_reason_to_rpc_error(reason: AcceptanceRejectReason) -> RpcError {
     match reason {
         AcceptanceRejectReason::MaxFeeExceeded => RpcError::InvalidParams("max-fee-exceeded"),
-        other => RpcError::Internal(reject_reason_to_frozen_string(other)),
+        other => RpcError::TxRejected(reject_reason_to_frozen_string(other)),
     }
 }
```

`RpcError::code()` already maps `TxRejected` → `CORE_VERIFY_REJECTED` (-26) — no change needed in `error.rs`. The doc comment is updated in the same patch so the code and comment agree.

metaphorics (commented):
Gate-and-merge hold: blocking Gate 4 findings, head `217179f` (conflict-resolved onto main, MERGEABLE; follow-up `217179f` restores seven main-side `.outline` ledger files the merge commit had wrongly deleted — verified restored).

Blocking, three findings (all verified against the branch tree):

1. Gate 4 class 2 (trust boundary bypass), `crates/mempool/src/accept.rs:277` vs `crates/mempool/src/gateway.rs:570-595`: the new full verification is never invoked on the production admission path. `sendrawtransaction` goes through `MempoolGateway::admit_transaction`, which evaluates `standardness::evaluate_one` and mutates via `replace_transaction` — `verify_transaction` (finality, duplicate inputs, overspend, scripts) runs nowhere in that path. A resolved-prevout transaction that is non-final or otherwise consensus-invalid can be committed despite the new checks. Reachable on any crafted submission.
2. Gate 4 class 1 (wrong result), `crates/mempool/src/pool.rs:327`: the new post-trim survival check lives only in `insert_entry`, but gateway admission mutates exclusively via `replace_transaction` (gateway.rs:594-595). A low-fee non-conflicting submission to a full mempool can insert, be immediately trimmed, and still return a committed success that RPC caches and reports — while the transaction is absent from the mempool.
3. Gate 4 class 3 (error path), `crates/rpc/src/error.rs:33-38` vs `crates/rpc/src/handlers/tx.rs:reject_reason_to_rpc_error`: the new `TxRejected` (`-26`) variant is defined, code-mapped, and asserted by the new test (tx.rs:2586), but the live mapper still folds every non-max-fee policy rejection into `RpcError::Internal`. A missing-input submission reports JSON-RPC internal error (`-32603`) instead of the intended rejection — contradicting the PR's own test expectation. (The variant is never constructed on the production path.)

Test debt: new tests cover missing-input and dry-run acceptance but not the consensus-invalid/finality/overspend admission cases the RPC path bypasses, nor a size-trimmed `sendrawtransaction` through `replace_transaction`. All three fixes are author design decisions (lock-scoped verification plumbing, replacement-path survival, error taxonomy), so holding rather than redesigning here. No merge on this oid until then.

metaphorics (commented):
## Review: three blocking findings on the mempool acceptance path

All line references verified against the PR head (`217179f`).

---

### Finding 1 — `admit_transaction` skips consensus verification (reject)

**Site:** `crates/mempool/src/gateway.rs:564–595`

`MempoolGateway::admit_transaction` runs `standardness::evaluate_one` (line 573) then `pool.replace_transaction` (line 595), but never calls `bitcoin_rs_consensus::verify_transaction`. The only caller of `verify_transaction` in the crate is `accept::check_acceptance` (`accept.rs:267`), and nothing in the gateway path invokes `check_acceptance`.

`evaluate_one` (`standardness.rs:347–392`) checks: already-in-pool, coinbase/missing-inputs, `is_standard_tx`, min-relay-fee, max-fee, replacement, package limits. It does **not** check:

- **Finality** (`LockTime` against `ctx.height` / `ctx.locktime_cutoff`)
- **Overspend** (`value_in < value_out`)
- **Script validity** (signature verification under `VerifyFlags::STANDARD`)
- **Input value overflow** (checked in `check_acceptance` at `accept.rs:240`)

A non-final, overspending, or invalid-signature transaction passes the policy gate and is committed to the pool.

**Why this is not a comment-scope patch:** The fix requires running `verify_transaction` inside the `pool.write()` guard in `admit_transaction`, between the duplicate check (step 3) and `evaluate_one` (step 4). But `verify_transaction` needs a `UtxoView` — the `MempoolUtxoView` that layers mempool parents over chain UTXOs. `AdmissionRequest` currently carries only `PackageTxContext` (fee, vsize, sigop_cost, missing_inputs), not the resolved prevout view. The RPC handler resolves prevouts outside the lock (`resolve_full_context` at `tx.rs:895`) and discards the `Vec<(OutPoint, TxOut)>` — it only keeps the scalar summary.

The smallest correct change is:

1. `AdmissionRequest` gains a field carrying the resolved prevouts (or a `MempoolUtxoView` handle) so `verify_transaction` can run under the write lock.
2. Inside `admit_transaction`, after the duplicate check and before `evaluate_one`, construct the view and call `verify_transaction(tx, &view, request.height, locktime_cutoff, VerifyFlags::STANDARD)`.
3. Map `ConsensusError` to a new `AdmitError` variant (or to `AcceptanceRejectReason` — but that enum has no consensus variant today).
4. The `locktime_cutoff` must come from somewhere — `AdmissionRequest` does not carry it, and `check_acceptance` receives it via `AcceptContext`. Either `AdmissionRequest` grows a `locktime_cutoff: u32` field or the gateway derives it from chain state it does not currently hold.

This is lock-scoped plumbing that changes a public struct and the RPC handler's context resolution, not a mechanical mapper fix. **Rejecting as a comment-scope patch.**

---

### Finding 2 — post-trim survival check missing on the replace path (reject)

**Site:** `crates/mempool/src/pool.rs:327–338` vs `crates/mempool/src/rbf.rs:177–207`

`insert_entry` (pool.rs:328) calls `commit_insert` then checks `contains_txid` (line 337) — if the newly inserted entry was evicted by the size-limit trim inside `commit_insert`, it returns `Err(MempoolError::Full)`. The comment at line 332 explains why: reporting success for a transaction that is not in the pool hands the caller a receipt for a tx that `sendrawtransaction` will treat as accepted.

`replace_transaction` (rbf.rs:177) calls `commit_insert` at line 206 and returns `Ok(self.finish_mutation(changes))` unconditionally — **no survival check**. If the replacement entry is trimmed by the size-limit eviction inside `commit_insert`, `replace_transaction` still returns `Ok`, and `admit_transaction` (gateway.rs:624) returns `Ok(AdmitOutcome::Committed(result))`. The RPC handler then reports success to the sender and calls `ctx.add_transaction(tx)`.

**Why this is not a comment-scope patch:** The survival check must be added to `replace_transaction` in `rbf.rs`, after `commit_insert` and before `finish_mutation`. The check is `if !self.contains_txid(&candidate.tx.txid()) { return Err(...) }`. But `replace_transaction` returns `Result<MutationResult, RbfError>`, and `RbfError` has no `Full` variant — `MempoolError::Full` is not an `RbfError`. The `RbfError::Mempool` variant wraps `MempoolError`, so `Err(RbfError::Mempool(MempoolError::Full))` would work, but this changes the error contract of `replace_transaction` and the gateway's error mapping at gateway.rs:599–608 would need a new arm to map it to `AdmitError::Policy(AcceptanceRejectReason::...)` — and `AcceptanceRejectReason` has no "trimmed" variant.

The smallest correct change is:

1. In `rbf.rs:replace_transaction`, after `let replacement = self.commit_insert(prepared);` (line 206), before `changes.extend(...)`, add: `if !self.contains_txid(&candidate.tx.txid()) { return Err(RbfError::Mempool(MempoolError::Full)); }`.
2. In `gateway.rs:admit_transaction`, the existing `RbfError::Mempool(MempoolError::Policy(...))` arm (line 601) does not match `MempoolError::Full`. Either add a specific arm mapping `RbfError::Mempool(MempoolError::Full)` to a new `AdmitError` variant, or map it to an existing `AcceptanceRejectReason` — but none fits "evicted by size limit during admission."

This changes error contracts across two crates. **Rejecting as a comment-scope patch.**

---

### Finding 3 — `reject_reason_to_rpc_error` folds non-MaxFee rejections into `Internal` (patch)

**Site:** `crates/rpc/src/handlers/tx.rs:852–855`

`reject_reason_to_rpc_error` maps every non-`MaxFeeExceeded` reason to `RpcError::Internal`, which carries code `-32603` (`INTERNAL_ERROR`). But `RpcError::TxRejected` (error.rs:33) exists, `code()` already maps it to `CORE_VERIFY_REJECTED` (`-26`, error.rs:89), and the test at tx.rs:2586 asserts `matches!(error, RpcError::TxRejected(_))`. The variant is never constructed in production code — only the test expects it.

The doc comment on line 849 even says "all others are internal errors (-32603)", which is wrong: a policy rejection (min-relay-fee, non-standard, missing-inputs, replacement, package-limit) is a transaction rejection, not an internal server failure. Bitcoin Core uses `RPC_VERIFY_REJECTED` (-26) for these.

**Patch** (verified with `git apply --check` in worktree `~/.omp/wt/p101`):

```diff
--- a/crates/rpc/src/handlers/tx.rs
+++ b/crates/rpc/src/handlers/tx.rs
@@ -849,7 +849,7 @@
 fn reject_reason_to_rpc_error(reason: AcceptanceRejectReason) -> RpcError {
     match reason {
         AcceptanceRejectReason::MaxFeeExceeded => RpcError::InvalidParams("max-fee-exceeded"),
-        other => RpcError::Internal(reject_reason_to_frozen_string(other)),
+        other => RpcError::TxRejected(reject_reason_to_frozen_string(other)),
     }
 }
```

`RpcError::code()` already maps `TxRejected` → `CORE_VERIFY_REJECTED` (-26) — no change needed in `error.rs`. The doc comment on line 849 should also be updated from "all others are internal errors (-32603)" to "all others are transaction rejections (-26)" but that is a comment-only change and can follow.

This patch makes the test at tx.rs:2586 pass against the production code path instead of only against a test-only constructor.

PR #102, opened 2026-08-22
feat(mining): select ancestor packages in topological order

rabbitson87:
**Stacked on #101** — review that one first; this branch contains it, and the
diff below is only the two commits after it.

Item 1 of the `getblocktemplate` work order, plus three things that turned out
to be unenforceable without it.

## The defect

Selection walked the Pareto front, which is ordered by fee rate and carries no
dependency information at all. Two ways that produces a template no miner can
use:

- **A high-fee child sorts ahead of the low-fee parent it spends**, so the list
  is not topologically ordered. A miner that serializes it as given builds an
  invalid block.
- **The weight cut-off can land inside a package**, leaving a child whose parent
  was dropped. Same outcome.

Selection now follows Core's `addPackageTxs`: candidates in descending
**ancestor** fee-rate order, each taken together with the unconfirmed ancestors
it still needs, and a package that does not fit skipped whole rather than
truncated. That also makes **child-pays-for-parent work**, which fee-rate order
could not — a child's fee now lifts its parent into the block instead of being
discarded with it.

### Entry ids are not a dependency order

Tempting shortcut, and wrong: the mempool's arena reuses freed slots, so a
parent accepted while the low slots were occupied can hold a **higher** id than
a child inserted after those slots were released.
`entry_ids_are_not_a_dependency_order` builds exactly that pool and asserts the
precondition, so the case cannot quietly stop being the case.

## Three things this makes enforceable

- **Coinbase reserve.** Selection filled the entire four million weight units.
  Adding a coinbase to a full block makes it oversize. It now holds back
  `DEFAULT_BLOCK_RESERVED_WEIGHT` (8000, Core's `policy/policy.h`), while
  `weightlimit` still advertises the full figure, as Core does.
- **Sigop budget.** `TemplateTransaction.sigops` was hardcoded `0`, which left
  `sigoplimit` meaningless — a miner cannot budget sigops it is told are free.
  The template now reports the count acceptance derived (#101) and selection
  bounds the block by it.
- **Witness commitment.** `default_witness_commitment` was a
  `BlockTemplateParams` **input**. It is a function of the selected set, so a
  caller-supplied value is one that can disagree with the set it describes —
  the same shape as the defects this campaign has been removing. It is now
  derived from the selection, and the field is gone.

## Mutation audit

9 mutations, **9 killed**, baseline green, run serially.

| mutation | result |
| --- | --- |
| the package is ordered by entry id rather than by dependency | killed |
| a candidate is taken without its ancestors | killed |
| the sigop budget is not enforced | killed |
| candidates are ordered by own fee rate, not ancestor fee rate | killed |
| a package that does not fit ends selection instead of being skipped | killed |
| the coinbase reserve is not held back | killed |
| the witness commitment ignores the selected transactions | killed |
| the coinbase placeholder is left out of the witness tree | killed |
| the template reports every transaction as sigop-free | killed |

**Four fixture faults, fixed in the second commit** — the first audit run found
them and all four are the same failure of separation:

- `entry_ids_are_not_a_dependency_order` gave parent and child equal fee rates,
  so the parent was reached on its own turn and the two landed in **separate
  one-element packages**. Ordering inside a package was never exercised;
  sorting a package by entry id went unnoticed.
- **Nothing distinguished ancestor-fee-rate ordering from own-fee-rate
  ordering.** For a two-transaction chain they agree, and every fixture was a
  two-transaction chain. `candidates_are_ordered_by_ancestor_fee_rate` builds
  the case where they diverge: a rich child whose parent is large and cheap, so
  the package pays poorly per weight unit and displaces better-paying
  independent transactions.
- `break` versus `continue` on a package that does not fit was
  indistinguishable, because the only oversize package was also the last
  candidate. `a_package_that_does_not_fit_does_not_end_selection` puts a
  smaller candidate behind it.
- Nothing read `TemplateTransaction.sigops`. The budget being enforced
  internally is a different claim from the miner being told the number.

The witness-commitment test checks against rust-bitcoin's own `witness_root` /
`compute_witness_commitment` over a block assembled from the template, rather
than a constant copied out of this implementation.

## Note on the audit harness

Two runs were invalidated by the environment before the numbers above were
trustworthy, and the failure mode is worth knowing about for this repo:
`/mnt/d` is a DrvFs mount whose mtime granularity is coarse enough that cargo
**reuses the previous binary**, so an uncompiled mutation reports as SURVIVED —
and, in the other direction, a restored file keeps running the mutated build
while `git diff` shows the tree is clean. Sleeping is not a reliable fix;
stamping explicitly increasing mtimes is. Every "survivor" here was re-checked
by applying it by hand: two were real fixture faults, two were build-cache
artifacts, and only the by-hand check separates them.

The `getblocktemplate` handler itself still returns its stub. Wiring
`BlockTemplateParams` from chain state, the `rules` handshake, the IBD and
peer-count preconditions, template caching, and `mode: "proposal"` are the rest
of the work order and follow separately.

## Verification

- `cargo clippy -p bitcoin-rs-mining -p bitcoin-rs-rpc --all-targets -- -D warnings` — no finding in any file this PR touches.
- `cargo test -p bitcoin-rs-mining --no-default-features` — green, unfiltered, after a forced rebuild.

---

## Merge order

**Stacked on #101** (`feat/mempool-acceptance`) — package selection reads the sigop cost #101 puts on each mempool entry. Order: **#101 → #102 → #103 → #105**. That base is also why GitHub has never run CI on this PR; it was verified locally with CI's exact commands.


metaphorics (commented):

Dependency: its base PR #101 is held by blocking acceptance and atomicity defects.

Independent blocking findings in this child diff:

1. `crates/mining/src/policy.rs:49`: candidates are sorted once by full ancestor-package fee rate. After one package includes a shared ancestor, sibling descendants are not re-ranked by their now-smaller marginal package. A cheap shared parent with two children can therefore lose fee to a lower-value independent transaction at a tight block boundary. Use iterative modified-package scoring or update affected descendants after each accepted package; score only not-yet-included ancestors and retain deterministic tie-breakers/topological emission.

2. `crates/mining/src/template.rs:90`: selection gives mempool transactions the full block sigop limit and reserves nothing for coinbase sigops. A selected set at the limit plus a miner-supplied payout script with CHECKSIG exceeds consensus limits. Reserve the Core coinbase sigop allowance from the selection budget while advertising the full limit, and use the same reduced budget in the RPC estimate.

rabbitson87:
Both independent findings are repaired, and #101 — with its three repairs — is merged in below this. 9 mutations, 8 killed, 1 equivalent.

## 1. Sibling re-ranking

Candidates were ranked once by their full ancestor package and never again. Each accepted package now re-scores its descendants and pushes them back into the queue; superseded copies are recognised on the way out through a map of each candidate's live score.

The descendant walk is transitive, over the spend index — Core's `CalculateDescendants` reached from `UpdatePackagesForAdded`. A first attempt walked only direct children and the audit caught it.

**One thing worth recording**, because it is the kind of mistake that looks right: I first implemented this as *lazy* re-scoring — recompute a candidate's score when it reaches the front of the queue, re-queue it if it fell. That is the usual lazy decrease-key trick and it is wrong here in the one direction that matters. These scores **rise**, and a score that rises never reaches the front to be re-checked. The fixture below failed against it, which is how I found out.

Package fee rates are now compared by cross-multiplication rather than by dividing. Integer division collapses packages that differ by less than one satoshi per vbyte into ties, and the tie-break then settles by age what the fee should have settled — taking the worse package because it is older. There is a test for exactly that.

## 2. Coinbase sigop reserve

`DEFAULT_COINBASE_RESERVED_SIGOPS` (Core's `DEFAULT_COINBASE_OUTPUT_MAX_ADDITIONAL_SIGOPS`, 400) comes off the selection budget, as the weight reserve already did, while `sigoplimit` still advertises the full figure. Pinned the same way: one sigop short of fitting with the reserve honoured, and fitting with one more.

## Tests

Four new, each built so the two behaviours disagree rather than so the new one passes:

- **the sibling** — one cheap parent, a rich child that pulls it in, a second child whose package is 30 100 per unit while the parent is unpaid and 60 000 once it is in, and a rival at 40 000 between them. Ranked once, the rival takes the last slot; re-ranked, the sibling does.
- **the grandchild** — three deep, so a walk that stops at direct children leaves the grandchild scored against an ancestor already in the block.
- **the fraction** — two packages two satoshis apart over the same size, with the poorer one older. Dividing ties them and age wins.
- **the sigop reserve**.

## The one mutation I did not close

Dropping the superseded-copy guard changes no outcome: `included` already prevents a double-take, and a re-attempt cannot fit where the first attempt did not, because the budget only shrinks. It is recorded as equivalent rather than counted as a pass, and the comment in the code now says what the guard does and does not do instead of claiming a property it does not hold up.

## Verification

`cargo fmt --all --check` clean. `cargo clippy -p bitcoin-rs-mining --all-targets` clean. `mining`, `mempool` and `rpc` suites green, re-run after a forced rebuild.

gosunuts:
Current-main review: this PR is effectively superseded by the mining architecture now on `main`. The mining crate already owns dependency-closed package selection over a mempool mining snapshot, with weight/size/sigop constraints, while the node owns the candidate lifecycle.

The design goal here remains correct, but rebasing this stacked branch would mostly reconstruct functionality that now exists in a different ownership model. I would close this as superseded and file only any specific package-selection behavior that current `main` still lacks.

metaphorics (commented):
Gate-and-merge: fixed-then-stalled, head `7700b7c`.

Fixed on this branch (`7700b7c`, verified: `cargo check -p bitcoin-rs-mining --all-targets` and `-p bitcoin-rs-rpc --lib` pass, file `rustfmt --check` clean): Gate 4 class-1 estimate bug — `estimate_current_block` reduced the weight budget by `DEFAULT_BLOCK_RESERVED_WEIGHT` but passed the raw 80,000 sigop budget without subtracting `DEFAULT_COINBASE_RESERVED_SIGOPS`, against `select_transactions`' documented contract. A mempool whose selection totals 79,600–80,000 sigops would have shipped an over-stated `currentblocktx`/`currentblockweight`.

Held: stack stall — parent PR 101 (`feat/mempool-acceptance`) is held for conflict resolution, and this PR's base is 101's head, so it cannot land first. Re-gate (including CI, which does not run on PR-branch bases) once 101 lands and this retargets to `main`. No merge on this oid until then.

metaphorics:
@greptileai

metaphorics:
@sourcery-ai review

metaphorics:

metaphorics:

metaphorics:
Closing as superseded by the mining overhaul already on main (a12176ff lineage): `select_packages`/`residual_package` in `crates/mining/src/policy.rs` implements the same dependency-closed topological ancestor-package selection this stack's `Ranked`/`select_transactions` implements, with stricter test coverage (12 behavior tests incl. BIP68 skip, atomic residual skip, coinbase reservation vs 3 here) and the sigops reservation at template assembly. `git cherry` shows the two as parallel implementations, and main's is landed and green. Nothing here is lost that main lacks; the branch is preserved.

PR #103, opened 2026-08-22
feat(rpc): build getblocktemplate from chain state

rabbitson87:
**Stacked on #102** (which is stacked on #101) — review those first; this branch
contains them, and the diff below is only the last commit.

Item 6 of the Core-compatibility audit, minus proposal mode.

## The defect

`getblocktemplate` returned a **field-complete stub**: `coinbasevalue: 0`, an
all-zero `target`, `bits: "00000000"`, no transactions, `mintime` and `curtime`
both zero.

Field-complete is what made it dangerous rather than merely incomplete. A miner
that trusted it builds an invalid block — and `core_compat.rs` could not see
the problem, because every key was present and every value had the right type.
It was also absent from `DEVIATIONS.md`, so nothing recorded that the answer was
fiction.

## What it returns now

Assembled through the mining crate, from chain state:

- **`bits` and `target`** come from `expected_next_bits`, which this change
  splits out of `validate_header_nbits`. The template now asks the same
  question a validator asks; deriving difficulty separately would be a second
  implementation of the retarget rules to drift from the first.
- **`mintime`** is the median time past plus one. **`curtime`** is now, never
  earlier than `mintime`.
- **`transactions`, `coinbasevalue`, `sigoplimit`, `weightlimit`,
  `default_witness_commitment`** come from the package selection in #102.

## Core behaviour that was missing entirely

- **The BIP22 rule handshake is enforced.** `rules` must name `segwit`, and
  `signet` must be named on signet, refused with Core's
  `RPC_INVALID_PARAMETER` (-8). Core makes the argument mandatory for a reason:
  without it a pre-segwit miner is handed a template it will turn into an
  invalid block.
- **Preconditions.** On mainnet, `RPC_CLIENT_NOT_CONNECTED` (-9) with no peers
  and `RPC_CLIENT_IN_INITIAL_DOWNLOAD` (-10) when the tip is over a day old.
  Core skips both off mainnet, so a test network can still build templates
  while catching up.
- **`mode: "proposal"` is refused** rather than silently answered as if it were
  a template request. Implementing it needs a validate-without-connecting path
  this node does not have; that is the next change.
- **`rules`** names `csv`, `!segwit` and `taproot`, plus `!signet` on signet.
  **`capabilities`** is empty rather than advertising `"proposal"` we do not
  support.

## Caching

Templates are cached against the tip hash and the **mempool sequence number**,
with a five-second bound on how stale `curtime` may get. Miners poll this method
hard, and assembly walks and sorts the whole mempool under the read lock
transaction acceptance needs to write — rebuilding per request would put that
walk on a caller-controlled path, the same defect class as #92 and #100.

## Mutation audit

12 mutations, **12 killed**, baseline green before and after, run serially.

| mutation | result |
| --- | --- |
| the segwit rule handshake is not enforced | killed |
| proposal mode falls through and is answered as a template | killed |
| the template does not ask the chain for the next difficulty | killed |
| the target is the unmeetable all-zero value again | killed |
| mintime ignores the median time past | killed |
| the csv rule is not advertised | killed |
| the segwit rule marker is not advertised | killed |
| a mainnet node with no peers still serves a template | killed |
| a stale mainnet tip still serves a template | killed |
| tip age is never checked | killed |
| the cache is never consulted | killed |
| a mempool change does not invalidate the cache | killed |

The peers-precondition mutation lives in an integration test, so it was run
against `--test handler_smoke` by hand rather than through the `--lib` harness.

**The cache tests do not measure time.** Two calls in the same second produce
identical output whether or not the cache was consulted, so they plant a marker
in the cached entry and check whether it comes back — and the invalidation test
plants the same marker, changes the mempool, and checks that it does not.

## One test changed rather than fixed

`handler_smoke`'s fixture is a mainnet context with **no peers whose tip is the
2009 genesis block**. Bitcoin Core refuses to build a template in either state,
so the correct assertion there is the refusal, and that is what it now asserts.
The happy path moved to unit tests that can build a regtest context with a
real eleven-block chain — eleven so the median-time-past window is full, since a
shorter chain would let a `mintime` bug hide behind a degenerate median.

## Verification

- `cargo clippy -p bitcoin-rs-rpc -p bitcoin-rs-chain --all-targets -- -D warnings` — no finding in any file this PR touches.
- `cargo test -p bitcoin-rs-rpc -p bitcoin-rs-chain -p bitcoin-rs-mining -p bitcoin-rs-mempool --no-default-features --no-fail-fast` — green, unfiltered.

---

## Merge order

**Stacked on #102** (`feat/mining-package-selection`), which is stacked on #101. The handler assembles templates from the selection those two provide. Order: **#101 → #102 → #103 → #105**. That base is also why GitHub has never run CI on this PR; it was verified locally with CI's exact commands.


metaphorics (commented):

Dependencies: base PR #102 and ancestor #101 are already held by blocking defects.

Independent blockers:

1. `crates/rpc/src/handlers/mining.rs:167-175`: template assembly captures only one applied hash, then independently reloads the applied tip for height, MTP, and next difficulty. A concurrent tip advance can return a template with parent A and height/target/time from B, producing an invalid candidate. Capture one `TipSnapshot`, derive every chain-dependent field from that snapshot and one tree view, then verify the same applied tip before caching/returning; retry if it moved.

2. `crates/rpc/src/handlers/mining.rs:285-289`: `active_rules` advertises CSV unconditionally and Taproot whenever Segwit is active. That reports Taproot during the long Segwit-before-Taproot interval and CSV before its own activation. Derive CSV, Segwit, and Taproot independently from contextual deployment state for the captured parent.

rabbitson87:
Both independent blockers are repaired, with #101 and #102 — and their repairs — merged in below. 5 mutations, 4 killed, 1 equivalent.

## 1. One tip, one tree view

Assembly captured the tip hash and then reloaded the applied tip again for the height, again for the median time, and again for the next difficulty. The tip is now captured once as a `TipSnapshot` and every chain-dependent field comes from it and from a single block-tree read guard. `median_time_past` and `next_bits` are gone — their only callers were those reloads.

After assembly the applied tip is compared against the one the template was built for, and the template is rebuilt if the chain moved. Two retries, then a refusal; a refusal is a worse answer than a template but it is not a wrong one.

**The regression test races.** A thread publishes successive tips while four hundred `getblocktemplate` calls run, and every returned template must name the block one below its own height as its parent. The assertion is exact; only the crossing is probabilistic, so the test can fail only when the bug is present and can never flake once the tip is captured once. Reintroducing the per-field reload kills it.

## 2. Deployments answer for themselves

`csv` was unconditional and `taproot` rode on segwit. Each now comes from its own activation height.

The mainnet interval that second one covers is about four years: every block between segwit and taproot would have been described to a miner as one taproot rules applied to.

The existing rules test asserted `csv` on an eleven-block regtest chain — regtest activates CSV at 432, so that assertion was pinning the invention itself. It now asserts the field is absent there and says why. The new table walks one height below and one height at each activation on mainnet and regtest, so a rule that reported itself as always-on and one that reported itself as never-on both fail.

## The mutation I did not close

Removing the post-assembly tip re-check changes nothing a test can see: the template is internally consistent either way, and the cache records the tip it was built for, so a stale entry can never be served under another tip's key. What the re-check buys is **freshness** — work for the tip as it is now rather than one that stopped being the tip mid-assembly — and observing that needs the tip to move *during* assembly, which needs a hook this code does not have.

It is kept, because it is what you asked for and it is the right behaviour under load. It is recorded as equivalent rather than counted as covered, and the comment in the code now says what it buys and what it does not.

## Verification

`cargo fmt --all --check` clean. `cargo clippy -p bitcoin-rs-rpc --all-targets` clean in every file this PR touches. `rpc`, `mining` and `mempool` suites green, re-run after a forced rebuild.

gosunuts:
Current-main review: this branch is no longer the right path for `getblocktemplate`. `main` now has transport-neutral candidate assembly in `bitcoin-rs-mining` and a node-owned `MiningControl` lifecycle that supplies BIP22/BIP23 data to RPC.

The old stacked implementation overlaps that newer design substantially. I would close this as superseded rather than rebase it. Any remaining RPC compatibility gaps should be fixed directly on top of the current `MiningControl` contract.

metaphorics (commented):
Gate-and-merge hold: stack stall plus a blocking Gate 4 finding, head `7774b91`.

Stall: parent PR 101 is held for conflict resolution and this PR's base is 101's head (via 102), so it cannot land first. Re-gate once the stack lands below it.

Blocking, Gate 4 (concurrency + resource): `mining_notifications: Receiver<()>` (`crates/rpc/src/context.rs:894`, `unbounded()` at :938) is one shared crossbeam receiver for all longpoll waiters (`crates/rpc/src/handlers/mining.rs:148-151`), fed on every tip change (`set_chain_tip`) and every mempool acceptance (`accept_transaction`, context.rs:1149). An mpsc receiver delivers each notification to exactly one waiter, so with two or more concurrent longpollers only one wakes per template change and the rest sleep up to the 60 s timeout on a stale template. And with no waiters the unbounded queue grows one entry per accepted transaction without bound. This needs a broadcast (or per-request subscription) wakeup — an author-level design change, beyond a minor fix. No merge on this oid until then.

metaphorics (commented):
## Long-poll wakeup defect: one mpsc `Receiver<()>` shared by all waiters

### The problem

`Context` creates a single `crossbeam_channel::unbounded()` channel
(`context.rs:938`, `:1003`) whose `Receiver<()>` is shared across every
concurrent `getblocktemplate` long-poll call (`mining.rs:148-151`).

Two defects follow from `mpsc` semantics:

1. **Each notification wakes exactly one waiter.** `recv_timeout` consumes
   one message from the queue; the remaining N−1 waiters sleep until their
   60-second timeout, then return a stale template. With N miners polling,
   only 1/N tip-change notifications reach a waiter within the deadline.

2. **Unbounded queue growth.** `set_chain_tip` (`:1107`) and
   `accept_transaction` (`:1149`) call `send(())` on every event. When no
   waiter is parked (the common case — miners poll at intervals, not
   continuously), every send enqueues a unit that is never drained. The
   queue grows without bound for the lifetime of the process.

### Why mpsc cannot do it

`mpsc` is point-to-point: one consumer per `Receiver`, one message per
`recv`. It cannot broadcast to N waiters, and it cannot collapse redundant
notifications. The required semantics are:

- **Wake-all**: every parked waiter must be released on each notification.
- **No queue growth**: a notification with zero waiters leaves no residue.
- **N concurrent pollers**: the primitive must support multiple simultaneous
  waiters without starvation.

### Fix: `parking_lot::Condvar` + `Mutex<u64>` generation

The codebase already uses this exact pattern: `crates/node/src/shutdown.rs`
pairs `parking_lot::Condvar` with `Mutex<T>` and `notify_all` / `wait_for`.
`parking_lot` is already a direct dependency of `crates/rpc`.

The diff replaces the unbounded mpsc with:

- **`mining_wake: Condvar`** — `notify_all()` wakes every parked waiter.
- **`mining_wake_generation: Mutex<u64>`** — a generation counter incremented
  under the lock before each `notify_all`. This is necessary because
  `accept_transaction` does **not** change `mining_template_id` (only
  `set_chain_tip` does). A `Mutex<()>` would lose tx-only wakeups: the
  waiter would re-check the unchanged id and re-enter `wait_for` for another
  full 60 s. The generation counter gives the waiter an independent signal
  that template inputs changed.

**Waiter** (`mining.rs`): captures the generation under the lock *before*
checking `mining_template_id`, then loops on `wait_for` until either the id
changes (tip advance) or the generation changes (tx accept or tip advance),
breaking on timeout. Capturing the generation before the id check closes the
race where a send fires between the check and the lock acquisition.

**Senders** (`context.rs`): both `set_chain_tip` and `accept_transaction`
increment the generation under the lock, release it, then `notify_all`.

`cargo check -p bitcoin-rs-rpc` passes. `git apply --check` passes on a
clean checkout of `7774b91`.

```diff
diff --git i/crates/rpc/src/context.rs w/crates/rpc/src/context.rs
index 460eb49..f93091b 100644
--- i/crates/rpc/src/context.rs
+++ w/crates/rpc/src/context.rs
@@ -14,9 +14,8 @@ use bitcoin_rs_mempool::{
 };
 use bitcoin_rs_primitives::{Hash256, Network, OutPoint as InternalOutPoint};
 use compact_str::CompactString;
-use crossbeam_channel::{Receiver, Sender, unbounded};
+use parking_lot::{Condvar, Mutex, RwLock};
 use hashbrown::HashMap;
-use parking_lot::RwLock;
 
 const SERIALIZED_BLOCK_HEADER_LEN: usize = 80;
 
@@ -890,8 +889,14 @@ pub struct Context {
     pub mining_template_id: Arc<ArcSwap<CompactString>>,
     /// Most recently assembled `getblocktemplate` response, if still current.
     pub mining_template_cache: Arc<RwLock<Option<CachedBlockTemplate>>>,
-    /// Receiver notified when mining template inputs change.
-    pub mining_notifications: Receiver<()>,
+    /// Condvar that wakes all `getblocktemplate` long-poll waiters when
+    /// mining template inputs change (chain tip advance or mempool tx).
+    pub mining_wake: Condvar,
+    /// Generation counter paired with `mining_wake`. Each send site
+    /// increments under the lock before `notify_all`; waiters capture it
+    /// before sleeping and exit when it changes, so tx-only wakeups
+    /// (which do not alter `mining_template_id`) are not lost.
+    pub mining_wake_generation: Mutex<u64>,
     /// Optional outbound channel that submits decoded blocks back to the node's
     /// `BlockSync::tick` for the canonical apply path. `None` when no node is
     /// wired (tests, embedded callers).
@@ -905,7 +910,6 @@ pub struct Context {
     pub added_nodes: Arc<parking_lot::RwLock<Vec<std::net::SocketAddr>>>,
     /// Active ZMQ PUB notifications.
     pub zmq_notifications: Arc<[ZmqNotification]>,
-    mining_sender: Sender<()>,
 }
 // SAFETY: `Context` is shared by RPC worker threads. Each mutable subsystem
 // handle behind it uses atomics, channels, or locks for interior mutation.
@@ -935,7 +939,6 @@ impl Context {
     #[must_use]
     #[allow(clippy::arc_with_non_send_sync)]
     pub fn new() -> Self {
-        let (mining_sender, mining_notifications) = unbounded();
         let coin_stats_listener = bitcoin_rs_coinstats::CoinStatsListener::new(
             bitcoin_rs_coinstats::CoinStats::default(),
         );
@@ -963,13 +966,13 @@ impl Context {
             block_body_source: None,
             mining_template_id: Arc::new(ArcSwap::from_pointee(CompactString::new("0"))),
             mining_template_cache: Arc::new(RwLock::new(None)),
-            mining_notifications,
+            mining_wake: Condvar::new(),
+            mining_wake_generation: Mutex::new(0),
             inbound_blocks_sender: None,
             p2p_outbound_sender: None,
             banned: Arc::new(parking_lot::RwLock::new(Vec::new())),
             added_nodes: Arc::new(parking_lot::RwLock::new(Vec::new())),
             zmq_notifications: Arc::from(Vec::<ZmqNotification>::new()),
-            mining_sender,
         }
     }
     /// Builds a context that shares pre-existing handles owned elsewhere
@@ -1000,7 +1003,6 @@ impl Context {
         added_nodes: Arc<parking_lot::RwLock<Vec<std::net::SocketAddr>>>,
         tx_index: Option<Arc<dyn TxIndexQuery>>,
     ) -> Self {
-        let (mining_sender, mining_notifications) = unbounded();
         Self {
             chain_tip,
             applied_tip,
@@ -1020,7 +1022,8 @@ impl Context {
             block_body_source: None,
             mining_template_id,
             mining_template_cache: Arc::new(RwLock::new(None)),
-            mining_notifications,
+            mining_wake: Condvar::new(),
+            mining_wake_generation: Mutex::new(0),
             inbound_blocks_sender,
             p2p_outbound_sender,
             banned,
@@ -1028,7 +1031,6 @@ impl Context {
             prune_service: None,
             chain_control: None,
             zmq_notifications: Arc::from(Vec::<ZmqNotification>::new()),
-            mining_sender,
         }
     }
 
@@ -1104,7 +1106,8 @@ impl Context {
         self.mining_template_id
             .store(Arc::new(CompactString::from(tip.hash.to_string_be())));
         self.chain_tip.store(Some(Arc::new(tip)));
-        let _ignored = self.mining_sender.send(());
+        *self.mining_wake_generation.lock() += 1;
+        self.mining_wake.notify_all();
     }
 
     /// Publishes a new best-applied-block tip.
@@ -1146,7 +1149,8 @@ impl Context {
         };
         // A new mempool transaction changes what the next block template would
         // contain, so long polls waiting on it should wake.
-        let _ignored = self.mining_sender.send(());
+        *self.mining_wake_generation.lock() += 1;
+        self.mining_wake.notify_all();
         Ok(result)
     }
 
diff --git i/crates/rpc/src/handlers/mining.rs w/crates/rpc/src/handlers/mining.rs
index f03079f..72902fd 100644
--- i/crates/rpc/src/handlers/mining.rs
+++ w/crates/rpc/src/handlers/mining.rs
@@ -145,10 +145,16 @@ fn wait_for_long_poll(ctx: &Arc<Context>, request: Option<&Value>) {
     else {
         return;
     };
+    let mut generation = ctx.mining_wake_generation.lock();
+    let waited = *generation;
     if longpollid == ctx.mining_template_id.load().as_str() {
-        let _result = ctx
-            .mining_notifications
-            .recv_timeout(Duration::from_mins(1));
+        let deadline = Duration::from_mins(1);
+        while longpollid == ctx.mining_template_id.load().as_str() && *generation == waited {
+            let result = ctx.mining_wake.wait_for(&mut generation, deadline);
+            if result.timed_out() {
+                break;
+            }
+        }
     }
 }
```

PR #105, opened 2026-08-22
feat(rpc): answer getblocktemplate block proposals

rabbitson87:
## What

`getblocktemplate mode: "proposal"` (BIP23) was refused with an RPC error. A miner asking whether a block it assembled would connect got nothing it could act on. This answers the question.

## The path that already existed

Answering a proposal means validating a full block against the tip **without connecting it**, and the node had no such path. It turned out to have one already, unnamed: everything `apply_block_admitted` does before `persist_undo` — its first write of any kind — is pure validation.

| | |
|---|---|
| predecessor lookup | `applied_predecessor` |
| proof of work | `header.validate_pow` |
| block rules | `verify_block_rules_borrowed_contextual_with_txids_and_witness_hint` |
| BIP30 / BIP34 | `check_bip30_and_bip34` |
| difficulty continuity | `check_pow_limit_and_continuity` |
| transactions and scripts | `verify_block_transactions` |
| coinbase maturity | `check_coinbase_maturity_with_tx_plan` |
| BIP68 | `check_bip68_sequence_locks` |
| coinbase amount | `check_coinbase_amount` |

That prefix is now `validate_block_for_apply`, and `test_block_validity` is the connect path minus its second half. **A proposal is weighed by the same function a connected block is weighed by** — not by a second set of checks written to resemble it. A validator that drifted would answer "valid" for blocks the node then rejects, which is worse than answering nothing.

No `dry_run` flag threads through the mutation sites, and no scratch overlay was needed: the checks read the live UTXO set and write nothing, and `build_utxo_changes` computes the block's value totals without committing them.

## Proof of work is the one check a proposal skips

Through a `PowCheck` switch, because a proposal is a block whose nonce has not been searched for yet. Core does the same with `fCheckPOW = false`. The *declared* difficulty is a different question and stays checked — `check_pow_limit_and_continuity` sits below the switch, not behind it.

## Wiring

`bitcoin-rs-node` depends on `bitcoin-rs-rpc`, not the reverse, so the RPC crate cannot reach the validator. `ChainControl` gains the method and the node installs it at startup, exactly as it already does for `invalidate_block`. **A node with no validator installed refuses the proposal rather than approving it** — answering `null` there would tell a miner the block is valid on the authority of nothing having been checked.

## What the handler answers

- `duplicate` / `duplicate-invalid` / `duplicate-inconclusive` for a block the tree already holds, by its node status
- `inconclusive-not-best-prevblk` for one built on another tip — the UTXO set this node can weigh a block against is the one at its own tip
- `null` for a block that would connect
- otherwise the reject reason

Reject reasons carry Core's own token where one corresponds — `bad-cb-amount`, `bad-txnmrklroot`, `bad-txns-inputs-missingorspent` — transcribed from the vendored Core tree (`src/validation.cpp`, `src/consensus/tx_check.cpp`, `src/consensus/tx_verify.cpp`), not from memory. A rejection with no Core counterpart passes through as its own message rather than borrowing a token that means something else; `DEVIATIONS.md §9` records that and the two block-wide rejections reported with Core's per-transaction word.

`capabilities` now names `proposal`, and proposals are answered **before** the `rules` handshake and before the connected / still-syncing preconditions, as Core answers them: a proposal asks about one specific block, it does not ask for work.

## Tests

15 new, all mutation-audited.

**node** — the dry run leaves the tip where it was and the block still applies afterwards (a run that had spent the seeded coin would answer differently the second time); a block with no valid nonce is accepted while the connect path still rejects the same block with `ProofOfWork`; an over-paying coinbase is refused *through the dry run*, which is the last gate in the sequence and the one needing resolved prevouts, so reaching it proves the whole path ran; a wrong merkle root is refused.

**rpc** — every verdict above, both error codes (`-3` missing `data`, `-22` undecodable, over both halves of the decode), the missing validator, the rules handshake not applying, and the advertised capability.

**reject reasons** — the Core tokens, and that an unmapped rejection keeps its own message.

### Mutation audit

14 mutations, **14 killed**. One survivor first time round was a real coverage gap, not a false alarm: the fixture used `deadbeef`, which is valid hexadecimal, so the hex-decode arm never ran. Closed in the second commit, then both arms re-audited and killed.

## Stacking

Based on `feat/rpc-getblocktemplate` (#103, which is on #102 → #101) and **merges `fix/consensus-coinbase-amount` (#104)**. #104 is not optional here: without the coinbase-amount rule this would answer "valid" for a block that creates money, which is precisely the claim proposal mode must not make falsely. Once #104 merges the diff shrinks to this change alone.

## Verification

WSL, `-j4`. `cargo fmt --all --check` clean. `cargo clippy -p bitcoin-rs-node -p bitcoin-rs-rpc --all-targets` shows only the three `expect_used` findings that are on `main` and that #94 fixes. Full `node` + `rpc` suites green, re-run after a forced rebuild to rule out a stale binary.

---

## Merge order

Top of the `getblocktemplate` stack until now: **#101 → #102 → #103 → #105**. **#109** (`feat/rpc-mininginfo-next`) is stacked on this one in turn — `getmininginfo`'s `next` object needs the `expected_next_bits` split out in #103, and its `currentblockweight` needs the assembled-template cache added there. So the full order is **#101 → #102 → #103 → #105 → #109**.

This PR also **contains #104** (`fix/consensus-coinbase-amount`), merged in: without the coinbase-value rule a proposal would answer "valid" for a block that creates money. Once #104 lands on `main` the diff here shrinks to the proposal change alone.

That base is also why GitHub has never run CI on this PR: the workflows trigger on `main`-based pull requests only, so it was verified locally with CI's exact commands.


metaphorics (commented):

Dependencies: #101, #102, and #103 are held by blocking defects.

Independent concurrency blocker in `crates/node/src/apply.rs:2310-2314`: `test_block_validity` calls the connect-path validation without `handles.begin_chain_transition()`. A concurrent connect or disconnect can change the applied tip and committed UTXO set between predecessor/activation/DAA reads, prevout/script checks, and coinbase-fee calculation. Proposal mode can then validate against mixed chainstates or report an old-parent proposal valid against a new UTXO set.

Acquire the established chain-transition guard at the start of `test_block_validity` and retain it through `check_coinbase_amount`, preserving the node's lock order. Add a deterministic concurrent transition regression test. The handler's preliminary `applied_hash` check is advisory and cannot replace serialization.

rabbitson87:
The independent blocker is repaired, with #101, #102 and #103 — and their repairs — merged in below this. 2 mutations: 1 killed by test, 1 rejected by the compiler.

## The guard

`test_block_validity` now takes `begin_chain_transition()` first, in the node's established order — admission permit, then the transition lock — through the same constructor every applier uses.

## Holding it to the end is a borrow, not a convention

Acquiring at the top and trusting the scope would leave "released early" as something only a comment forbids. So `validate_block_for_apply` and `check_coinbase_amount` both take `&ChainTransition<'_>`. Its fields are private, so `begin_chain_transition` is the only thing that can produce one: a caller that has not taken the lock cannot supply the argument, and one that releases it early stops compiling. `apply_block_admitted` already held the proof and now passes it down instead of merely naming it.

That is what the second mutation is. Inserting `drop(transition)` before `check_coinbase_amount` gives **E0382**, borrow of a moved value — recorded as invalid rather than counted as covered, because the compiler refusing it is the stronger result.

## The test

A transition is held open, a proposal is asked for on another thread, and the answer must not arrive; only once the transition is released may it. Deterministic in the direction that matters: without the guard the proposal answers immediately, so the first assertion fails **every** run rather than probabilistically. Reverting the whole fix and keeping the test kills it, which is mutation one.

## One thing found on the way

Threading the proof made `apply_block_admitted`'s `_transition` parameter genuinely read, and `clippy::used_underscore_binding` **ICEs** on that call — `slice_index_fail` inside `opt_span_lint`, rustc 1.95.0. Dropping the underscore is the right rename anyway (the parameter is now used), so nothing is worked around; recording it because a clippy ICE is easy to mistake for a broken tree.

## Verification

`cargo fmt --all --check` clean. `cargo clippy -p bitcoin-rs-node -p bitcoin-rs-rpc --lib --tests` clean under **both** `--no-default-features --features fjall` and default features. `node` and `rpc` suites green, 497 + 262 and the rest.

Unrelated and untouched: `crates/rpc/benches/{chaininfo,blocklookup}.rs` fail `clippy::as_conversions` under `--all-targets`. Those files are byte-identical to `origin/main` here, so the finding is main's and not this branch's.

gosunuts:
Current-main review: proposal mode is now implemented through the newer node-owned mining coordinator, which dry-runs the ordinary apply validation path and exposes proposal/submission through `MiningControl`. That makes this old stacked branch structurally superseded.

The important invariant from this PR — proposal validation must reuse the same validation path as connect rather than maintain a second validator — should remain, but it should be reviewed on the current `MiningCoordinator`/apply path. I would close this branch as superseded and address any remaining validation-depth gaps in a focused PR against current `main`.

metaphorics (commented):
Gate-and-merge hold: stack stall plus a blocking Gate 4 finding, head `8069868`.

Stall: this stack lands bottom-up and PRs 101–103 below it are held, so it cannot land first. Re-gate once it retargets.

Blocking, Gate 4 class 5 (contract drift): this PR adds required trait method `ChainControl::test_block_validity` (`crates/rpc/src/context.rs:741`, no default body), but `main` has since gained `impl ChainControl for NodeInvalidator` (`crates/rpc/tests/policy_contract.rs:1643`) without that method. Landing this as-is breaks the `policy_contract` test target's compilation on `main`. Resolution is an author decision (a default impl with explicit fail-loud semantics, or migrating the main-side implementor), so holding for the author rather than redesigning here.

Note: this PR also touches `DEVIATIONS.md` (+26), which `main` has since retired by deletion — flagging now so the landing path handles it (accept the retirement) instead of resurrecting the ledger. No merge on this oid until then.

metaphorics (commented):
## Merge breakage: `ChainControl::test_block_validity` is required, but main's `NodeInvalidator` doesn't implement it

This PR adds `fn test_block_validity(&self, block: &bitcoin::Block) -> Result<(), BlockRejectReason>` to `ChainControl` as a **required** method (no default body). Main has since added `crates/rpc/tests/policy_contract.rs`, which contains `impl ChainControl for NodeInvalidator` — a test fixture that implements only `invalidate_block`. On merge, that impl will fail to compile.

### Chosen resolution: implement the method for `NodeInvalidator`

**Why not a defaulted trait method.** `ChainControl` currently has two methods, both required. A default would make `test_block_validity` the first optional method in a trait whose contract is "the node installs implementations at startup" — the doc comment says so explicitly. A fail-loud default (`Err(BlockRejectReason("not installed"))`) would compile silently and read as "the feature is wired but refuses," which is worse than a compile error for a test fixture that already has `ApplyHandles` and can delegate to the real validator. The production impl (`RpcChainControl` in `run.rs`) already implements the method correctly; only the test fixture is missing.

**The fix** follows the existing `NodeInvalidator` style: delegate to `bitcoin_rs_node::apply::test_block_validity` (which is `pub`), map `ApplyError` into `BlockRejectReason` via `to_string()`. The production impl uses the private `reject_reason` mapper for Core-compatible tokens; the test fixture uses `Display` instead, which is sufficient for a test harness and avoids exposing `reject_reason`.

```diff
--- a/crates/rpc/tests/policy_contract.rs
+++ b/crates/rpc/tests/policy_contract.rs
@@ -16,21 +16,22 @@ use std::error::Error;
 use bitcoin_rs_mempool::eviction::mempool_min_fee_sat_per_kvb;
 use bitcoin_rs_mempool::{
     AdmissionOrigin, Mempool, MempoolEntry, MempoolGateway, MempoolLimits, MempoolObserver,
     MutationEnvelope, MutationOutcome, PolicyError, RbfError, RemovalReason, ReplacementCandidate,
 };
 use bitcoin_rs_node::reorg::{ReorgError, invalidate_block};
 use bitcoin_rs_node::{Network, NodeConfig, state::NodeState};
 use bitcoin_rs_primitives::encode::double_sha256;
 use bitcoin_rs_primitives::{Block, Hash256, OutPoint, Tx, TxIn, TxOut, Txid, consensus_bytes};
 use bitcoin_rs_rpc::context::{
-    ChainControl, ChainControlError, ChainHandles, Context, ContextHandles, IndexHandles,
+    BlockRejectReason, ChainControl, ChainControlError, ChainHandles, Context, ContextHandles,
+    IndexHandles,
     MempoolHandles, MiningHandles, NetworkHandles,
 };
 use bitcoin_rs_rpc::{Handler, RpcError};
 use bitcoin_rs_utxo::{BlockChanges, UtxoAdd};
 use sonic_rs::{JsonContainerTrait as _, JsonValueTrait, json};
 
 fn p2wpkh_script() -> Vec<u8> {
     // P2WPKH: `OP_0`, push-20, and a fixed 20-byte key hash.
     [vec![0x00, 0x14], vec![0x11; 20]].concat()
 }
@@ -1641,20 +1642,27 @@ struct NodeInvalidator {
 }
 
 impl ChainControl for NodeInvalidator {
     fn invalidate_block(&self, hash: Hash256) -> core::result::Result<(), ChainControlError> {
         invalidate_block(&self.handles, hash).map_err(|error| match error {
             ReorgError::UnknownBlock(_) => ChainControlError::UnknownBlock,
             ReorgError::CannotInvalidateGenesis => ChainControlError::Genesis,
             other => ChainControlError::Failed(other.to_string()),
         })
     }
+    fn test_block_validity(
+        &self,
+        block: &bitcoin::Block,
+    ) -> core::result::Result<(), BlockRejectReason> {
+        bitcoin_rs_node::apply::test_block_validity(&self.handles, block, None)
+            .map_err(|error| BlockRejectReason(error.to_string()))
+    }
 }
 
 fn open_regtest_state() -> Result<(NodeState, tempfile::TempDir), Box<dyn Error>> {
     let dir = tempfile::tempdir()?;
     let mut config = NodeConfig::default_for_network(Network::Regtest);
     config.data_dir = dir.path().join("node");
     config.p2p_listen.clear();
     let state = NodeState::open(config, None)?;
     Ok((state, dir))
 }
```

`git apply --check` passes against a worktree at this PR's head merged with main (where `policy_contract.rs` is present).

### Also: `DEVIATIONS.md` was deleted on main

Main retired `DEVIATIONS.md` by deletion. This PR adds §10 (+26 lines) to it. On merge this is a modify/delete conflict — accept main's deletion and drop the §10 addition. The content describes `reject_reason` mapping, which lives in `run.rs` comments already; if the rationale belongs somewhere, it belongs there, not in a file main has removed.

metaphorics (commented):
## Merge breakage: `ChainControl::test_block_validity` is required, but main's `NodeInvalidator` doesn't implement it

This PR adds `fn test_block_validity(&self, block: &bitcoin::Block) -> Result<(), BlockRejectReason>` to `ChainControl` as a **required** method (no default body). Main has since added `crates/rpc/tests/policy_contract.rs`, which contains `impl ChainControl for NodeInvalidator` — a test fixture that implements only `invalidate_block`. On merge, that impl will fail to compile.

### Chosen resolution: implement the method for `NodeInvalidator`

**Why not a defaulted trait method.** `ChainControl` currently has two methods, both required. A default would make `test_block_validity` the first optional method in a trait whose contract is "the node installs implementations at startup" — the doc comment says so explicitly. A fail-loud default (`Err(BlockRejectReason("not installed"))`) would compile silently and read as "the feature is wired but refuses," which is worse than a compile error for a test fixture that already has `ApplyHandles` and can delegate to the real validator. The production impl (`RpcChainControl` in `run.rs`) already implements the method correctly; only the test fixture is missing.

**The fix** follows the existing `NodeInvalidator` style: delegate to `bitcoin_rs_node::apply::test_block_validity` (which is `pub`), map `ApplyError` into `BlockRejectReason` via `to_string()`. The production impl uses the private `reject_reason` mapper for Core-compatible tokens; the test fixture uses `Display` instead, which is sufficient for a test harness and avoids exposing `reject_reason`.

This file arrives from main — rebase onto main (or merge it) first, then apply the patch; it was `git apply --check`-ed against exactly that merged tree. The same trial merge shows a dozen content conflicts beyond this method and `DEVIATIONS.md` (chain, mempool, mining, node, rpc), so this patch unblocks the trait impl, not the whole merge.

```diff
--- a/crates/rpc/tests/policy_contract.rs
+++ b/crates/rpc/tests/policy_contract.rs
@@ -16,22 +16,22 @@ use std::error::Error;
 use bitcoin_rs_mempool::eviction::mempool_min_fee_sat_per_kvb;
 use bitcoin_rs_mempool::{
     AdmissionOrigin, Mempool, MempoolEntry, MempoolGateway, MempoolLimits, MempoolObserver,
     MutationEnvelope, MutationOutcome, PolicyError, RbfError, RemovalReason, ReplacementCandidate,
 };
 use bitcoin_rs_node::reorg::{ReorgError, invalidate_block};
 use bitcoin_rs_node::{Network, NodeConfig, state::NodeState};
 use bitcoin_rs_primitives::encode::double_sha256;
 use bitcoin_rs_primitives::{Block, Hash256, OutPoint, Tx, TxIn, TxOut, Txid, consensus_bytes};
 use bitcoin_rs_rpc::context::{
-    ChainControl, ChainControlError, ChainHandles, Context, ContextHandles, IndexHandles,
-    MempoolHandles, MiningHandles, NetworkHandles,
+    BlockRejectReason, ChainControl, ChainControlError, ChainHandles, Context,
+    ContextHandles, IndexHandles, MempoolHandles, MiningHandles, NetworkHandles,
 };
 use bitcoin_rs_rpc::{Handler, RpcError};
 use bitcoin_rs_utxo::{BlockChanges, UtxoAdd};
 use sonic_rs::{JsonContainerTrait as _, JsonValueTrait, json};
 
 fn p2wpkh_script() -> Vec<u8> {
     // P2WPKH: `OP_0`, push-20, and a fixed 20-byte key hash.
     [vec![0x00, 0x14], vec![0x11; 20]].concat()
 }
 
@@ -1641,20 +1641,28 @@ struct NodeInvalidator {
 }
 
 impl ChainControl for NodeInvalidator {
     fn invalidate_block(&self, hash: Hash256) -> core::result::Result<(), ChainControlError> {
         invalidate_block(&self.handles, hash).map_err(|error| match error {
             ReorgError::UnknownBlock(_) => ChainControlError::UnknownBlock,
             ReorgError::CannotInvalidateGenesis => ChainControlError::Genesis,
             other => ChainControlError::Failed(other.to_string()),
         })
     }
+
+    fn test_block_validity(
+        &self,
+        block: &bitcoin::Block,
+    ) -> core::result::Result<(), BlockRejectReason> {
+        bitcoin_rs_node::apply::test_block_validity(&self.handles, block, None)
+            .map_err(|error| BlockRejectReason(error.to_string()))
+    }
 }
 
 fn open_regtest_state() -> Result<(NodeState, tempfile::TempDir), Box<dyn Error>> {
     let dir = tempfile::tempdir()?;
     let mut config = NodeConfig::default_for_network(Network::Regtest);
     config.data_dir = dir.path().join("node");
     config.p2p_listen.clear();
     let state = NodeState::open(config, None)?;
     Ok((state, dir))
 }
```

### Also: `DEVIATIONS.md` was deleted on main

Main retired `DEVIATIONS.md` by deletion. This PR adds §10 (+26 lines) to it. On merge this is a modify/delete conflict — accept main's deletion and drop the §10 addition. The content describes `reject_reason` mapping, which lives in `run.rs` comments already; if the rationale belongs somewhere, it belongs there, not in a file main has removed.

PR #106, opened 2026-08-22
fix(rpc): measure the getchaintxstats window the way Core measures it

rabbitson87:
## What

`getchaintxstats` measured its window with the wrong clock, and four other things about the call turned out not to be Core's either. This is the last item of the Core-compatibility RPC audit.

## The window was not a duration

`window_interval` was **the tip's raw timestamp minus the smallest raw timestamp in the window**. Bitcoin Core takes the difference of the **median-time-past of the window's two boundary blocks**.

Both halves of that are wrong, and they compound:

- **Raw block times are not ordered.** A miner only has to beat the median of the previous eleven blocks and may stamp up to two hours ahead, so the difference of two header timestamps is not an elapsed time. Every consensus rule that needs a clock uses the median for exactly this reason.
- **Minimum over the window, not the boundary.** One early-stamped block anywhere inside the window stretched the reported interval past the window that was actually asked for.

`txrate` divides by the interval, so it inherited both.

The interval now comes from the block tree — which holds every header it has accepted and is restored in full, so it answers for windows the in-process record log has never seen.

## Four more divergences the rewrite surfaced

| | before | Core, and now |
|---|---|---|
| window of 0 blocks | `window_interval: 0`, `window_tx_count: 0`, `txrate: 0.0` | all three omitted — those are measurements, and none was taken |
| window that did not advance | `txrate: 0.0` | omitted; dividing by a non-positive interval is not a rate |
| out-of-range block count | clamped silently | refused with `-8`; silently answering about a *different* window is the worse failure. The **default** count is still clamped, as in Core — the caller did not choose it |
| `blockhash` argument | ignored entirely, every answer was about the tip | selects the block the window ends at; a block off the active chain is refused rather than measured through a chain that does not exist |

The off-chain check compares the active chain's block *at that height* against the hash. A stale block keeps its height, so comparing heights would accept it.

## Counts: omitted, not zeroed

`txcount` and `window_tx_count` are now absent when this node cannot know them, which is Core's own signal for an unknown count (`m_chain_tx_count == 0`). They used to be `0` — and a real chain never has zero transactions, genesis alone contributes one, so a caller reading zero would conclude the chain was empty rather than that the node could not say.

Where "cannot know" differs from Core's is written down in **DEVIATIONS.md §10**: Core carries a cumulative count on every block index entry, this node has a durable one for the applied tip only plus a record log that is rebuilt empty on every open. Closing that gap means storing the count per height, which is a storage change and not this one.

`window_tx_count` is folded by height with the last record winning, so a height the log holds twice after a reorg is counted once.

## Housekeeping

`fold_block_records` loses `window_tx_count`, `earliest_window_time` and `tip_time`. Nothing reads an earliest in-window timestamp any more.

**That has a consequence for [#89](https://github.com/gosuda/bitcoin-rs/pull/89).** #89 replaces this fold with a binary-searched `BlockLog` and spends part of its machinery — and one of its fixtures, the one whose timestamps dip — on computing `earliest_window_time` quickly. That quantity is one Core does not compute, and after this PR it has no readers. Whichever of the two merges second should drop it and keep the window-count search. Flagging it rather than resolving it here: the two PRs are independent and the conflict is small and obvious.

This branch also adds `RpcError::InvalidParameter` (-8), which [#103](https://github.com/gosuda/bitcoin-rs/pull/103) adds identically on the `getblocktemplate` stack. The merge is a one-hunk dedupe.

## Tests

Twelve new, all mutation-audited.

The interval test asserts the median-time answer **and** asserts that the raw-timestamp difference is a different number, so the previous implementation could not have passed it. The median-time-past oracle is written out from Core's definition rather than read from the block tree — an oracle that called the tree could not disagree with the tree.

The rest: the rate derived from that interval, the zero-block window, the non-advancing window, the refused and clamped block counts (with the paired accept one below the limit), the `blockhash` argument including its unknown and off-chain rejections, and a window the log cannot cover.

Three existing tests changed and say why in their doc comments: two asserted a zero count where the field is now absent, and one asserted the raw-timestamp interval.

### Mutation audit

14 mutations, **14 killed**. Three needed a second pass: one did not build, one had been reformatted out from under its anchor, and one was run against a test that could not reach the line — recorded as invalid and re-run rather than counted as a pass.

## Stacking

Based on **#99** (`fix/rpc-chaintxstats-txcount`), which is based on **#96** (`feat/durable-chain-tx-count`). Order: **#96 → #99 → this**. #99 is the PR that made `txcount` and `time` survive a restart; this one builds directly on that resolution.

## Verification

WSL, `-j4`. `cargo fmt --all --check` clean. `cargo clippy -p bitcoin-rs-rpc --all-targets` clean apart from the `server.rs` / `tests/auth.rs` findings that are already on `main`. Full `rpc` and `node` suites green, re-run after a forced rebuild to rule out a stale binary.

## Summary by Sourcery

Make `getchaintxstats` report Core-compatible windows, counts, rates, and validation results.

New Features:
- Align `getchaintxstats` window selection and measurement with Bitcoin Core, including median-time-based intervals and support for historical active-chain block hashes.
- Report unknown transaction counts as omitted fields instead of misleading zero values.

Bug Fixes:
- Correct window intervals and transaction rates that previously used raw timestamps and the earliest timestamp in the window.
- Reject invalid explicit block counts and off-chain block hashes instead of silently clamping or measuring the wrong chain.
- Omit window measurements for empty or non-advancing windows and prevent incomplete logs from producing under-counted results.

Enhancements:
- Use block-tree data and synchronized chain state to provide consistent statistics across restarts, historical selections, and reorgs.
- Restrict durable cumulative transaction counts to the applied tip while deriving historical and window counts only from reliable sources.
- Add the Core-compatible `InvalidParameter` RPC error.

Tests:
- Add comprehensive coverage for median-time intervals, transaction rates, empty and non-advancing windows, parameter validation, block-hash selection, reorg handling, count durability, and incomplete logs.

Chores:
- Remove obsolete earliest-window timestamp aggregation from block-record folding.


metaphorics (commented):

Blocking findings after rebasing the live diff on current `main`:

1. `crates/rpc/src/handlers/chain.rs` reads `applied_tip` and `chain_tx_count` independently. Node writers publish them under `ApplyHandles::chain_transition`, with the tip store before the count update. A concurrent `getblockchaininfo` or `getchaintxstats` call can therefore pair one tip with another tip's cumulative count. The repair needs the RPC context to share that exact transition mutex, snapshot both values under it, and carry the paired count through the handler. That requires changes outside this PR's file allowlist (`rpc/context.rs` plus node runtime wiring), so it is blocking under the queue gate.

2. The optional `blockhash` path validates membership against the header-tree tip (`active_node_at_height`) rather than the applied tip. When headers are ahead or on another branch, it can accept a header-only block or reject a block on the applied branch. Resolve ancestry from `applied_tip.tip_id`.

3. The unknown-count fallback reports `txcount: 0` for an empty log at a genesis applied tip. Genesis contains a coinbase transaction; zero is a false count. Omit the field when neither the durable counter nor a complete log knows it.

4. `txrate` narrows `window_tx_count` and the MTP interval through `i32`, silently clamping valid larger operands and changing the rate. Convert the original `u64` and `i64` values to `f64` without that narrowing.

5. `window_tx_count` scans the whole `BlockLog` under its read lock and allocates a `HashMap` per request. Large windows block the writer and undo the prefix-sum bound already on `main`. Use binary-searched covered ranges and `tx_count_before` prefix subtraction.

6. A historical active-chain `blockhash` always omits `txcount`, even when a complete log prefix from genesis makes the cumulative count known. Restrict the durable shortcut to the applied tip, then use the complete prefix for historical endpoints.

The explicit block-count bound is not a finding: current Bitcoin Core intentionally rejects positive counts greater than or equal to the selected block height, while allowing explicit zero at genesis.

No repair commit was pushed because the atomic snapshot fix crosses files the PR did not change. Split or expand the PR deliberately, then re-run the full gate.

rabbitson87:
All six are repaired, on top of current `main`. 6 mutations, all killed.

**On the split-or-expand question: expanded, deliberately.** Finding 1 cannot be repaired inside the handler — the pair has to be published and read under one lock — and splitting it out would leave this PR blocked on a second one for a defect that is this PR's own. The added files are `rpc/context.rs`, `node/state.rs` (one accessor) and `node/run.rs` (one wiring line).

## 1. The pair is read under the transition

`Context` shares `ApplyHandles::chain_transition` and reads the applied tip and the cumulative count under it. That is the serialization Bitcoin Core gets from holding `cs_main` across the whole of `getchaintxstats`; this holds it for two loads.

**The test races and asserts a closed form.** The mover publishes tip-then-count in the node's order, under the node's lock, with the fixture giving block `h` exactly `h + 1` transactions — so the count through height `h` is `(h + 1)(h + 2) / 2` and the assertion is arithmetic rather than a comparison against a second read. The record log is left empty so a complete log cannot mask a torn pair. Removing the lock kills it.

## 2. Ancestry from the applied tip

`node_at_height_from(applied_tip.tip_id, …)`. A header inserted one above the applied tip is now refused; against the header tree's best chain it was accepted.

## 3. No zero at genesis

The `(height == 0).then_some(0)` branch is gone. `getchaintxstats_emits_core_shape_with_zero_blocks` asserted that zero; it now asserts the field is absent and says why.

## 4. `txrate` without narrowing

`u64_to_f64` / `i64_to_f64` — already on `main`, added there for this exact reason. Pinned at operands past `u32::MAX`: at 5e9 transactions over 10 seconds the narrowing reports 429,496,729.5 against the true 500,000,000, and the test asserts the two differ rather than only that the new one is right.

## 5. Prefix subtraction, not a scan

`window_tx_count(log, low, high)` binary-searches both boundaries and subtracts two prefix sums; `cumulative_tx_count_through` does the same for a chain total. No `HashMap`, no walk, and the log's read lock is held for two searches.

Checked the way `main` checks its own: swept against a walk of the log at every boundary pair, including the duplicate height a reorg leaves behind and the range that runs past the last record. An off-by-one at the lower bound dies on it.

## 6. Historical counts

The durable counter's shortcut is restricted to the applied tip — the only block it tracks — and a log reaching genesis answers for the rest. A fixture with the counter set to a distinctive number checks both directions: the tip still answers from the counter, and a block six deep answers from the log rather than being handed the tip's total.

## One consequence worth flagging

`chain_stats` loses its production caller. The window is measured between two median times now and may end at any block on the applied chain, which its signature cannot express — so the two narrower reads replace it. It stays in the tree for `benches/chaininfo.rs` and `chain_stats_matches_the_fold_it_replaced`, which are what make the prefix sums measurable at all, and its doc comment now says where the callers went.

## On the block-count bound

Noted, and unchanged: an explicit count is validated rather than clamped, positive counts at or above the selected block's height are refused, and an explicit zero is allowed.

## Verification

`cargo fmt --all --check` clean. `cargo clippy -p bitcoin-rs-rpc -p bitcoin-rs-node --lib --tests` clean under both `--no-default-features --features fjall` and default features. `rpc` and `node` suites green.

Mutations: the pair read without the transition — killed. Membership from the header-tree tip — killed. A partial log answering as a chain total — killed. `txrate` narrowed through 32 bits — killed. The durable counter answering for any block — killed. The window including its lower bound — killed.

Two of those survived their first run, against the tests that were already here; the tests above were written because they did.

gosunuts:
Current-main review: most of this PR's behavior appears to have been absorbed. `getchaintxstats` on `main` now accepts a blockhash, verifies active-chain membership, validates explicit window size, and derives the interval from block-tree median-time-past rather than the old raw/minimum timestamp logic.

Because the surrounding RPC/chain data model has changed since this branch was cut, I would not merge it as-is. Please compare the remaining edge cases against current `main`; if nothing material remains, this can close as superseded. If one or two semantics are still missing, they should be extracted into a small current-main PR.

metaphorics (commented):
Gate-and-merge hold: Gate 2, head `d09bc27`.
QA is clean: full-diff review found no behavioral defect (window matches Core `min(DEFAULT_WINDOW, tip_height)`, saturating MTP arithmetic, zero-interval guard, no surviving dropped-helper references). Scope is the one chaintxstats concern.

metaphorics:
@greptileai

metaphorics:
@sourcery-ai review

metaphorics:

metaphorics:

metaphorics:

metaphorics:
@cursor fix the pr comments and feedbacks + resolve conflicts.

metaphorics:

PR #107, opened 2026-08-22
feat(p2p): count what each connection actually moves

rabbitson87:
## What

`getpeerinfo` answered seven fields with zeros and one with the wrong address, and nothing in the response said so. This is the last Class B finding from the Core-compatibility audit — the class where *the field is a placeholder that looks like data*.

| field | was | now |
| --- | --- | --- |
| `bytessent`, `bytesrecv` | `0` | what the connection moved |
| `lastsend`, `lastrecv` | `0` | when it last moved something |
| `addrbind` | a copy of `addr` | the local end of the connection |
| `timeoffset` | `0` | the peer's clock against ours |
| `pingtime`, `minping` | `0.0` | **absent** |
| `getnetworkinfo.timeoffset` | `0` | the median of the peers' offsets |
| `getnetworkinfo.version` | `10000` | Core's `CLIENT_VERSION` from the package version |

## Why a wrapper and not a hook

A connection's bytes move through **three separate pieces of code**: the handshake, the reader loop, and a writer thread that owns a `try_clone`d socket. A counter added at any one of them reports a fraction of the traffic as though it were all of it.

`CountingStream<S>` wraps the socket *before* the handshake and counts in `Read`/`Write`, so every byte is counted wherever it is written from. Its `try_clone` shares the same `Arc<PeerCounters>`, so the writer thread lands in the same place as the reader.

It counts what the socket actually took, not what it was offered — a short write is the case that separates the two, and there is a test for it.

## Omitted, not faked

`pingtime`, `minping` and `pingwait` are gone from the response. This node never sends a ping, so it has never measured a round trip, and Core omits all three until it has one (`rpc/net.cpp` guards each on a real measurement). `0.0` there is not merely wrong — it is the best latency a peer could possibly have, so it reads as the healthiest connection in the list.

## timeoffset is a median

Per peer it is the peer's declared time in its `version` minus local time at that moment, computed once and never revisited — Core's definition.

For the node as a whole, `getnetworkinfo.timeoffset` is the **median** of those. The figure exists to tell an operator that their own clock is wrong, so a single peer claiming an absurd time must not be able to raise that alarm. There is a test with an outlier three orders of magnitude out that does not move the answer.

## What is still a placeholder

Written down in **DEVIATIONS.md §11** rather than left silent: `bytessent_per_msg` / `bytesrecv_per_msg`, `minfeefilter`, `addr_processed` / `addr_rate_limited`, `permissions`, `relaytxes`, the reduced `connection_type` set, and `getnetworkinfo`'s `localaddresses` and `networks`.

`presynced_headers` / `synced_headers` / `synced_blocks` report `-1`, which is Core's own value for a peer it has no figure for, so they are not on that list.

## Tests

Eleven new.

- **stream** — a short write counts what the socket took; reads accumulate; an empty read is not activity and stamps no time.
- **`PeerInfo`** — the offset carries its sign in both directions; the bind address is kept apart from the peer address.
- **RPC** — the byte counts and timestamps (driving real traffic through a `CountingStream`, so the test covers the wiring and not the rendering), the bind address, the offset, the ping omission, the median with an outlier, and the no-peers case.

### Mutation audit

12 mutations, **12 killed**. Two survivors on the first pass were real fixture faults, not false alarms:

- the empty-read assertion compared two timestamps taken in the same second, so a stamp-on-every-read bug compared equal;
- nothing covered `try_clone`, which is precisely where an under-count would come from.

Both are closed in the second commit and re-audited.

## Verification

WSL, `-j4`. `cargo fmt --all --check` clean. Workspace tests green. `cargo clippy --workspace --all-targets` clean apart from the findings already on `main` (`server.rs`, `tests/auth.rs`, `apply.rs`, `mempool/orphan.rs`, `mempool/standardness.rs`).

## Summary by Sourcery

Expose accurate connection traffic and timing metadata in peer and network RPC responses while removing unsupported placeholder measurements.

New Features:
- Report per-connection bytes sent and received, activity timestamps, local bind addresses, and peer clock offsets through `getpeerinfo`.
- Compute `getnetworkinfo.timeoffset` from qualifying outbound peer samples and derive its reported client version from the package version.

Bug Fixes:
- Remove misleading zero-valued ping metrics from `getpeerinfo` when no latency measurement exists.
- Correct `addrbind` to identify the local connection endpoint instead of duplicating the peer address.

Enhancements:
- Add shared socket-level traffic accounting across handshakes, readers, and cloned writer streams.
- Align network RPC metadata and time-offset behavior more closely with Bitcoin Core compatibility expectations.

Tests:
- Add coverage for traffic counting, short writes, cloned sockets, timestamps, bind addresses, signed offsets, ping omission, version reporting, and median time-offset behavior.

Chores:
- Document remaining intentionally unsupported or placeholder Core-compatible fields.


metaphorics (commented):

Blocking metrics defect in `crates/rpc/src/handlers/network.rs:164`: the even-peer median uses `lower.saturating_add(upper) / 2`. When both middle peer offsets are near `i64::MAX` or `i64::MIN`—values peers can supply in version timestamps—the sum saturates and the reported median is roughly half the correct magnitude.

Use an overflow-safe signed average that preserves truncation toward zero, for example quotient/remainder decomposition: `lower / 2 + upper / 2 + (lower % 2 + upper % 2) / 2`. Add positive- and negative-overflow pair tests.

Byte accounting, shared cloned-socket counters, endpoint selection, handshake coverage, and RPC field mapping passed review after local conflict resolution. No branch changes were pushed.

rabbitson87:
Rebased on current `main`, and I found three defects in my own `timeoffset` work while re-reading Core's `node/timeoffsets.cpp` against it. Fixed here rather than waiting for review to find them.

**Inbound peers were being sampled.** Core takes offsets from outbound peers only, and states the reason at the call site in `net_processing.cpp`:

> Don't use timedata samples from inbound peers to make it harder for others to create false warnings about our clock being out of sync.

Anyone can open an inbound connection and declare any time they like. Medianing over every peer hands that person the operator's belief about whether this machine's clock is wrong — which is the entire purpose of the field, so it is the whole of the defect. The regression puts nine inbound peers an hour ahead against five outbound peers at ten seconds: the inbound crowd wins a median over all peers, and must not move this one.

**Fewer than five samples answered anyway.** `TimeOffsets::Median` returns zero below five offsets — *"Only calculate the median if we have 5 or more offsets"*. Four strangers all an hour ahead were reported as an hour, which reads as a confident finding about the local clock. Now four gives zero and five gives the offset, so the zero is the floor rather than an inability to report.

**Even sample counts were averaged.** Core indexes `sorted[size / 2]` whatever the parity, and declines to interpolate — *"approximate median is good enough, keep it simple"*. My code averaged the two middle values, and my comment claimed that was Core's behaviour. It is not; averaging answers a number no peer reported and differs from Core on every even count.

## One difference left, now named

Core medians a rolling deque of the last 50 offsets it sampled, which outlives the connections that produced them. This medians the peers connected now. The two agree while the peer set is stable and diverge after churn, where Core still remembers a departed peer's sample. Recorded in `DEVIATIONS.md` §10 rather than left as an accident.

## Verification

`cargo fmt --all --check` clean. `cargo clippy -p bitcoin-rs-rpc -p bitcoin-rs-p2p -p bitcoin-rs-primitives --lib --tests` clean. `rpc`, `p2p`, `primitives`, `node` suites green.

The `DEVIATIONS.md` merge renumbered this branch's section from §11 to §10; `main` landed a §9 in the meantime.

rabbitson87:
The overflow is gone, and I had already removed it before reading your review — for a different reason, which is worth saying plainly since it means we arrived at the same line from opposite directions.

## The averaging is not overflow-safe, it is absent

You asked for a signed average that survives the extremes. Re-reading `node/timeoffsets.cpp` against this branch I found the averaging was wrong on its own terms: Core indexes `sorted[size / 2]` whatever the parity and declines to interpolate — *"approximate median is good enough, keep it simple"*. My comment claimed averaging was Core's behaviour. It is not, and averaging answers a number no peer reported on every even sample count.

So the fix is `sorted[len / 2]`, which removes the arithmetic rather than making it safe. That satisfies your finding as a side effect, and I would rather not leave "satisfied as a side effect" as the only evidence.

**`extreme_offsets_are_reported_at_their_own_magnitude`** is that evidence: six samples with both middle values at `i64::MAX`, `i64::MIN`, and one either side, and the reported offset must be a sample some peer actually sent and must not be that sample halved. Confirmed by mutation — restoring `lower.saturating_add(upper) / 2` fails it.

## Two more defects in the same function

Found in the same pass, both fixed in the commit before this one:

- **Inbound peers were being sampled.** Core takes offsets from outbound peers only, and says why at the call site: *"Don't use timedata samples from inbound peers to make it harder for others to create false warnings about our clock being out of sync."* This is the same attack your finding describes, reached without needing extreme values — anyone who can open inbound connections could move the reported offset. The regression puts nine inbound peers an hour ahead against five outbound at ten seconds.
- **Fewer than five samples answered anyway.** Core returns zero below five offsets. Four strangers all an hour ahead were reported as an hour.

## One difference left, named rather than implied

Core medians a rolling deque of the last 50 offsets, which outlives the connections that produced them; this medians the peers connected now. They agree while the peer set is stable and diverge after churn. Recorded in `DEVIATIONS.md` §10.

## Verification

`cargo fmt --all --check` clean. `cargo clippy -p bitcoin-rs-rpc --lib --tests` clean. `rpc` and `p2p` suites green.

gosunuts:
Current-main review: this one still looks materially useful. I do not see the `PeerCounters`/connection accounting from this PR on `main`, so unlike several older RPC PRs this is not simply superseded.

It does need a fresh rebase/review, though. P2P lifecycle ownership has changed significantly since this branch was cut, and #215/#217 make it especially important that connection counters remain owned by `p2p` rather than creating another node-side peer-lifecycle invariant. The `CountingStream` approach is still a good fit for that boundary if it survives the rebase cleanly.

metaphorics (commented):
Gate-and-merge: QA minor, head `2a005bc`. One-concern (per-connection traffic accounting + RPC presentation), no lockfile/ledger issues.

Minor, Gate 4 class 1 without plausible reachability: `record_sent`/`record_recv` in `crates/p2p/src/counters.rs:62-66` use wrapping `fetch_add`, so past `u64::MAX` bytes a counter resets to a small value. `u64::MAX` bytes on one connection (~16 EiB) is not a reachable input, so this is not blocking — but saturation is strictly more honest for a cumulative counter, and the fix stays inside the PR file set. `maintainerCanModify: false`, so delivering as a patch rather than pushing:

```diff
--- a/crates/p2p/src/crates/p2p/src/counters.rs
- counter.fetch_add(bytes, Ordering::Relaxed)  // wrapping
+ counter.fetch_update(Ordering::Relaxed, Ordering::Relaxed, |v| v.checked_add(bytes).or(Some(u64::MAX))).ok();
```

(applied to both send and receive counters), plus a boundary test seeding each counter near `u64::MAX` and asserting saturation. Suggested, not required for landing.

Gate 2 governs the merge: holding for green checks on this head. No merge on this oid until then.

metaphorics (commented):
Gate-and-merge hold: Gate 2, head `2a005bc`.
QA minor only: a separate comment carries a suggested saturating-counter patch (unreachable `u64::MAX` wrap, no plausible input). One-concern; no other findings.

metaphorics:
@greptileai

metaphorics:
@sourcery-ai review

metaphorics:

metaphorics:

metaphorics:

PR #108, opened 2026-08-23
fix(rpc): report getchaintips against the chain the node has connected

rabbitson87:
## What

`getchaintips` was answering about the wrong chain. Two defects, both of which appear the moment headers run ahead of applied blocks — which is every minute of initial sync.

## The active tip was the best *header*, not the connected block

`ctx.chain_tip` is the block tree's best header. `ctx.applied_tip` is the block this node has actually connected. Bitcoin Core asks `ActiveChain().Tip()`, which is the connected one.

So a header the node had never validated reported itself as `active`.

## And the connected tip was not listed at all

Only leaves were enumerated. **During initial sync the applied tip is never a leaf** — the header tree already describes its descendants — so the one chain the node had validated was missing from the response while an unvalidated header claimed its place.

Core inserts `active_chain.Tip()` into the tip set explicitly, leaf or not. So does this now.

## `NodeStatus` is not a validation record

This is the part that made the status mapping wrong rather than merely incomplete.

`BlockTree::insert_node` stamps whichever node carries the most work `Active` and demotes the one it displaced to `Stale` (`tree.rs::publish_tip_if_best`). On a header-first node that means **every accepted header on the best chain is `Active` while its block sits unconnected**. It is a position marker in the best *header* chain, not a record of how far validation got — which is exactly what Core's `status` is derived from.

The mapping is now built from what is decidable here: a block is connected exactly when it is the applied tip or an ancestor of it, and a leaf is never an ancestor of anything, so every leaf but the applied tip is unconnected *now*. The only question left is whether it ever was, and `Stale` — displaced from the best chain — is the one signal the tree keeps about a branch it stopped following.

| status | now emitted for |
| --- | --- |
| `active` | the applied tip |
| `invalid` | `NodeStatus::Invalid` |
| `valid-fork` | `NodeStatus::Stale` |
| `headers-only` | every other leaf |
| `valid-headers` | never — this node applies a block as it arrives, so "data present, never validated" does not exist |

**DEVIATIONS.md §12** carries the table and names the residual honestly: a branch that was *followed* but never *connected* — a header that was briefly the most work and lost a race — reads as `valid-fork` where Core would say `headers-only`. Telling those apart needs a per-block record of validation depth that the tree does not keep. It is a narrower error than the one it replaces.

## Ordering

Core sorts by height descending with no special place for the active tip; this sorted the active tip to the front. A fork longer than the active chain is precisely what an operator runs this command to find, and putting the active tip first buried it. Ties now break on the hash, so the order is stable across calls instead of following slab layout.

## Tests

Five new, plus two existing fixtures updated to set the applied tip — which is what a node that has connected a block does.

- the applied tip is listed while headers run ahead (fails on the old code: it returned one entry, not two)
- the header tip reads `headers-only`, with the distance it leads by as `branchlen`
- the active tip's branch length is zero
- ordering is height-descending
- a branch left behind by a reorg reads `valid-fork` with its branch length — the fixture drives a real displacement through `insert_node` and asserts the tree actually demoted the old tip before testing the RPC

### Mutation audit

9 mutations, **8 killed, 1 equivalent**. The equivalent one removed the `is_active` shortcut for `branchlen`: `compute_branchlen` walks from the leaf to the first node on the active chain, and for the applied tip that is itself, so it returns 0 either way. Recorded as equivalent rather than counted as a survivor — the shortcut is redundant, not untested.

Two mutations first came back as anchor misses because rustfmt had reflowed the code they targeted; re-run against the formatted source, both killed.

## Verification

WSL, `-j4`. `cargo fmt --all --check` clean. `cargo clippy -p bitcoin-rs-rpc --all-targets` clean apart from the `server.rs` / `tests/auth.rs` findings already on `main`. Full `rpc` suite green, re-run after a forced rebuild to rule out a stale binary.


metaphorics (commented):

Blocking status defects:

1. `crates/rpc/src/handlers/chain.rs:257-263`: `NodeStatus::Stale` does not prove a block was connected. The header tree marks a displaced best-work header stale even when the node never received or applied its block. Reporting such leaves as `valid-fork` overstates validation. With the current data model, map non-applied stale leaves to `headers-only`; reserve `valid-fork` until the tree records evidence that a branch was connected.

2. `crates/rpc/src/handlers/chain.rs:252-265`: applied-tip identity takes precedence over `Invalid`. `invalidate_block` marks the subtree invalid under the tree lock before it disconnects and republishes `applied_tip`, so a concurrent call can label the invalid old tip `active`. Make `Invalid` precedence explicit and emit `active` only when the snapshotted applied node is not invalid.

Applied-tip selection and branch-length arithmetic otherwise passed review after local conflict resolution. No branch changes were pushed.

rabbitson87:
Both blocking findings repaired. My apologies for the delay on these — I had checked for reviews with a broken shell loop, read the resulting 404 as "no review yet", and self-audited instead of reading what you wrote. The concurrency finding is one my own audit missed, and worse, I had just committed a comment arguing the ordering was safe.

## 1. `Stale` does not prove a block was connected

You are right, and the existing test was pinning the overstatement: `a_branch_left_behind_by_a_reorg_is_a_valid_fork` builds its fixture by inserting *headers* and never applies a2's block, then asserted `valid-fork`. That is exactly the case where the claim is false.

**`valid-fork` is now never emitted.** Reporting `headers-only` where Core would say `valid-fork` under-states; the reverse over-states, and only one of those is safe for the person reading `getchaintips` to find out what this node verified. The test is renamed and now asserts `headers-only` with the reason.

I considered reaching for evidence rather than dropping the status — the block-record log holds an entry for every block this process applied, so a stale leaf found there really was connected. I did not do it: the log is rebuilt empty on every open, so it answers for this process only, and a status that is right until you restart is the kind of field-shaped placeholder this campaign has been removing. Recorded in `DEVIATIONS.md` §10 as what emitting `valid-fork` would need.

## 2. `invalid` now outranks the applied tip

Confirmed at the source: `reorg::invalidate_block` calls `tree.invalidate_subtree(root)` under the write lock and releases it at the end of that block, and `execute_loaded_plan` disconnects and republishes `applied_tip` afterwards. A call landing between the two sees both facts and, deciding on the tip first, labels a block the node has just rejected as the chain it is following.

**No ordering closes the window** — the two are published separately, so my earlier "tree guard, then tip" argument covers the connect path and not this one. What removes the wrong answer is preferring the fact that cannot go stale: a block that is invalid was invalid before the call and stays invalid after it, while "this is the applied tip" is precisely what the invalidation is in the middle of changing. `Invalid` is matched first, ahead of the tip comparison.

The regression reproduces the window exactly rather than racing for it: invalidate the subtree, leave the applied tip where invalidation found it, and assert the block reports `invalid` and that nothing reports `active`. Deterministic, and it fails every run against the previous ordering.

## Verification

`cargo fmt --all --check` clean. `cargo clippy -p bitcoin-rs-rpc --lib --tests` clean. `rpc` suite green.

The branch also carries a merge of current `main`; `DEVIATIONS.md` renumbered from §12 to §10 in the process.

gosunuts:
Current-main review: most of the substantive fix is already present. `main` now explicitly includes the applied tip even when headers run ahead and distinguishes the applied tip from header-only/stale leaves.

There is still one visible semantic difference worth isolating: current `main` sorts the active tip first, while this PR argues for Core's height-descending ordering. I would not keep this whole stale branch open for that residual. Prefer closing it as mostly superseded and, if Core-compatible ordering is still desired, submit that as a small focused change against current `main`.

metaphorics (commented):
Gate-and-merge hold: Gate 2, head `82fa556`.
QA is clean: full-diff review found no reachable wrong result (guard-before-tip ordering, Invalid-ahead-of-active mapping, saturating branchlen, deterministic height+hash sort). Single-file, one concern.

PR #109, opened 2026-08-23
feat(rpc): answer getmininginfo with the numbers a miner mines against

rabbitson87:
## What

`getmininginfo` was missing five of Bitcoin Core's fields, and answered a sixth with a different quantity than the one it names.

| field | before | now |
| --- | --- | --- |
| `bits` | **absent** | the tip's nBits |
| `target` | **absent** | the tip's target |
| `blockmintxfee` | **absent** | the feerate floor selection applies |
| `next` | **absent** | `{height, bits, difficulty, target}` of the block being mined |
| `currentblockweight`, `currentblocktx` | a fresh mempool selection, always present | the last assembled template, absent until there is one |
| `warnings` | `""` | `[]` |

## `next` is the number the call exists to report

A miner asks `getmininginfo` what it is about to work against. The tip's difficulty is history; **the next block's is the answer**, and it was the one field not there.

It comes from `expected_next_bits` — the same function header validation asks — at the same candidate timestamp `getblocktemplate` builds with. So this can never advertise a difficulty the node would then reject, and the two RPCs cannot disagree.

The handler was already reading the tip's `bits` and throwing everything but `difficulty` away, so `bits` and `target` cost nothing to add.

## The block-assembly fields described something else

`currentblockweight` and `currentblocktx` name *the last assembled block*. They were running a **fresh package selection over the whole mempool on every call**, under the mempool read lock, and reporting that.

Two separate problems: it answers a different question (an estimate of what a template *would* contain, not what the last one *did*), and it pays a mempool-sized cost for an informational RPC — the same shape as the `getrawmempool` finding earlier in this audit.

They now read the template this node cached when it last assembled one, and are **absent** until one exists, which is what Core does (`BlockAssembler::m_last_block_weight` is an optional). A populated mempool is not an assembled block.

The weight includes the reserve held back for the header, the transaction count and the coinbase, because that is the weight the assembled block will have — Core's field documents exactly that.

## Smaller

`warnings` becomes an array. The string is Core's deprecated shape, kept behind `-deprecatedrpc=warnings`.

`blockmintxfee` reports Core's `DEFAULT_BLOCK_MIN_TX_FEE`, which is **1 satoshi per kvB** — read from the vendored source rather than from memory, which would have said 1000.

Not added: `signet_challenge`, which this node does not carry. Noted rather than faked.

## Tests

Six new, two replaced.

The replaced pair pinned the mempool-estimate reading (`currentblockweight_reflects_mempool_when_populated`, `currentblocktx_zero_when_mempool_empty`); their successors assert the fields are **absent** with a populated mempool and no template, and correct after `getblocktemplate` assembles one. Both say why in their doc comments.

New: the tip's bits and target, the next block's height and target (equal to the tip's, because regtest never retargets — and the test asserts that equality rather than a literal), the array-shaped warnings, and the feerate floor.

### Mutation audit

12 mutations, **12 killed**. Three needed a second pass: two anchors had been reflowed by rustfmt, and one matched in two places (the template-cache read appears in `getblocktemplate` too) — recorded as anchor misses and re-run against the formatted source rather than counted as passes.

## Stacking

Based on **#105**, the top of the `getblocktemplate` stack, because two of its fields need what that stack built: `next` needs `expected_next_bits` (split out in #103), and `currentblockweight` needs the assembled-template cache (added in #103). Order: **#101 → #102 → #103 → #105 → this**.

## Verification

WSL, `-j4`. `cargo fmt --all --check` clean. `cargo clippy -p bitcoin-rs-rpc --all-targets` clean apart from the `server.rs` / `tests/auth.rs` findings already on `main`. `rpc` and `node` suites green, re-run after a forced rebuild to rule out a stale binary.

## Summary by Sourcery

Bring `getmininginfo` in line with Bitcoin Core by reporting the values miners need for the current and next block.

New Features:
- Expand `getmininginfo` with tip and next-block mining targets, the applicable minimum transaction fee, and cached block-assembly statistics.

Bug Fixes:
- Report mining information consistently from a single chain snapshot and align next-block difficulty with the validation rules and template candidate time.
- Stop deriving current block statistics from a fresh mempool selection and omit them until a template has been assembled.
- Return `warnings` in the current array-shaped RPC format.

Enhancements:
- Refactor chain-dependent mining information calculation to share one tip and tree view.

Tests:
- Add coverage for tip and next-block targets, fee floor, warning shape, template-backed block statistics, and consistency while the chain changes.


metaphorics (commented):

Dependencies: ancestor PRs #101, #102, #103, and #105 are held.

Independent blockers:

1. `crates/rpc/src/handlers/mining.rs:381`: the handler captures one tip but reloads applied state for `blocks`, network hash rate, and median time. A concurrent transition can combine one tip's bits/target with another tip's height/MTP/hashrate, including at a difficulty boundary. Derive every chain-dependent field from the captured `TipSnapshot` and its `tip_id` under one corresponding tree view.

2. `crates/rpc/src/handlers/mining.rs:450-451`: `currentblockweight` adds `DEFAULT_BLOCK_RESERVED_WEIGHT`, which is a selection reserve, not serialized block weight. It systematically reports unused policy headroom as consumed weight. Store and report the actual completed candidate block weight, including concrete coinbase and overhead; do not use the reserve as a proxy.

rabbitson87:
Blocker 1 is repaired. Blocker 2 I am pushing back on, with the source — the number this PR prints is the number Core prints. #101, #102, #103 and #105 are merged in below this, with their repairs. 4 mutations, all killed.

## 1. One tip, one tree view

The handler captured a tip and then reloaded applied state for `blocks`, for the hash rate, and for the median time behind the next difficulty. Every chain-dependent field is now produced by a single `ChainFields::read` from the captured `TipSnapshot` under a single tree guard, or not at all. `estimate_network_hashps` takes the tip and the tree view instead of loading its own, so it cannot describe a different block than the difficulty printed beside it.

**The test races and its assertions are exact.** Four hundred calls against a thread publishing successive tips, and two pairings must hold in every answer:

- `next.height` is one past `blocks`.
- `networkhashps` is non-zero **exactly** when `blocks` is — at genesis the window has no second block to span, so a zero rate beside a positive height, or a positive rate beside height zero, is a torn read and nothing else.

Both endpoints are checked before the race starts, so a fixture that stopped discriminating would fail loudly rather than pass vacuously. Restoring either reload kills it.

## 2. `currentblockweight` — Core reports the reserve

This one I think is wrong, and the vendored tree is the reason.

`node/miner.cpp` starts `nBlockWeight` at `block_reserved_weight` (`DEFAULT_BLOCK_RESERVED_WEIGHT{8000}`, `policy/policy.h`), adds each selected transaction, and snapshots `m_last_block_weight` **before** it builds the coinbase — so the coinbase's real weight never enters the figure at all. `rpc/mining.cpp` prints that snapshot, and its own help text is explicit:

> "The block weight (including reserved weight for block header, txs count and coinbase tx) of the last assembled block (only present if a block was ever assembled)"

So it is an upper bound by construction, not a measurement, and it is the one Core has always published. Storing the completed candidate's serialized weight instead would make this node print a smaller number than Core for the identical template, and a miner comparing the two would read the difference as a bug in one of them. Since Core-compatibility is what this PR is for, that is a deviation rather than a fix.

The figure is unchanged. What is added is the citation — at `assembled_block_totals` and in `getmininginfo_reports_the_weight_core_reports`, which pins an empty template at exactly the reserve — so the next reader does not mistake the change for a correction. If you want the measured weight anyway, say so and it goes in `DEVIATIONS.md` as a deliberate departure rather than in silently.

## What the tests do not cover

A tear specifically between the MTP read and `expected_next_bits` is not observable on regtest: there is no retargeting, so the next target is the same whichever guard it was read under. The fix is structural — `ChainFields::read` takes the guard by reference, so acquiring a second one requires changing the signature — but I am recording the gap rather than implying the race test covers it. A mainnet retarget-boundary fixture would; that is a bigger fixture than this PR should grow.

## Verification

`cargo fmt --all --check` clean. `cargo clippy -p bitcoin-rs-rpc -p bitcoin-rs-node -p bitcoin-rs-mining --lib --tests` clean under both `--no-default-features --features fjall` and default features. `rpc`, `node` and `mining` suites green.

Mutations: `blocks` reloading applied height — killed. Hash rate reloading the applied tip — killed. Weight dropping the reserve — killed. Count including the coinbase — killed.

gosunuts:
Current-main review: this is now superseded by the newer mining stack. `main` has a node-owned mining coordinator with generation-keyed candidate caching and exposes mining facts from the current candidate lifecycle rather than rebuilding the old `getblocktemplate` stack.

The RPC fields this PR cared about should be checked against the current `MiningInfo`/`MiningControl` projection, but the branch itself is no longer a good merge base. I would close it as superseded and file only any remaining field/value mismatch against current `main`.

metaphorics (commented):
Gate-and-merge hold: stack stall only, head `5ed53dc`. QA is clean.

Stall: this stack lands bottom-up and PRs 101–105 below it are held, so it cannot land first. Re-gate (including CI) once it retargets to `main`.

QA note, for the record: the one candidate finding (hard-coded `blockmintxfee` 1 sat/kvB) was checked and refuted — the constant equals Core's `DEFAULT_BLOCK_MIN_TX_FEE`, the mempool acceptance floor already guarantees no admitted transaction sits below it, and package-level inclusion matches Core's own semantics, so the reported floor is accurate. No finding; no test debt (single handler field off an existing fixture path). No merge on this oid until the stack lands below it.

metaphorics:
@greptileai

metaphorics:
@sourcery-ai review

metaphorics:

metaphorics:

metaphorics:

metaphorics:
Closing as superseded: the getmininginfo next-block semantics this PR delivers landed on main via #101 (finality evaluated at next block height, commit ae62ad51) and 585ef727 (tip's compact target at top level). Its base branch feat/mining-package-selection is closed as #102 (superseded by main's overhaul). Branch preserved.

PR #110, opened 2026-08-23
fix(rpc): analyse descriptors instead of guessing about them

rabbitson87:
## What

`getdescriptorinfo` never parsed the descriptor it was handed, and `deriveaddresses` never derived anything.

## The dangerous one

```json
"issolvable": false,
"hasprivatekeys": false
```

Both were **constants**. Any descriptor carrying keys got the first one wrong, and the second is wrong in the direction that matters: someone checking whether a descriptor they were about to share carried their private key was told **no** — while the response handed the descriptor straight back, `xprv` and all.

Core returns the *canonical* form, which is the parse re-serialized and therefore public by construction. That is now what this returns, and there is a test asserting the `xprv` does not come back out.

## The quiet one

`deriveaddresses` returned `[]` for every descriptor except `addr(...)`, behind a `TODO(miniscript)`.

An empty array is not "this node did not look" — it reads as *this descriptor describes no addresses*, and a caller cannot tell the two apart. A wallet enumerating receive addresses gets silence and no error.

## Both were a parse away

`miniscript` was already a workspace dependency and `bitcoin-rs-wallet` already used it. `descriptor::analyse` and `descriptor::derive_addresses` do the parse in the crate that owns the dependency; the handlers stay thin.

What follows from parsing:

- `issolvable`, `hasprivatekeys`, `isrange` — computed. `hasprivatekeys` comes from the `KeyMap` `parse_descriptor` hands back, which is exactly Core's `provider.keys.size() > 0`, not a guess at the text.
- `multipath_expansion` — reported when the descriptor has one.
- ranged descriptors derive one address per index, inclusive at both ends.
- multipath descriptors derive one array per expansion, in specifier order, as Core shapes them.
- the range argument is **required** for a ranged descriptor and **refused** for a fixed one.

## `addr()` and `raw()`

miniscript does not model them — they name an output without saying how to spend it — so they are handled apart from it and reported `issolvable: false`, which is what they are.

What is inside the brackets is now checked. `addr(x)` used to be accepted and echoed back with a checksum; `x` is not an address, and it is now refused the way Core refuses it. `raw()` derives an address when the script has one and says so when it does not.

## Error codes

Core's split, kept: a descriptor that does not parse is **-5**; a range that does not match a descriptor that parsed fine is **-8**. A client can tell "I sent you nonsense" from "I asked the wrong thing about something valid".

This adds `RpcError::InvalidParameter` (-8), which [#103](https://github.com/gosuda/bitcoin-rs/pull/103) and [#106](https://github.com/gosuda/bitcoin-rs/pull/106) each add identically on other branches — the merge is a one-hunk dedupe — and `RpcError::InvalidAddressOrKey` (-5), which is new here.

## Tests

Nine. The four that pinned the old behaviour are replaced and say why in their doc comments: one asserted the echoed descriptor, three asserted the empty array.

New coverage: the checksum is of the input while the descriptor is of the parse; an address descriptor is unsolvable and a key descriptor is not; a private key is reported and not echoed; unparseable text is refused with -5; an `addr()` descriptor derives its own address; a key descriptor derives a real one; a ranged descriptor derives one address per index, all distinct; the range argument must match.

### Mutation audit

12 mutations, **12 killed**. One survivor was a real coverage gap, not a false alarm: the fixed-descriptor range check was tested with an `addr()` descriptor, which takes the early return and never reaches the miniscript path's own check — so that arm was untested. Closed in the second commit and re-audited.

Two mutations had to be rewritten to be judged at all: one failed to compile and one tripped a compiler ICE. Recorded as invalid and re-run in a form that builds, rather than counted as passes.

## Left alone

`walletcreatefundedpsbt` still returns an empty PSBT with `fee: 0` while reporting success, and `scantxoutset` unspents still omit Core's `blockhash`. Both are real, both are separate from this, and neither is quietly fixed here.

## Verification

WSL, `-j4`. `cargo fmt --all --check` clean. `cargo clippy -p bitcoin-rs-rpc -p bitcoin-rs-wallet --all-targets` clean apart from the `server.rs` / `tests/auth.rs` findings already on `main`. Tests green, re-run after a forced rebuild.

One note on running them: `cargo test -p bitcoin-rs-wallet` **alone** fails `psbt_roundtrip` with "portable script backend cannot verify this non-taproot spend". That is pre-existing on `main` — the wallet crate does not enable `bitcoin-rs-consensus/kernel` by itself, so testing it in isolation gets the portable backend. It passes whenever a crate that does is in the same build, which is how CI builds the workspace.

## Summary by Sourcery

Analyze descriptors through miniscript and provide Bitcoin Core-compatible descriptor metadata, validation, canonicalization, and address derivation.

New Features:
- Implement descriptor parsing and analysis for `getdescriptorinfo`, including solvability, private-key detection, range status, canonicalization, and multipath expansion.
- Implement address derivation for supported descriptors, including ranged, multipath, `combo()`, `addr()`, and standard `raw()` descriptors.

Bug Fixes:
- Prevent private keys from being returned in canonical descriptor responses.
- Validate descriptor checksums instead of replacing or ignoring supplied checksums.
- Correctly distinguish invalid descriptors from invalid derivation ranges using Bitcoin Core-compatible RPC errors.
- Reject descriptor keys from incompatible networks and prevent unsafe oversized derivation requests.

Enhancements:
- Require derivation ranges for ranged descriptors and reject ranges for fixed descriptors.
- Return canonical descriptors with checksums and derive addresses according to Bitcoin Core response shapes.

Tests:
- Replace legacy descriptor behavior tests with coverage for canonicalization, solvability, private-key redaction, checksum validation, address derivation, ranges, multipath expansion, combo descriptors, and network validation.


metaphorics (commented):
Gate hold: descriptor analysis is not safe to merge.

Blocking trust-boundary findings:

1. `crates/wallet/src/descriptor.rs:150-153`: `addr()` and `raw()` strip supplied checksums without validating them. Invalid checksums are accepted. Validate any supplied checksum before parsing; `deriveaddresses` must require one.

2. `crates/wallet/src/descriptor.rs:192-195`: checksumless miniscript descriptors are accepted by `deriveaddresses`. Bitcoin Core requires exactly one valid descriptor checksum for every derive request. Put one required-checksum gate before both parsing branches.

3. `crates/rpc/src/handlers/wallet.rs:84-89`: a `[0, 4294967295]` range drives billions of derivations and retained addresses. Enforce Core's maximum derivation count (currently 1,000,000 inclusive entries) before work starts and return `Range is too large`.

4. `crates/wallet/src/descriptor.rs:192-197`: xpub/xprv network kind is never checked against the requested chain. A testnet key can be accepted on mainnet and rendered as a `bc1...` address. Validate every extended key, including multipath/private keys, before derivation.

5. `crates/wallet/src/descriptor.rs:153-158`: canonical `addr()`/`raw()` descriptors are returned without checksums. Core's canonical descriptor string includes `#checksum`; append it while keeping the separate checksum member.

The branch update started fresh CI, but these request-validation and resource-exhaustion defects hold the PR regardless. No additional changes were pushed.

rabbitson87:
Rebased on current `main`. Self-reviewed against Core's `CheckChecksum` and found two defects in my own work, fixed here.

## 1. A wrong checksum was accepted and silently replaced

The text was split on its last `#`, the supplied checksum thrown away, and a freshly computed one put in the response. So a caller who mistyped a descriptor was told it was fine and handed back **a different descriptor than the one they meant** — and catching exactly that typo is the checksum's entire purpose.

Worse, a test pinned the behaviour: `getdescriptorinfo_reports_the_checksum_of_what_it_was_given` asserted the checksum was "computed, not echoed", which is true and beside the point. It has been replaced by Core's rules.

`checked_checksum` now implements `CheckChecksum` from `script/descriptor.cpp` refusal for refusal, with Core's own messages:

| input | answer |
|---|---|
| `addr(…)#qqqqqqqq` | `Provided checksum '…' does not match computed checksum '…'` |
| `addr(…)#short` | `Expected 8 character checksum, not 5 characters` |
| `addr(…)#aaaaaaaa#bbbbbbbb` | `Multiple '#' symbols` |

All `-5`, as Core's are.

## 2. `deriveaddresses` did not require a checksum

Core passes `require_checksum = true` at that call site **and only there** (`rpc/output_script.cpp:315`), and the reason is what those addresses are for: someone sends money to them. A mistyped descriptor derives perfectly good addresses that nobody holds the keys for, and the checksum is the only thing between a typo and that.

`getdescriptorinfo` keeps it optional, which is also Core's choice. The two call sites now differ on purpose rather than by accident, and there is a test on each side saying so.

## The checksum was certifying itself

Every existing test compared this implementation against its own output — an 8-character result, bech32 charset, not equal to the input. None compared it to the specification. Added the BIP380 vector, `raw(deadbeef)#89f8spxm`, which Core answers identically.

The `deriveaddresses` fixtures append a computed checksum rather than hard-coding one per descriptor, since their subject is derivation. The checksum's own correctness is pinned by the vector, and its enforcement by the three refusals.

## One pre-existing failure, not from this branch

`crates/wallet/tests/psbt_roundtrip.rs` fails under `--no-default-features` — on `main` exactly as here, which I checked before assuming. It needs the kernel script backend, and that reaches the wallet crate only through workspace feature unification, so CI's `cargo test --workspace` passes it while a per-crate run gets `portable script backend cannot verify this non-taproot spend`. Worth its own issue: a test whose outcome depends on another crate's defaults leaking in is green for the wrong reason.

## Verification

`cargo fmt --all --check` clean. `cargo clippy -p bitcoin-rs-rpc -p bitcoin-rs-wallet --lib --tests` clean. `rpc` and `wallet` suites green apart from the pre-existing failure above.

rabbitson87:
All five findings repaired. Apologies for answering the first two before reading your review — I had checked for reviews with a broken shell loop, took the 404 as "none yet", and self-audited instead. Findings 1 and 2 I reached independently in that pass; 3, 4 and 5 are yours and are fixed now.

## 3. The range is bounded before any work starts

`ParseDescriptorRange` refuses `high >= low + 1000000` and any `high` with its top bit set. Both checks are in the parser, so `[0, 4294967295]` is refused before a single derivation rather than after four billion of them. The test asserts the refusals are **fast** as well as correct — a bound checked inside the loop would also return an error, eventually, having already done the damage.

One thing worth recording: I first expected `[0, 2147483647]` to answer "End of range is too high". It does not, in Core or here — `2^31 - 1` is the largest legal index, so it clears that check and trips the size limit instead. That ordering is why Core has both messages, and the test now pins it.

## 4. Extended keys are checked against the network they will be rendered for

You are right that this is unreachable in Core, and the reason is worth stating because it explains why the check has to exist here at all: `DecodeExtPubKey` resolves the version bytes against the running chain's Base58 prefix, so a `tpub` on mainnet fails to decode and the descriptor never parses. rust-bitcoin decodes both prefixes into one type and keeps the network as a field — so the descriptor parses fine and the mismatch survives to derivation, where `address(network)` re-encodes with *this* network's prefix and returns a `bc1...` address for a testnet key. Well-formed, and nobody holds the private key.

Checked at `NetworkKind` granularity: BIP32 has two prefix sets, one for mainnet and one shared by every test network, so a signet key and a testnet key are genuinely indistinguishable and refusing between them would refuse something valid. Runs over every key including multipath, before any derivation.

## 5. The canonical descriptor carries its own checksum

Confirmed at `DescriptorImpl::ToString`, which ends in `return AddChecksum(ret)` — so this applies to every descriptor, not only `addr()`/`raw()`, and to each multipath expansion. Returning the bare form meant a caller who copied it out got something `deriveaddresses` then refuses for having no checksum; the test round-trips exactly that.

The checksum on the canonical form is of the **canonical** string while the `checksum` field stays the **input's**, which is not a duplication: they differ whenever the parse rewrites anything, an `xprv` replaced by its `xpub` most obviously.

## 1 and 2, for completeness

Fixed in the previous commit. `checked_checksum` implements `CheckChecksum` refusal for refusal — multiple `#`, wrong length, mismatch — and `deriveaddresses` requires one while `getdescriptorinfo` does not, matching Core's two call sites. Added the BIP380 vector `raw(deadbeef)#89f8spxm`, since every existing checksum test compared the implementation against its own output rather than the specification.

## Verification

`cargo fmt --all --check` clean. `cargo clippy -p bitcoin-rs-rpc -p bitcoin-rs-wallet --lib --tests` clean. `rpc` and `wallet` suites green apart from `psbt_roundtrip`, which fails identically on `main` — it needs the kernel script backend and reaches it only through workspace feature unification.

rabbitson87:
## Against the new repository boundary: this stays, whole

The boundary being drawn is "up to the wallet-facing interface; wallet functionality moves to its own repository". These two methods are on the node side of that line, and the sharpest evidence is Core's own placement:

| method | Core puts it in |
|---|---|
| `getdescriptorinfo` | `src/rpc/output_script.cpp` |
| `deriveaddresses` | `src/rpc/output_script.cpp` |

Neither needs a wallet, a key, or any spendable state — they parse a descriptor and derive scripts. Core's kernel library makes the same cut harder still: **the vendored tree this repository pins has no `src/wallet/` directory at all**, and these two are still in it.

So nothing here is scope to drop. The five findings this PR closes — checksum validation, the required checksum on `deriveaddresses`, the derivation bound, the network check, the canonical checksum — are all about node-side RPC behaviour.

## What the boundary does touch here, and why not in this PR

The implementation lives in `crates/wallet/src/descriptor.rs`, so the node's RPC surface currently reaches into the wallet crate for it. Under the new boundary that is the wrong crate — not because the functionality is wallet functionality, but because the node should not depend on a crate that is leaving.

Moving it is a follow-up, not a shrink of this PR:

- it is a module move plus a re-export, touching no behaviour;
- doing it here would rewrite a branch that has already been reviewed in detail and repaired against that review, and would mix a structural move into a correctness fix;
- the destination depends on where the boundary work lands the descriptor code, which is not decided yet.

Happy to do it as a follow-up the moment the target is settled, or to fold it in here if you would rather it move now.

gosunuts:
Current-main review: the feature need is still real, but this branch is outdated structurally. It was written around the in-tree wallet/miniscript ownership that has since been removed. Current `main` keeps descriptor RPCs node-side and still has placeholder behavior (`issolvable`/`hasprivatekeys` fixed false; `deriveaddresses` limited to `addr()`).

So I would not merge this PR as-is. The right follow-up is to reimplement the useful descriptor parsing/derivation behavior directly in the current walletless RPC/util boundary, without reintroducing the removed wallet crate dependency. This PR can then close as superseded by that rewrite.

metaphorics (commented):
Gate-and-merge hold: Gate 2, head `0dca5c2`.
QA is clean: full-diff review of the descriptor trust boundary found correct enforcement on all paths (range bounds pre-work, checksum Required/Optional split matching Core, -5/-8 error mapping, network check before derivation, no private-key leakage). One concern.

metaphorics:
@greptileai

metaphorics:
@sourcery-ai review

metaphorics:

metaphorics:

metaphorics:

metaphorics:

metaphorics:
@cursor Resolve, validate, and fix the pr comments and feedbacks + fix merge conflicts in the pr

Issue #123, opened 2026-08-25
Support mempool explorer through the native Esplora API


gosunuts:
## Goal

Support upstream [`mempool/mempool`](https://github.com/mempool/mempool) as the recommended full explorer frontend for bitcoin-rs.

mempool should run as a separate application/container and connect directly to bitcoin-rs through its standard APIs:

```text
browser
  -> mempool frontend/backend
  -> Bitcoin Core RPC + Esplora
  -> bitcoin-rs
```

Do not embed mempool, depend on its source tree, or require a separate electrs service.

## Why this fits bitcoin-rs

mempool already supports an Esplora backend for address and transaction lookup while using Bitcoin Core RPC for node data.

Once #124 provides a native Esplora-compatible API, bitcoin-rs can provide the same data directly from its own `TxLookup`, `ScriptIndex`, block store, and mempool without running another blockchain indexer.

This keeps explorer UI and explorer-specific state outside the node while giving users a full-featured frontend through a standard interface.

## Work

- Complete the Esplora surface required by mempool under #124
- Verify the Core RPC methods used by mempool against bitcoin-rs
- Run upstream mempool with its Esplora backend pointed directly at bitcoin-rs
- Make block, transaction, address/history, mempool, fee, and broadcast paths work without electrs
- Add a reproducible Docker or local integration example

## Acceptance criteria

- [ ] Upstream mempool runs against bitcoin-rs without source patches or a fork
- [ ] mempool uses bitcoin-rs Core RPC and native Esplora API directly
- [ ] Address and transaction history work through bitcoin-rs `ScriptIndex`
- [ ] Transaction lookup works through bitcoin-rs `TxLookup`
- [ ] Live mempool and broadcast paths use the normal bitcoin-rs mempool/P2P path
- [ ] No Electrum/electrs service or additional blockchain index database is required
- [ ] mempool remains a separate deployable application/container
- [ ] A reproducible Docker or local integration example is provided

Related: #78, #101, #114, #124

gosunuts:
Closing this in favor of #125, which now replaces the Electrum runtime with a node-owned `ScriptIndex` and adds an Esplora-compatible HTTP surface suitable for `mempool/mempool` integration. The current direction is to use `bitcoin-rs` as the synced/indexed backend, with explorer readiness after IBD and index catch-up rather than supporting partial explorer semantics during sync.

PR #125, opened 2026-08-25
Replace Electrum with ScriptIndex and Esplora

### LINKED ISSUE #124: Replace Electrum API with generic script indexing and Esplora
## Goal

Replace the Electrum-specific API with generic node-owned indexes and expose application-facing blockchain data through an Esplora-compatible HTTP API.

Keep indexing independent from protocol:

- `TxLookup` / `--txindex`: arbitrary historical tx lookup and Bitcoin Core compatibility
- `ScriptIndex`: wallet/explorer lookup by script/address, including current UTXOs and historical funding/spending activity

`ScriptIndex` exists to serve wallet and explorer queries such as address/script UTXO lookup and transaction history without scanning the chain.

## Why Esplora instead of Electrum

Esplora is a better general application API for bitcoin-rs:

- stateless HTTP instead of a stateful TCP/session protocol
- directly usable by browsers, wallets, explorers, CLI tools, and ordinary HTTP clients
- resource-oriented tx/block/address/mempool endpoints
- no wallet-specific subscription state in the node
- the same indexes serve wallet and explorer use cases

Electrum-specific subscriptions and server/session methods do not add node data capabilities once `TxLookup` and `ScriptIndex` exist.

## Work

- Generalize the current Electrum scripthash/history index into a protocol-independent `ScriptIndex`
- Preserve Core `--txindex` semantics independently
- Add Esplora-compatible tx, block, address/scripthash, mempool, fee, and broadcast endpoints
- Remove the Electrum TCP/JSON-RPC server and subscription/session code
- Reuse existing block, mempool, tx lookup, and compact funding/spending index data; do not add an Esplora-specific database
- Keep the Esplora HTTP surface independent from index configuration: endpoints backed by unavailable optional indexes should report that capability as unavailable rather than requiring a separate `--esplora` mode

## Capability model

Esplora is an API contract, not an indexing mode.

- block/tip endpoints -> chain state / block store
- mempool endpoints -> mempool
- broadcast -> normal mempool admission + P2P
- arbitrary historical tx lookup -> `TxLookup`
- wallet/explorer address/scripthash UTXO and history queries -> `ScriptIndex`

## Acceptance criteria

- [ ] Esplora is backed directly by node state and generic indexes
- [ ] `ScriptIndex` supports wallet/explorer address and scripthash UTXO lookup without scanning the chain
- [ ] `ScriptIndex` also supports historical address/scripthash transaction queries
- [ ] Esplora does not require a separate enable flag; endpoint availability follows node capabilities
- [ ] Core `getindexinfo` advertises txindex only when `--txindex` is explicitly enabled
- [ ] Electrum protocol code is removed without losing its useful indexing capabilities

Related: #117, #123

gosunuts:
## Summary

Closes #124.

- Replaces the Electrum server/API with Esplora routes on the existing node HTTP listener.
- Keeps the electrs-compatible physical ScriptIndex rows (`Funding`, `Spending`, `TxConfirmed`) while removing the old Electrum runtime and configuration.
- Reduces configuration to `--scriptindex[=true|false]`; there is no history mode or separate Esplora bind flag.
- Leaves chain, block, transaction, mempool, and fee Esplora endpoints available without ScriptIndex; script/address endpoints return service unavailable until ScriptIndex is enabled and caught up.
- Removes remaining dead ScriptIndex response state (`unspent` embedded in history snapshots and redundant index-info API), obsolete index reset API/tests, and retired Electrum benchmark artifacts/references.
- Marks the old plan/deviation ledger as historical; preserves `electrs` naming only where it describes the retained on-disk row layout.

## Validation

- `cargo test -p bitcoin-rs-rpc --lib` (234 passed)
- `cargo test -p bitcoin-rs-index` (50 passed)
- `cargo check -p bitcoin-rs-rpc`
- `cargo check -p bitcoin-rs-node --no-default-features --features fjall` is blocked on Windows by the pre-existing non-Windows `signal_hook::iterator` import. The default node test path is also blocked here because `libbitcoinkernel-sys` requires CMake on PATH.


gosunuts (commented):
The overall direction is good: removing the Electrum session/protocol layer, keeping the node-owned indexing work, and exposing a generic `ScriptIndex` is the right shape for #124. I would not merge this version yet, though, because the new HTTP surface is not actually Esplora-compatible and it drops behavior the old Electrum path already had.

1. **P1 — Match the actual Esplora response contract instead of forwarding Core RPC JSON.**

   Several routes currently dispatch Core RPC and return that result directly (`/tx/:txid`, `/block/:hash`, `/mempool`), while `/address/:address/txs` and `/scripthash/:hash/txs` return only `txid[]`. These shapes do not match Esplora. Esplora expects its own transaction/block/status/mempool schemas, paginated address history, etc. The current `/fees/recommended` route is also a mempool.space-style endpoint; standard Esplora exposes `/fee-estimates`.

   Reusing Core RPC internally is fine, but `esplora.rs` needs to be a representation/composition layer that converts node data into the Esplora contract. This is important for #123 as well; upstream `mempool/mempool` will not be able to treat the current surface as an Esplora backend.

2. **P1 — Restore mempool overlay for address/script queries.**

   The removed Electrum implementation combined confirmed history with the live mempool: it detected mempool funding outputs, mempool spends of confirmed outputs, and unconfirmed script activity. The new `ScriptIndexQuery` only returns confirmed chain-derived state, and the Esplora UTXO/history routes expose it directly.

   That means an address can report a confirmed UTXO as unspent after it has already been spent in the mempool, and newly received unconfirmed outputs/activity are absent. Esplora address/script views need current mempool state too.

   I would keep `ScriptIndex` purely confirmed/durable and compose it with the node `Mempool` at query time:

   `ScriptIndex (confirmed) + Mempool (volatile) -> Esplora address/script view`

   Do not persist mempool rows into `ScriptIndex` just for this.

3. **P2 — Be explicit about the v3 index-format break or rebuild automatically.**

   This changes the index format to v3 and `IndexWriter::open()` now rejects every older version. That also affects users who had only `--txindex` enabled, not just removed Electrum users. Since this is derived data, automatically resetting/rebuilding an old format may be simpler than hard-failing startup. If hard failure is intentional, the docs need to state that existing v2 `txindex` data must also be rebuilt.

There are also current CI blockers that need fixing before merge:

- workspace tests fail because `StorageError` was removed from the imports but is still used by the node test target in `txindex_worker.rs`;
- clippy/kernel-parity fail on six `doc_markdown` warnings for unquoted `ScriptIndex` references under `-D warnings`.

The target architecture I think this PR should land on is:

`TxLookup -> Core txindex`

`ScriptIndex -> confirmed script funding/spending data`

`Mempool -> unconfirmed state`

`Esplora -> projection over BlockTree + TxLookup + ScriptIndex + Mempool`

Once those boundaries are preserved and the wire format is actually Esplora-compatible, the Electrum removal looks like a net simplification.

gosunuts:
Addressed in a048ae2.

- Replaced Core-RPC JSON forwarding with Esplora transaction, block, mempool, fee-estimate, address, and scripthash projections, including chain pagination routes.
- Address/scripthash views now compose confirmed ScriptIndex data with live mempool funding/spending activity. Confirmed UTXOs spent in mempool are hidden; unconfirmed outputs and transactions are included without persisting mempool rows.
- Older/unversioned derived TxLookup/ScriptIndex formats now reset in bounded batches and rebuild automatically; UTXO and block-body data are untouched.
- Restored the test-only `StorageError` import and corrected `ScriptIndex` rustdoc links.

Validation: `cargo test -p bitcoin-rs-rpc --lib` (234), `cargo test -p bitcoin-rs-index` (51), `cargo test -p bitcoin-rs-mempool` (103), and `cargo clippy -p bitcoin-rs-rpc --lib --no-deps -- -D warnings` all pass. The Windows node check remains blocked by the pre-existing non-Windows `signal_hook::iterator` import.

gosunuts:
Addressed the documentation review and resolved the rebase conflicts. `index-read-path.md` now retains the ScriptIndex resolver benchmark and cost analysis, while clearly treating retained Electrum dispatch material as historical evidence. `optimization-candidates.md` is restored as the broader cross-subsystem candidate survey and its current wording now refers to ScriptIndex. Pushed in `2d679b3` and `0ba2dee`.

Issue #78, opened 2026-08-13
Define and differentially verify the Bitcoin Core-compatible RPC, REST, and ZMQ contract


gosunuts:
## Problem

The project exposes a useful Bitcoin Core-shaped interface:

- synchronous HTTP JSON-RPC with Core method names;
- a deliberately small REST gateway for remote validators;
- ZMQ hash/raw topics and pubsequence block connect/disconnect notifications;
- Core-style configuration and getzmqnotifications metadata.

The implementation has handler smoke tests, selected response-schema tests, REST edge-case tests, and ZMQ transport/wiring tests. Those are a solid foundation but do not define a complete supported API contract or establish drop-in compatibility with Bitcoin Core.

Current gaps include:

- no machine-readable inventory of supported, partial, disabled, and unsupported methods;
- selected JSON result-key tests rather than broad behavioral/error differential coverage;
- no versioned Core reference target for API compatibility;
- REST intentionally implements only chaininfo and headers endpoints;
- pubsequence omits mempool A/R events (#74);
- ZMQ is best-effort and has no durable recovery contract (#77);
- some wallet methods are intentionally disabled because this node has no private keys;
- documented gaps remain in reorg/mempool and index readiness semantics.

## Goal

Publish and enforce a narrow, explicit, versioned Bitcoin Core compatibility contract for RPC, REST, and ZMQ. Every claimed supported surface must be verified against a pinned Bitcoin Core reference through deterministic fixtures and differential tests. Unsupported and intentionally partial surfaces must fail or report capability status predictably.

This issue does not require implementing every Core API method.

### 1. Freeze the compatibility target and surface manifest

- Pin the Bitcoin Core major/minor release used for each compatibility claim.
- Add a machine-readable manifest for every exposed RPC, REST route, and ZMQ topic:
  - supported;
  - supported with documented deviations;
  - disabled by product policy;
  - not implemented.
- State parameter forms, prerequisites, required indexes, pruning behavior, auth policy, and readiness behavior.
- Generate or validate human API reference material from the manifest.

### 2. JSON-RPC differential harness

For each supported method, compare bitcoin-rs and the pinned Core reference using deterministic regtest and fixture chains.

Cover:

- success JSON shape, types, optional fields, ordering where wire-visible, and numeric semantics;
- positional and named parameters where applicable;
- invalid params, unknown objects, pruned/missing data, disabled index, and auth failures;
- RPC error code and stable error-category/message behavior;
[...]

PR #126, opened 2026-08-25
feat(rpc): pin the Core reference and enforce a compatibility manifest

rabbitson87:
Closes the first scope item of #78: a pinned Bitcoin Core version and a machine-readable inventory of every RPC method, REST route and ZMQ topic this node exposes.

## The pin is load-bearing, and says what it is

The node links libbitcoinkernel built from the tree `libbitcoinkernel-sys` vendors, so naming that crate's version names the Core revision. `bitcoinkernel 0.2.1` → `libbitcoinkernel-sys 0.3.0` → Bitcoin Core **31.99.0**.

The manifest says plainly that 31.99 is Core's master convention for "after 31.x, before 32.0" — a pre-release snapshot, not something an operator can install from a release tarball. A compatibility contract quietly pointing at a moving target is worse than none.

## Nothing claims `supported`

The vocabulary is #78's four states plus one. `supported` means implemented **and** differentially verified against the pinned reference, and nothing qualifies, because the harness in scope item 2 does not exist yet.

Folding that into "supported" is exactly the non-goal #78 names — *treating a Core method name as proof of behavioral compatibility* — so there is a state for "implemented, and nothing has checked it". 55 of the 60 live methods are in it. Five carry recorded deviations, found by reading the implementations rather than assumed:

| method | what differs |
|---|---|
| `getrpcinfo` | `active_commands` always empty, `logpath` always `""` |
| `getmemoryinfo` | `mode=stats` only; RSS, not Core's locked-pool accounting |
| `estimaterawfee` | one estimate replicated across all three horizons |
| `estimatesmartfee` | this node's estimator; no `errors` array |
| `getmempoolinfo` | `usage` estimates different structures; `maxmempool` bounds vsize, not `usage` |

The 12 private-key methods are `disabled`. Anything absent is `not_implemented` and answers `-32601` — asserted, not stated.

## The manifest is enforced, not described

This is the part I care about. A document nobody checks rots, and a rotted compatibility contract is worse than none: a client reads it, believes it, and finds out at runtime. So it is embedded with `include_str!` and cross-checked against what it describes.

- **The dispatcher's own source is embedded and its match arms parsed**, so the method sets must agree in both directions. A method added without an entry fails; so does an entry for a method that does not dispatch.
- Every entry called disabled must be **refused as disabled**, and every entry not called disabled must **not** be. The second is what stops a private-key method being quietly downgraded to "implemented" and advertising a signing capability this node does not have.
- A method the manifest omits must answer `-32601`, so its closing sentence is a checked claim.
- A `deviation` entry must **say what the deviation is**, machine-readably — a status alone tells a client that this node differs and not how.
- `supported` is refused while `differential_harness = false`, so the state cannot become a synonym for "implemented" one entry at a time.
- **The pinned kernel versions are checked against `Cargo.lock`.** A dependency bump that moves the Core source has to come back through this file, which is exactly when the claims need re-reading.
- ZMQ topics are checked from both sides: against a restated list in the RPC crate, and against `ZmqTopic` itself in the node crate, where the names that go on the wire live.

## Mutation audit

Eight mutations, all killed: a dispatched method losing its entry; the manifest promising a method that does not dispatch; an entry claiming `supported`; the pinned kernel drifting from `Cargo.lock`; a misnamed ZMQ topic; a disabled method claimed as implemented; a deviation that says only "differs"; and a method added to the dispatcher with no claim.

**Two of those survived the first run** — the disabled downgrade and the empty deviation text. The two checks that now catch them exist because of it.

The differential harness (#78 item 2), REST and ZMQ behavioral fixtures (items 3 and 4), and capability/readiness reporting (item 5). This is the inventory they get written against — item 2 in particular needs an agreed shape for running a pinned Core alongside the suite, which is a discussion rather than a commit.

## Verification

`cargo fmt --all --check` clean. `cargo clippy -p bitcoin-rs-rpc -p bitcoin-rs-node --lib --tests` clean under both `--no-default-features --features fjall` and default features. `rpc` and `node` suites green. One line added to `Cargo.lock` (`toml` as an rpc dev-dependency; already a workspace dependency).

## Summary by Sourcery

Pin the Bitcoin Core compatibility reference and enforce a machine-readable inventory of the node’s exposed APIs and known compatibility differences.

New Features:
- Add a machine-readable Bitcoin Core compatibility manifest covering exposed RPC methods, REST routes, and ZMQ topics.
- Pin compatibility claims to the Bitcoin Core revision supplied by the locked kernel dependencies.

Bug Fixes:
- Prevent ranged descriptor requests from being incorrectly accepted by `deriveaddresses` without wallet support.

Enhancements:
- Enforce consistency between the compatibility manifest and the RPC dispatcher, REST router, ZMQ publisher, and dependency lockfile.
- Record explicit behavioral deviations and distinguish unverified implementations from differentially supported APIs.

Documentation:
- Document the pinned Core reference, API surface, compatibility statuses, and known deviations in a TOML manifest.

Tests:
- Add compatibility checks ensuring declared and dispatched RPC methods, REST routes, and ZMQ topics remain aligned.
- Verify unsupported methods return `-32601`, disabled capabilities remain refused, deviations are described, and unsupported `supported` claims are rejected.


rabbitson87:
## Against the new repository boundary: no change yet, and the manifest is where it lands later

The boundary being drawn — up to the wallet-facing interface, wallet functionality to its own repository — cuts through this manifest's RPC list. Core's own placement is the sharp criterion, and the pinned kernel tree makes it sharper still: **it ships with no `src/wallet/` directory at all.**

| in the manifest | Core's placement | under the boundary |
|---|---|---|
| `getdescriptorinfo`, `deriveaddresses` | `src/rpc/output_script.cpp` | stay |
| `scantxoutset` | `src/rpc/blockchain.cpp` | stays |
| `combinepsbt`, `finalizepsbt` | `src/rpc/rawtransaction.cpp` | stay |
| `walletcreatefundedpsbt`, `walletprocesspsbt`, `bumpfee` | not in the kernel tree | leave |
| the 12 disabled private-key methods | not in the kernel tree | leave |

Only five of the twenty wallet-group entries are wallet functionality by that line. The rest are node RPCs that happen to be grouped under `wallet` here — worth regrouping when the move happens.

**Nothing changes in this PR.** The manifest is enforced against the dispatcher: `the_manifest_and_the_dispatcher_name_the_same_methods` fails in both directions, so an entry cannot be removed before the method is. Editing it ahead of the code would break the build to describe a state that does not exist yet.

That constraint is the useful part. When the boundary work removes those methods, the manifest has to be updated in the same commit or the suite goes red — so the compatibility record cannot silently keep advertising a surface that left.

gosunuts:
Current-main review: this needs an architectural decision before merge. The compatibility-manifest idea is useful, but #176 introduced a different ownership model where `corepc-types::v31` is the typed RPC response contract. Keeping both as independent sources of response-shape truth risks recreating the exact drift this PR was intended to prevent.

I would not merge #126 unchanged while #176 remains the intended direction. Either narrow this PR to capabilities that corepc cannot provide — pinned-reference provenance, method inventory, value-level/differential status — or retire it in favor of #176 plus a separate differential-compatibility layer. One schema owner at the RPC boundary is preferable.

metaphorics (commented):
Gate-and-merge hold: blocking Gate 4 finding, head `6ea37d4`.

Blocking, Gate 4 class 1 (reachable wrong result in the default build): `every_listed_method_dispatches` in `crates/rpc/src/compat_manifest.rs:151-164` dispatches every documented non-`not_implemented` row unconditionally, including `getzmqnotifications` — but `handlers.rs:191-195` registers that method only under `feature = "zmq"`, and `crates/rpc/Cargo.toml` defaults to no features. The default-feature test build therefore reaches `MethodNotFound` and fails; the set-equality test (`live_registry` is feature-pruned vs the unconditional manifest) trips on the same row. Reproducible by `cargo test -p bitcoin-rs-rpc` with default features. Needs an author decision (filter feature-gated rows vs `cfg`-gating the test), so holding rather than redesigning here. No merge on this oid until then.

metaphorics (commented):
## Fix: filter feature-gated rows in compat manifest tests

### Problem

`every_listed_method_dispatches` and `the_manifest_and_the_dispatcher_name_the_same_methods` iterate all `[[rpc]]` rows from `docs/api/core-compat.toml`, including `getzmqnotifications`. That method is registered in `DISPATCH_TABLE` only under `#[cfg(feature = "zmq")]` (`handlers.rs:191`), while the default test build has no features active. The dispatch test hits `MethodNotFound` for that row, and the set-equality test trips because the TOML lists a method the dispatcher doesn't serve.

### Convention

The repo already handles this in two places:
- `handlers.rs` gates the `DISPATCH_TABLE` entry with `#[cfg(feature = "zmq")]` and has paired `#[cfg(feature = "zmq")]` / `#[cfg(not(feature = "zmq"))]` tests in `registry_tests` (`handlers.rs:435-461`).
- `manifest.rs` carries a `feature` field on every `Entry` and exposes `Entry::shipped()`, which returns `false` when the feature is inactive (`manifest.rs:118-128`). `is_registered` already consults `shipped()` (`manifest.rs:1734-1738`).

The fix follows the `manifest::Entry::shipped` convention rather than raw `cfg` gating, because the TOML manifest doesn't carry feature metadata — the Rust `MANIFEST` table does. A `shipped_rpc_methods()` helper filters to rows whose feature is active in the current build, and both tests consult it.

### Patch

```diff
diff --git i/crates/rpc/src/compat_manifest.rs w/crates/rpc/src/compat_manifest.rs
index 8387b0a..69e86d5 100644
--- i/crates/rpc/src/compat_manifest.rs
+++ w/crates/rpc/src/compat_manifest.rs
@@ -75,6 +75,7 @@ mod tests {
     use crate::context::Context;
     use crate::error::RpcError;
     use crate::handlers;
+    use crate::manifest::{self, SurfaceKind};
 
     /// `Cargo.lock`, so the pinned reference cannot quietly stop being pinned.
     const CARGO_LOCK: &str = include_str!("../../../Cargo.lock");
@@ -84,6 +85,17 @@ mod tests {
             .unwrap_or_else(|err| panic!("the compatibility manifest must parse: {err}"))
     }
 
+    /// RPC method names that ship in this build, according to the declared
+    /// manifest. Feature-gated rows (e.g. `getzmqnotifications` under
+    /// `feature = "zmq"`) appear only when their feature is active, matching
+    /// `handlers::is_registered_method` and `manifest::Entry::shipped`.
+    fn shipped_rpc_methods() -> BTreeSet<&'static str> {
+        manifest::entries_of_kind(SurfaceKind::Rpc)
+            .filter(|entry| entry.shipped())
+            .map(|entry| entry.name)
+            .collect()
+    }
+
     fn entries(table: &toml::Table, key: &str) -> Vec<toml::Table> {
         let Some(array) = table.get(key).and_then(toml::Value::as_array) else {
             panic!("the manifest must carry a `{key}` array");
@@ -121,6 +133,12 @@ mod tests {
     /// promise the node does not keep, which is worse: a client reads it,
     /// depends on it, and gets `-32601`.
     ///
+    /// Feature-gated rows (e.g. `getzmqnotifications` under `feature = "zmq"`)
+    /// are filtered to the methods that ship in this build, so the comparison
+    /// is against the dispatcher as compiled, not against a hypothetical
+    /// all-features build. This mirrors `manifest::Entry::shipped` and the
+    /// `#[cfg(feature)]` gating used in `handlers::registry_tests`.
+    ///
     /// Uses `handlers::live_registry()` (the dispatch table) rather than
     /// source-scanning, because main's dispatcher routes through a static
     /// `DISPATCH_TABLE` that is the single source of truth for what dispatches.
@@ -129,10 +147,12 @@ mod tests {
         let live: BTreeSet<String> = handlers::live_registry().map(String::from).collect();
         assert!(!live.is_empty(), "the live registry must not be empty");
 
+        let shipped = shipped_rpc_methods();
         let table = manifest();
         let listed: BTreeSet<String> = entries(&table, "rpc")
             .iter()
             .map(|entry| field(entry, "method"))
+            .filter(|method| shipped.contains(method.as_str()))
             .collect();
 
         let missing: Vec<&String> = live.difference(&listed).collect();
@@ -158,10 +178,15 @@ mod tests {
     /// private-key method from `disabled` to `implemented_unverified` would
     /// otherwise pass every other check here, and would advertise a signing
     /// capability this node does not have.
+    ///
+    /// Feature-gated rows behind a feature that is off in this build are
+    /// skipped, matching `manifest::Entry::shipped` and the `#[cfg(feature)]`
+    /// gating used in `handlers::registry_tests`.
     #[test]
     fn every_listed_method_dispatches() {
         let ctx = Arc::new(Context::new());
         let handler = crate::Handler::new(ctx);
+        let shipped = shipped_rpc_methods();
         let table = manifest();
 
         for entry in entries(&table, "rpc") {
@@ -170,6 +195,9 @@ mod tests {
                 continue;
             }
             let method = field(&entry, "method");
+            if !shipped.contains(method.as_str()) {
+                continue;
+            }
             let outcome = handler.dispatch(&method, &json!([]));
             assert!(
                 !matches!(outcome, Err(RpcError::MethodNotFound(_))),
```

### Verification

- `git apply --check` against `6ea37d4`: **passes**.
- `cargo check -p bitcoin-rs-rpc --lib`: **passes** (non-test build clean).
- Test compilation has pre-existing errors in `handlers/mempool.rs` (`compute_txid` not found, tuple arity mismatch) present on the PR head both with and without this patch — unrelated to this change.

metaphorics:
@greptileai

metaphorics:
@sourcery-ai review

metaphorics:

metaphorics:

metaphorics:

PR #146, opened 2026-08-25
Remove blockfilterindex and BIP157/158 compact-filter indexing

### LINKED ISSUE #143: Remove blockfilterindex and compact-filter indexing
## Goal

Remove `blockfilterindex` and the associated BIP157/158 compact-filter indexing path from bitcoin-rs unless a concrete in-tree consumer requires it.

The current product surface is already covered by authoritative chainstate, `TxLookup`, `ScriptIndex`, Core-compatible RPC, and Esplora. Compact filters are optional derived state and do not contribute to consensus validation or the current wallet/explorer path.

## Scope

- Remove the `--blockfilterindex` configuration surface.
- Remove compact-filter database opening/state and filter-index lifecycle code.
- Remove filter generation from block apply and any associated readiness/reorg bookkeeping.
- Remove or update RPC/config compatibility code that only exists for block-filter indexing.
- Remove obsolete tests, benchmarks, docs, and PLAN references tied only to this feature.
- Keep consensus, block storage, pruning, `TxLookup`, `ScriptIndex`, and Esplora behavior unchanged.

## Rationale

Compatibility should be implemented where it has a real consumer or interoperability value. Carrying an unused optional index adds storage, configuration, recovery, testing, and maintenance cost without serving the current product surface.

If a future concrete BIP157/158 light-client consumer appears, compact-filter support can be reintroduced around that requirement rather than retained speculatively.

gosunuts:
Closes #143

Rebased onto `main` after the Esplora/ScriptIndex rework; the removal was re-applied on top of it.

## Why

BIP157/158 compact filters are derived optional state whose only purpose is serving light clients. This node never served them:

- `crates/p2p/src/handshake.rs` advertises only `NETWORK | WITNESS` — `NODE_COMPACT_FILTERS` (bit 6) is never set, so Neutrino-style clients never select us as a filter peer.
- `wire.rs` decodes `getcfilters`/`getcfheaders`/`getcfcheckpt` into the generic message enum, but no handler anywhere answers them.
- The only reader of the index was the `getblockfilter` RPC. `getindexinfo` even reported `basicblockfilterindex` unconditionally — including when the index was disabled, which over-reported versus Core.

So the feature cost two storage column families across four backends, a config/CLI/env/`bitcoin.conf` surface, per-block GCS construction + row writes + filter-header reorg bookkeeping in the apply path, and a dedicated crate with tests and benches — while being unreachable by its intended consumers. Chainstate, pruning, `TxLookup`, `ScriptIndex`, and the Esplora surface are untouched.

## What changed

- **Deleted `crates/filters`** (GCS codec, cfheaders chaining, `FilterIndex`/`FilterIndexLike`) and its workspace/`Cargo.toml`/feature wiring in `node` and `rpc`.
- **Config surface**: removed `blockfilterindex` from `Config`, CLI (`--blockfilterindex`), env (`BITCOIN_RS_BLOCKFILTERINDEX`), TOML layering, and the `bitcoin.conf` compat parser/merger.
- **Apply path**: removed `filter_index` from `ApplyHandles`, BIP158 construction and row persistence, the `filter_header_cache` disconnect bookkeeping, and `wants_filters` gating. `ApplyScratch` loses the same-block spent output-script map that existed solely to feed filter construction, and its constructors become infallible.
- **State**: removed `open_filter_index`, `DisabledFilterIndex`, the `filter_index()` accessor, and the two filter-index open tests.
- **RPC**: removed `getblockfilter` (now method-not-found) and the `basicblockfilterindex` entries from `getindexinfo`; `Context` drops its `filter_index` handle.
- **Storage**: removed the `Filters`/`FilterHeaders` column families and renumbered the surviving discriminants. Per `docs/policies/db-migration.md` §3.1 this is a breaking change; fjall/MDBX/redb open tables by name and ignore retired ones, while a RocksDB datadir written by an older binary must be wiped and resynced per §6.2. No in-place migration code was added, per §6.2.
- **Benches/example/tests**: dropped the filter workloads and flag from `sync_apply_metrics`, `sync_pipeline`, `mainnet_prefix_replay`, and `verify_replay_durability`; converted tests that used the filter fixture to the plain no-index fixture; deleted filter-only tests (`RecordingFilterIndex` and friends).
- **Docs**: README, `CONCEPTS.md`, `docs/getting-started.md`, `docs/policies/source-compatibility.md`, `docs/policies/db-migration.md`, `docs/benchmarks/end-to-end-sync.md` updated; `DEVIATIONS.md` §3 rewritten to record the removal and its compatibility deltas; `PLAN.md` goal/tree/Task 9/RPC list updated (historical benchmark-evidence lines left intact). Historical run artifacts under `docs/benchmarks/data/` are untouched.
- **Windows build fix (drive-by, required for verification)**: main removed the `cfg(not(windows))` gates around `install_shutdown_handler`, but signal-hook has no `iterator` module on Windows, so the node crate did not compile on Windows at all. The unix handler is gated again and Windows gets a documented no-op stub with the same signature.

## Verification

- `cargo test -p bitcoin-rs-rpc`: 281 tests pass.
- `cargo test -p bitcoin-rs-node --no-default-features --features fjall`: 450 passed, 33 failed — every failure is the pre-existing Windows-environment `Os error 5 PermissionDenied` family in checkpoint/corpus I/O tests (same signature as on the previous main base; unrelated to this change). Net −9 tests versus main's filter suite = the deleted filter tests.
- `cargo test -p bitcoin-rs-storage`: failure set byte-identical to `origin/main` (3 pre-existing `block_file::tests::disk_usage_*` Windows failures).
- Node integration binaries (`rpc_wiring`, `sync_smoke`, `bitcoin_conf_compat`, `config_layered`): 33/33 pass.
- `cargo clippy -p bitcoin-rs-node -p bitcoin-rs-rpc -p bitcoin-rs-storage --all-targets`: same error/warning set as `origin/main` (main's new rpc benches carry pre-existing deny-level `as`-conversion lint debt that CI never compiles directly); zero new findings.
- `cargo fmt --all` applied.


metaphorics:
@codex review

rabbitson87 (commented):
Reviewed against #143's stated scope, on `pr146` merged onto `main` at 00938d8.

**The scope call is right and the storage note is the best part of this PR.** The `DEVIATIONS.md` §3 entry does the thing removals usually skip: it works out what an existing datadir does. Naming that the surviving `ColumnFamily` discriminants were renumbered, that this is a §3.1 breaking change, that fjall/MDBX/redb open tables by name and therefore reopen unchanged, that a RocksDB datadir from an older binary must be wiped per §6.2, and that a leftover `<datadir>/filters` directory is simply unused — that is four different answers to "will my node still start", and all four are needed. Please keep that standard.

Keeping the `NODE_COMPACT_FILTERS` service-bit name in `crates/p2p/src/peer_info.rs` and `crates/rpc/src/handlers/network.rs` is also correct, and I want to say so explicitly since it looks like residue on a `grep`. Decoding a bit a *peer* advertises is unrelated to whether *we* maintain an index; removing it would make `getpeerinfo` under-report a peer's real capabilities. Right call.

Two things to fix.

---

### 1. `crates/node/tests/config_layered.rs:42` is now dead input

```rust
let env: [EnvPair; 4] = [
    ("BITCOIN_RS_STORAGE_BACKEND", "redb"),
    ("BITCOIN_RS_DBCACHE_MB", "1024"),
    ("BITCOIN_RS_BLOCKFILTERINDEX", "true"),   // <-- nothing asserts on this any more
    ("BITCOIN_RS_LOG_LEVEL", "warn"),
];
```

`blockfilterindex` is the only name in that file that appears exactly once — the assertion that used to consume it went with the field, and the input stayed. The env `match` in `config.rs` ends in `_ => {}`, so this entry is now silently discarded.

That is not merely untidy: the array is the test's whole statement about layered precedence, and one of its four entries no longer participates. A reader counting inputs against assertions will find three of four. Either drop the line, or — better, if it is deliberate — keep it and assert what it now means, that a retired env key is ignored rather than fatal. The second version is a real test; the current state is neither.

### 2. `DEVIATIONS.md` says "ignored" for a surface that errors

> `-blockfilterindex` / `bitcoin.conf` `blockfilterindex` / `BITCOIN_RS_BLOCKFILTERINDEX`: no longer recognized; the option is ignored exactly like any other unsupported `bitcoin.conf` key.

Two of those three are ignored. The CLI flag is not. `Config::load_from_args` calls `ConfigLayer::try_parse_from(args)` with no `ignore_errors` and then `err.exit()`, so `bitcoin-rs --blockfilterindex=1` prints a clap error and exits nonzero — it does not start and ignore the flag.

That is the difference between "your old command line still works, the index just is not there" and "your old command line no longer starts the node", and an operator reading this entry to plan an upgrade would get it backwards. The behaviour itself is defensible; the sentence describing it is not.

Worth splitting the bullet into the two surfaces that drop the value and the one that refuses it.

---

### The same question appears in #148

#148 removes `--utreexo-mode` and `BITCOIN_RS_UTREEXO_MODE` and lands in the identical split: hard clap error on the flag, silent drop on the env var. Neither PR documents the asymmetry, and this one documents it inaccurately.

This is worth one decision rather than two: when a config option is retired, does the node refuse to start, or start and warn? Right now the answer depends on which surface the operator happened to use, which is the one answer nobody chose. A `WARN` on a recognized-but-retired key would serve both PRs and cost little.

### Conflict

`CONFLICTING` against current `main`, 16 commits behind, with six conflicts — `DEVIATIONS.md`, `PLAN.md`, `crates/node/benches/sync_pipeline.rs`, `crates/node/src/apply.rs`, `crates/node/src/state.rs`, `crates/rpc/tests/handler_smoke.rs`. Four of those are real code, so this is a genuine rebase rather than a text merge.

Suggest landing **#148 first** — its only conflict is `PLAN.md`, and its 1094 deleted lines shrink what this one has to rebase across.

I did not push a rebase, since this is your branch and I did not want to land on top of local work.

gosunuts (commented):
LGTM on the latest head `5352e6a`.

The compact-filter removal is internally consistent across the node/apply path, RPC/config surface, storage schema, replay tooling, and service advertisement. The earlier review gaps are fixed: the retired config input is gone, CLI-vs-env/bitcoin.conf behavior is documented correctly, and `getnetworkinfo` no longer advertises `NODE_COMPACT_FILTERS` locally.

CI's only red lane is the pre-existing `bitcoin-rs-consensus` clippy pair; the kernel parity tests themselves pass, and this PR does not touch consensus.

No code-review blocker from me. The branch is currently one commit behind `main` (`a241192`), which also touches `apply.rs` / `state.rs`, so the remaining work is integration with current main rather than a #146 correctness issue.

PR #148, opened 2026-08-25
Remove the experimental Utreexo node mode

### LINKED ISSUE #144: Remove Utreexo experimental node support
## Goal

Remove the in-tree Utreexo node mode, accumulator integration scaffolding, and related pruning/runtime surface from bitcoin-rs.

The current product direction already keeps authoritative UTXO state compact and separates wallet/explorer queries into `TxLookup` and `ScriptIndex`. Utreexo would only remove the local UTXO set by replacing it with proof transport, proof availability, accumulator persistence, reorg recovery, and bridge infrastructure. For the current full-node/server target, that complexity does not justify the remaining state reduction.

## Scope

- Remove the `bitcoin-rs-utreexo` crate from the workspace.
- Remove `--utreexo-mode` and related feature/configuration plumbing.
- Remove Utreexo accumulator/proof/bridge scaffolding and placeholder node integration.
- Remove `UtreexoOnlyCoordinator` and Utreexo-specific pruning paths/tests.
- Remove Utreexo parity gates, docs, PLAN/CONCEPTS/DEVIATIONS references, and stale compatibility notes.
- Remove Utreexo-specific Cargo features/dependencies from node, binary, pruning, and workspace manifests.
- Keep normal UTXO validation, pruning, block storage, `TxLookup`, `ScriptIndex`, Esplora, and BIP300/301 behavior unchanged.

## Rationale

Utreexo is valuable when minimizing validator state is itself a product requirement, such as mobile/embedded or proof-native/stateless deployments. That is not the current bitcoin-rs target.

The existing implementation is not an operational compact validator: it does not provide a production proof transport/provider path, accumulator lifecycle integrated with connect/disconnect, or complete compact-mode recovery semantics. Carrying the partial feature creates a second validation architecture and pruning model without a concrete consumer.

If a real compact-validator use case appears later, Utreexo can be reintroduced around that requirement instead of retaining speculative integration in the main node.

gosunuts:
Closes #144

## Why

The Utreexo integration was scaffolding, not an operational compact validator: it had no production proof transport/provider path, no accumulator lifecycle wired into connect/disconnect, and no complete compact-mode recovery semantics. It carried a second validation architecture and pruning model with no in-tree consumer, while the full-node/server target keeps authoritative UTXO state compact via the shard-table UTXO set, `TxLookup`, and `ScriptIndex` already.

## What changed

- **Deleted `crates/utreexo`** (`Accumulator`/`Bridge`/`Proof` wrappers over rustreexo) plus the workspace `rustreexo` and `bitcoin-consensus-encoding` pins that existed only for it.
- **Config surface**: removed `utreexo_mode` from `Config`, CLI (`--utreexo-mode`), env (`BITCOIN_RS_UTREEXO_MODE`), TOML layering, and the `bitcoin.conf` merge.
- **Cargo features**: removed the `utreexo` feature and `bitcoin-rs-utreexo` dep from `crates/node`, the feature-forwarding entries, and the bin's `utreexo` passthrough.
- **Pruning**: removed `UtreexoOnlyCoordinator`, `BlockProcessed`, `PrunePolicy::utreexo_only`/`is_utreexo_only`, and the `utreexo_no_blocks` test. `prune_then_reorg` keeps its immediate-deletion coverage with an explicit aggressive `PrunePolicy` literal (retention still floors at the 288-block reorg margin). Regular `-prune=` pruning is untouched.
- **Gates**: removed the G8 utreexo parity gate file and its `[[test]]` target; the G8 number stays reserved in `PLAN.md` so G9+ references remain stable.
- **Docs**: README, `CONCEPTS.md`, `docs/policies/source-compatibility.md`, node/pruning READMEs updated; `DEVIATIONS.md` records the removal where the old feature-passthrough note lived; `PLAN.md` goal/tree/Task 6/Task 11/G8 updated (the 2026-05-19 ultrareview log is left as a historical audit record).
- **Windows build fix (drive-by, required for verification)**: main removed the `cfg(not(windows))` gates around `install_shutdown_handler`, but signal-hook has no `iterator` module on Windows, so the node crate did not compile on Windows at all. The unix handler is gated again and Windows gets a documented no-op stub. (Same fix as #146.)

## Verification

- `cargo test -p bitcoin-rs-pruning`: 3/3 pass (including the converted `prune_then_reorg`).
- `cargo test -p bitcoin-rs-node --no-default-features --features fjall`: 460 passed, 32 failed — every failure is the pre-existing Windows `Os error 5 PermissionDenied` family in checkpoint/corpus I/O tests (same signature as on main; unrelated to this change).
- Node integration binaries (`rpc_wiring`, `sync_smoke`, `bitcoin_conf_compat`, `config_layered`): 33/33 pass.
- `cargo clippy -p bitcoin-rs-node --all-targets -- -D warnings`: clean except the pre-existing Windows-only `sync_blocks_dir` `unnecessary_wraps` artifact (`#[cfg(not(unix))]` no-op variant; ubuntu CI unaffected).
- Benches compile; `cargo fmt --all` applied; `Cargo.lock` regenerated without `bitcoin-rs-utreexo`/`rustreexo`.


metaphorics:
@codex review

rabbitson87 (commented):
Reviewed against #144's stated scope, on `pr148` merged onto `main` at 00938d8.

**The removal itself is clean.** After this change six `utreexo` mentions remain in the tree, and all six are deliberate historical notes (`DEVIATIONS.md`, `PLAN.md` ×3, a plan log). No dangling `[[test]]` entry, no orphaned feature, no dead `mod`. `PrunePolicy::utreexo_only()` and `UtreexoOnlyCoordinator` are gone together, so pruning is left with one policy model rather than two. That is the outcome #144 asked for.

Two things I'd want changed before this merges, and one question.

---

### 1. `crates/node/src/signal.rs` is not part of this change

This hunk has nothing to do with Utreexo. It rewrites `install_shutdown_handler` into a `#[cfg(not(windows))]` / `#[cfg(windows)]` pair where the Windows arm is:

```rust
Ok(thread::spawn(move || {
    let _guard = shutdown_tx;
    loop { thread::park(); }
}))
```

The technical claim in the doc comment is correct — `signal_hook::iterator` is `#[cfg(not(windows))]`, so the old unconditional `use` genuinely does not compile there. But the consequence is that **on Windows the node silently loses graceful shutdown**: SIGINT/SIGTERM never reach the channel, and the parked thread holds the `Sender` alive so nothing downstream observes a disconnect either. A build error is loud; a shutdown path that returns `Ok` and does nothing is not.

Three reasons this should be its own PR rather than a passenger here:

- **There is a gate for exactly this subject.** `bin/bitcoin-rs/tests/gates/g12_graceful_shutdown.rs` — "SIGTERM during IBD → all in-flight writes flush, RPC connections drain with 5 s deadline, snapshot written, exit code 0." It has no `cfg` in it and is untouched by this PR, so the gate keeps asserting a contract the Windows build no longer honours.
- **CI cannot see either state.** Every lane is `ubuntu-latest`, so neither the previous Windows build failure nor the new stub is covered. Whichever way this goes, it goes unverified.
- **It is undocumented, in a PR that documents its other behaviour change.** This PR correctly adds a `DEVIATIONS.md` note for the removed `utreexo` feature passthrough. A platform where graceful shutdown does not work is at least as reportable, and gets no entry.

`signal-hook` also stays an unconditional dependency with `features = ["iterator"]` after this change, so the dependency graph does not reflect the split either.

If the goal is a working Windows build, the honest shapes are a `ctrlc`-style handler that actually sends on the channel, or a compile-time refusal — not an `Ok(())` that means nothing.

### 2. The two ways to set this option now fail differently

`--utreexo-mode` is removed from `ConfigLayer`, and `load_from_args` uses `try_parse_from(...)` with `err.exit()` and no `ignore_errors`. So:

- `bitcoin-rs --utreexo-mode=true` → clap rejects the unknown flag, the process exits nonzero. **Loud.**
- `BITCOIN_RS_UTREEXO_MODE=true bitcoin-rs` → the env `match` in `config.rs` ends in `_ => {}`, so it is dropped. The node starts in ordinary full-UTXO mode. **Silent.**

Same setting, same removal, opposite outcomes. An operator who configured this through the environment gets no signal that the mode they asked for no longer exists — they just get a node quietly doing something else.

The `_ => {}` catch-all predates this PR, so this is not a defect it introduces; it is a defect it makes reachable. I'd rather see it named in `DEVIATIONS.md` than left for whoever hits it. Worth deciding once rather than per-option, because **#146 has the identical question** and answers it differently — see my note there.

### 3. Question: is `G8` reserved or retired?

`PLAN.md` now says "the gate number stays reserved so G9+ references remain stable". Reserving a number so later references do not shift is reasonable, but nothing enforces it — a future gate could take `g08_` and the PLAN text would be quietly wrong. If the intent is real, `g15_workspace_version_sync.rs` is the precedent for a cheap structural assertion. If the intent is just "do not renumber", saying that instead would read more plainly.

---

### Conflict

`CONFLICTING` against current `main`, but only in **`PLAN.md`** — the merge is otherwise clean and the 28-file code removal applies unchanged. It is 16 commits behind. Of the six conflicting PRs in the queue this is by far the cheapest to resolve, and it deletes 1094 lines, which shrinks the conflict surface for the others. I'd put it at the front of the conflict queue for that reason alone.

I did not push a rebase, since this is your branch and I did not want to land on top of local work.

PR #150, opened 2026-08-25
feat(rpc): check result shapes against Core's own declared schemas

rabbitson87:
Closes the second scope item of #78 as far as it can be closed without a running reference, and says exactly where that line is.

Stacked on **#126**, which introduces the manifest this reconciles against.

## Why not just run a `bitcoind`

For values, that is the only build, and #78 keeps it as a documented lane. It cannot be the *whole* answer here: the pinned reference is **31.99.0**, a master snapshot, and no release binary exists for it. A harness that cannot run until Core cuts a release verifies nothing in the meantime.

Core declares each method's result shape in its own source, beside the handler, as `RPCResult` literals — and they are not documentation that drifts. `RPCHelpMan::Check` asserts Core's own handlers against them in debug builds, so Core keeps them true.

- `tools/core-rpc-schema/extract.py` reads them into `docs/api/core-rpc-schema.json` — 108 methods, with provenance.
- `crates/rpc/src/compat/schema.rs` dispatches 20 methods and compares field names, JSON types, and which fields Core marks optional.

## The manifest records differences field by field, checked both ways

A field that starts differing fails. **A field the manifest records as differing and which no longer does also fails** — a stale deviation is a lie in the other direction, and quietly closing a gap without updating the document is the usual way a compatibility document rots.

Three real differences, none previously recorded:

| method | difference |
|---|---|
| `getblockchaininfo` | omits `bits`, `target` |
| `getmininginfo` | omits `bits`, `blockmintxfee`, `next`, `target` |
| `getmempoolinfo` | omits 5 policy fields; emits `mempool_sequence`, which Core carries on `getrawmempool` |

The `getmininginfo` row is exactly what #109 adds, arrived at independently — which is the first thing that gave me confidence the oracle is measuring something real.

## Three the first run got wrong, and they were mine

Worth naming, because all three would have been filed as findings against this node:

- **`STR_AMOUNT` is not a string.** Core maps it to `VNUM` and renders amounts as bare numbers; the name describes formatting, not the wire type. Guessing from names reported every amount field in every method as a type mismatch. The mapping is now transcribed from Core's `ExpectedType` in `rpc/util.cpp`, case for case.
- **`OBJ_DYN` keys are not literals.** `getindexinfo` declares `name` as a placeholder for whatever the keys turn out to be; comparing it as a key reported the placeholder missing and every real index extra.
- **The extractor was understating Core.** Fields appended by `list.emplace_back(...)` inside a conditional — `fullrbf` — have no leading brace and were missed, reporting a field Core has as one this node invented.

Each was checked against the Core source before being believed, in both directions.

## Where the line is

- **Top-level shape only.** Nested objects are not claimed: the extractor cannot verify them, and an unverified claim is worse than an absent one.
- **No values.** `differential_harness` stays `false` and `supported` stays unclaimable — a method whose shape matches can still answer a different number.
- **20 methods**, the ones that answer on a bare context. A fixture written to make a comparison pass is not evidence, so the list grows as fixtures that mean something get built. The count is asserted, so it cannot shrink unnoticed.

## Verification

`cargo fmt --all --check` clean. **CI's own clippy lanes** run directly — `-p bitcoin-rs --all-targets --features rocksdb,fjall,redb,mdbx,kernel -- -D warnings` and `-p bitcoin-rs-node --all-targets -- -D warnings` — both clean. `rpc` suite green, 11 compat tests.

Four mutations, all killed: a declared difference going stale; the manifest claiming a gap that is closed; `STR_AMOUNT` read as a string; `OBJ_DYN` keys compared as literals.

## Summary by Sourcery

Verify supported RPC result shapes against Bitcoin Core’s extracted declarations and keep every observed deviation synchronized with the compatibility manifest.

New Features:
- Add portable RPC result-shape compatibility checks against Bitcoin Core’s declared schemas for 20 methods and 23 representative invocations.
- Extract and embed Core RPC schemas with provenance, including top-level field names, JSON types, optionality, variants, and dynamic-object handling.

Bug Fixes:
- Correct schema extraction and comparison for amount fields, dynamic object keys, conditionally appended fields, and inherited ELISION fields.
- Detect both newly introduced schema differences and stale manifest deviations.

Enhancements:
- Record field-level schema deviations in the compatibility manifest while keeping value-level differential verification explicitly unsupported until a live Core reference is available.

Documentation:
- Document the distinction between schema verification and value-level differential testing, and define manifest fields for missing, extra, and mismatched result fields.

Tests:
- Add compatibility coverage and mutation tests for the compared RPC methods, schema parsing, type mapping, dynamic objects, optional fields, ELISION expansion, and extraction validation.


rabbitson87:
Reported the findings separately, as **#160**, and confirmed each against `main` at `2cf953f` rather than against this branch.

The three are real on `main`:

| method | Core declares non-optional | main emits |
|---|---|---|
| `getblockchaininfo` | `bits`, `target` | neither |
| `getmininginfo` | `bits`, `target`, `blockmintxfee`, `next` | none of the four |
| `getmempoolinfo` | `permitbaremultisig`, `maxdatacarriersize`, `limitclustercount`, `limitclustersize`, `optimal` | none of the five |

Plus `getmempoolinfo` emitting `mempool_sequence`, which Core carries on `getrawmempool`.

Reading `main` also turned up one the shape comparison **cannot** see, because both nodes have the field: `fullrbf` is emitted unconditionally here, while Core emits it only under `-deprecatedrpc=fullrbf`. That is in #160 too — worth knowing this instrument has that blind spot.

## And a correction to this PR

The manifest entries I wrote read as though each absence had been weighed and decided on. Nobody had — this comparison is what found them. One went further and explained why `bits` and `target` were absent, which dresses an omission as a judgement.

`deviation` in this manifest means "implemented, with a known and stated difference". These were differences nobody knew about, so they now say so and point at #160, where the decision gets made. Pushed as `dc4e8aa`.

Worth separating for whoever picks #160 up: `bits`/`target` are chain facts this node already holds, so emitting them is small. The five `getmempoolinfo` fields describe policy knobs this node may not have, and inventing a plausible number there would be worse than the omission — that one needs a decision, not a patch.

gosunuts:
Current-main review: this now overlaps with #176's typed `corepc-types::v31` response-contract approach. The useful part here is the independent comparison against Bitcoin Core's own declared RPC schemas, but it should not become a second source of truth for response construction.

If this PR is kept, I suggest narrowing it to a verification/oracle role on top of the chosen RPC contract. Otherwise, once #176 is finalized, this branch can be retired rather than merged unchanged.

metaphorics (commented):
Gate-and-merge hold: blocking Gate 4 findings, head `346188a` (conflict-resolved onto parent `feat/core-compat-manifest`, MERGEABLE).

Resolution note: the merge kept the parent's `compat_manifest.rs` layout, repointed the oracle at it (`schema.rs:370`), and deleted the resurrected `DEVIATIONS.md` (ledger retired) — verified, no markers, `cargo check -p bitcoin-rs-rpc --lib` passes.

Blocking, Gate 4 class 1 (false assurance on the oracle's core promise): in `crates/rpc/src/compat/schema.rs`, `type_matches` falls through to `_ => true` for any unknown kind tag (:96-106), so a field whose declared kind the comparator does not understand passes validation unconditionally. And in `tools/core-rpc-schema/extract.py`, a method whose result section yields no parseable `RPCResult` regions is skipped (`continue`), so an unparseable method is silently absent from the oracle rather than flagged. Together a Core change that renames a kind or restructures a result section degrades this check from enforcement to a green badge that proves nothing — on exactly the input (upstream drift) this PR exists to catch. These need author decisions (deny-unknown vs allowlist-with-warning; hard-fail vs explicit skip-list), so holding rather than redesigning here. No merge on this oid until then.

metaphorics (commented):
## Two blocking findings with patches

### Finding 1: `type_matches` falls through `_ => true` — unknown kinds pass unconditionally

**Location:** `crates/rpc/src/compat/schema.rs:96-106`

**Problem:** The catch-all `_ => true` means any type kind the oracle doesn't recognise is silently accepted as matching any JSON value. This is the exact false-pass the oracle exists to prevent: a future Core type (or a typo in the extracted schema) would make the oracle report zero differences for a field it never actually checked. The doc comment even acknowledges this by design ("a future Core adding a type must not turn into a spurious failure here"), but a spurious failure is visible and actionable — a false pass is invisible and is worse.

**Fix:** Deny unknown kinds (`_ => false`) while explicitly allowing Core's own opt-outs (`ANY`, `ELISION` → `true`). All 11 kinds present in the current `core-rpc-schema.json` (`ARR`, `ARR_FIXED`, `BOOL`, `NONE`, `NUM`, `NUM_TIME`, `OBJ`, `OBJ_DYN`, `STR`, `STR_AMOUNT`, `STR_HEX`) are already handled by explicit match arms, so no currently-passing fixture can break. The `_ => false` arm only fires for a *future* unknown kind — which is the safety property we want.

**Alternative considered:** Allowlist-with-warning (log unknown kinds but still return `true`). Rejected because the oracle has no logging surface and a warning that doesn't fail the check is indistinguishable from no check at all in CI. Deny-unknown is the option that cannot false-pass.

```diff
diff --git a/crates/rpc/src/compat/schema.rs b/crates/rpc/src/compat/schema.rs
--- a/crates/rpc/src/compat/schema.rs
+++ b/crates/rpc/src/compat/schema.rs
@@ -83,32 +83,41 @@ impl SchemaDiff {
 /// Whether a JSON value satisfies one of Core's declared result types.
 ///
 /// **Transcribed from Core's `ExpectedType` in `rpc/util.cpp`**, case for case,
 /// rather than inferred from the type names. That matters more than it looks:
 /// `STR_AMOUNT` is *not* a string. Core maps it to `VNUM` and renders amounts
 /// as bare numbers -- the name describes the formatting, not the wire type. A
 /// mapping built from the names reports every amount field in every method as a
 /// type mismatch, which is what mine did before I read the table.
 ///
 /// `ANY` and `ELISION` are Core's own opt-outs (`std::nullopt` there) and match
-/// anything. An unrecognised name does too: a future Core adding a type must
-/// not turn into a spurious failure here.
+/// anything. An unrecognised name does *not*: a future Core adding a type
+/// produces a mismatch the manifest must declare, rather than a silent pass
+/// that makes the oracle report nothing for a field it could not check. A
+/// spurious failure is visible and actionable; a false pass is invisible and
+/// is the failure mode this oracle exists to prevent.
 #[must_use]
 pub fn type_matches(kind: &str, value: &Value) -> bool {
     match kind {
         "NONE" => value.get_type() == JsonType::Null,
         "STR" | "STR_HEX" => value.get_type() == JsonType::String,
         "NUM" | "STR_AMOUNT" | "NUM_TIME" => value.get_type() == JsonType::Number,
         "BOOL" => value.get_type() == JsonType::Boolean,
         "ARR" | "ARR_FIXED" => value.get_type() == JsonType::Array,
         "OBJ" | "OBJ_DYN" => value.get_type() == JsonType::Object,
-        _ => true,
+        // Core's own opt-outs: `std::nullopt` in `ExpectedType`, matching anything.
+        "ANY" | "ELISION" => true,
+        // An unknown kind is a schema this oracle cannot check. Denying it
+        // surfaces a mismatch the manifest must declare; accepting it would
+        // silently pass any value, which is the false pass this oracle exists
+        // to prevent.
+        _ => false,
     }
 }
```

### Finding 2: `extract.py` silently skips methods with no parseable regions

**Location:** `tools/core-rpc-schema/extract.py:251-257` (the two `continue` statements in `main()`)

**Problem:** When `result_regions(body)` returns empty or `parse_result(regions)` returns `None`, the method is silently skipped with `continue`. A method whose `RPCResult` declaration drifts to a shape the extractor can't parse vanishes from `core-rpc-schema.json` entirely — the oracle then has no schema for it and reports nothing, which is indistinguishable from "matches perfectly." This is the same false-pass failure mode as Finding 1, in the extraction layer.

**Fix:** Collect skipped methods and hard-fail (`return 1`) with a diagnostic listing each one. A `NO_RESULT_METHODS` allowlist is provided for methods that genuinely declare no result (e.g. `stop`); landing a method there is an explicit decision, not a silent skip. The set starts empty — every method Core declares must either parse or be consciously allowlisted.

**Alternative considered:** A separate skip-list file. Rejected because the set is expected to be tiny (Core declares results for nearly every method) and a separate file is one more thing that can drift from the extractor's needs. An inline set next to `NONE_TYPES` keeps the policy visible at the point of use.

```diff
diff --git a/tools/core-rpc-schema/extract.py b/tools/core-rpc-schema/extract.py
--- a/tools/core-rpc-schema/extract.py
+++ b/tools/core-rpc-schema/extract.py
@@ -83,20 +83,27 @@ EXAMPLES_RE = re.compile(r"RPCExamples\{")
 # and its (usually empty) name.
 HEAD_RE = re.compile(
     r'\A\s*(?:"((?:[^"\\]|\\.)*)"\s*,\s*)?'
     r"RPCResult::Type::(\w+)\s*,\s*"
     r'"((?:[^"\\]|\\.)*)"'
 )
 
 # Core spells "this method returns nothing" a few ways.
 NONE_TYPES = {"NONE"}
 
+# Methods that genuinely declare no `RPCResult` (e.g. pure side-effect calls
+# like `stop`). The extractor hard-fails on any method it cannot parse; a
+# method landing here is an explicit decision that it returns nothing, not a
+# silent skip. Add a name only after confirming Core's source declares no
+# result schema for it.
+NO_RESULT_METHODS: set[str] = set()
+
 
 def method_blocks(source: str) -> list[tuple[str, str]]:
     """Split a translation unit into (method name, body) pairs.
 
     A body runs to the start of the next method definition, which is coarse but
     sufficient: the result declaration always precedes the handler lambda, and
     the handler cannot contain another `static RPCMethod` at column zero.
     """
     starts = [(m.group(1), m.start()) for m in METHOD_RE.finditer(source)]
     blocks = []
@@ -238,40 +245,58 @@ def main() -> int:
     parser.add_argument("--out", required=True, type=Path, help="schema JSON to write")
     args = parser.parse_args()
 
     rpc_dir = args.core / "src" / "rpc"
     if not rpc_dir.is_dir():
         print(f"no RPC sources under {rpc_dir}", file=sys.stderr)
         return 1
 
     methods: dict[str, dict] = {}
     sources = []
+    skipped: list[tuple[str, str]] = []
     for path in sorted(rpc_dir.glob("*.cpp")):
         source = path.read_text(encoding="utf-8", errors="replace")
         found = 0
         for name, body in method_blocks(source):
             regions = result_regions(body)
             if not regions:
+                if name in NO_RESULT_METHODS:
+                    continue
+                skipped.append((name, path.name))
                 continue
             schema = parse_result(regions)
             if schema is None:
+                if name in NO_RESULT_METHODS:
+                    continue
+                skipped.append((name, path.name))
                 continue
             # A name collision across files would silently drop one; Core has
             # none, and saying so out loud is cheaper than finding out later.
             if name in methods:
                 print(f"duplicate method {name} in {path.name}", file=sys.stderr)
                 return 1
             methods[name] = schema
             found += 1
         if found:
             sources.append({"file": f"src/rpc/{path.name}", "methods": found})
 
+    if skipped:
+        for name, file in sorted(skipped):
+            print(f"no parseable result schema for `{name}` in {file}", file=sys.stderr)
+        print(
+            f"{len(skipped)} method(s) had no parseable RPCResult region; "
+            "either the extractor drifted or the method genuinely declares none. "
+            "If the latter, add it to NO_RESULT_METHODS and rerun.",
+            file=sys.stderr,
+        )
+        return 1
+
     document = {
         "provenance": {
             "core_version": core_version(args.core),
             "extracted_from": "RPCResult declarations in the Core source tree",
             "sources": sources,
             "note": (
                 "Top-level result shape only: nested object and array fields are "
                 "not claimed, because this extractor cannot verify them. "
                 "Regenerate with tools/core-rpc-schema/extract.py."
             ),
```

### Verification

- `cargo check -p bitcoin-rs-rpc --lib` — passes (no warnings from changed files).
- `cargo test -p bitcoin-rs-rpc --lib compat::schema` — `coverage_is_recorded_so_it_cannot_silently_shrink` passes; `compared_methods_differ_from_core_exactly_as_declared` fails at `getmininginfo` ("does not answer on a bare context") — this is a **pre-existing failure** on the unmodified PR branch (confirmed by reverting the schema.rs change and observing the identical panic). The deny-unknown change introduces no new failures.
- All 11 type kinds in the current `core-rpc-schema.json` are handled by explicit match arms — none hit the `_ => false` catch-all — so no currently-passing fixture can break.
- `python3 -c "py_compile.compile('tools/core-rpc-schema/extract.py', doraise=True)"` — passes.

metaphorics:
@greptileai

metaphorics:
@sourcery-ai review

metaphorics:

metaphorics:

metaphorics:

metaphorics:

Issue #122, opened 2026-08-25
Keep the permanent test surface narrow around consensus and external compatibility


gosunuts:
## Principle

Tests are executable constraints. Keep the permanent test surface intentionally narrow, and do not use tests to freeze implementation outside Bitcoin consensus, external compatibility, durable storage/recovery semantics, and explicitly chosen product contracts.

A permanent test should carry information that must survive a substantially different implementation.

> **Tests define constraints, so add as few constraints as correctness requires.**
>
> In bitcoin-rs, the constrained surface should be unusually strong around consensus and compatibility, but unusually weak around internal architecture and algorithms.

This is not a request to reduce verification. Cheap implementation verification can be generated and run when code changes. The repository should permanently retain only the verification that also carries durable design information.

## The important distinction for bitcoin-rs

bitcoin-rs is many apparently small behaviors are actually Bitcoin contracts.

Do **not** delete a test merely because the implementation is simple or the test is unit-sized. If it protects Bitcoin consensus, canonical serialization, wire interoperability, Core parity, reorg/recovery correctness, or a durable on-disk format, it is high-value permanent executable documentation.

For example, the current `Tx::txid()` / `wtxid()` fixture test compares against the canonical `bitcoin` implementation. That is exactly the kind of permanent test we want: a completely different hashing implementation should still produce the same result.

Likewise the script stack's 1000-item maximum is a Bitcoin consensus constraint. The data structure used to enforce that limit is replaceable; the limit is not.

## Keep and strengthen

Keep tests and gates that encode durable externally meaningful behavior, especially:

- Bitcoin Core consensus-vector parity (`tx_valid`, `tx_invalid`, `script_tests`, `sighash`)
- consensus activation and script rules (BIP30/34/65/66/68/113/141/341/342, etc.)
- canonical transaction/block/hash serialization and endian semantics
- txid/wtxid/sighash equivalence with canonical implementations/vectors
- accepted/rejected block and transaction equivalence with Bitcoin Core/kernel where applicable
- full-IBD chainstate / UTXO parity, including MuHash evidence
- connect/disconnect and deep-reorg state equivalence
- snapshot import/export compatibility where the format is intentionally durable
- persistent storage reopen/crash-recovery/durability guarantees
- undo data sufficient to restore the exact prior chainstate
- P2P wire/protocol behavior required for interoperability
- RPC behavior intentionally compatible with Bitcoin Core or documented as a stable bitcoin-rs API
[...]

PR #161, opened 2026-08-26
fix(consensus): restore the kernel-parity lane on main

rabbitson87:
`kernel-parity` has been red on `main` since 30e8430 ("refactor(consensus): collapse transaction verification entrypoints"). Two mechanical leftovers of that refactor:

- **`verify_tx.rs:796`** still imports `verify_transaction_non_script` into the test module. The one remaining call site reaches it through `super::`, so the import is unused.
- **`verify_block.rs:473`** maps `|tx| tx.compute_txid()` where the method itself will do, which `redundant_closure_for_method_calls` refuses under `-D warnings`.

Both fail the lane's `cargo clippy -p bitcoin-rs-consensus --all-targets -- -D warnings` step. Because the errors are in a `lib test` target, the crate does not compile and **the lane never reaches its tests** — so the parity vectors have not run on `main` for as long as this has been broken.

## What changed since this PR was opened

It originally also carried an `fmt` fix. It no longer does. 19da503 ("style(consensus): satisfy Rust 1.95 formatting") landed the same reformatting on `main` independently, and rebasing onto 00938d8 absorbed it. The diff is now **2 files, 7 lines** — exactly the part `main` still needs, and nothing that duplicates work already done.

The title changed to match. Carrying a description that claims an `fmt` fix this branch no longer makes would be the smaller kind of lie, but still one.

## Verified

Run on this branch rebased onto `main` at 00938d8, using the lane's own steps rather than a paraphrase of them:

```
cargo fmt --all -- --check                                        ok
cargo clippy -p bitcoin-rs --all-targets -- -D warnings           ok
cargo clippy -p bitcoin-rs-consensus --all-targets -- -D warnings ok
cargo test -p bitcoin-rs-consensus --no-fail-fast -- --include-ignored  ok
```

The same two clippy errors reproduce on `origin/main` unpatched, so the lane's redness is confirmed present before and absent after, not assumed.

## No test

Deliberate. The constraint is already executable — it *is* the lane. A test asserting that source text satisfies a lint would be a second, weaker copy of a check CI already runs, and #122 asks for exactly the opposite: permanent tests should carry information that survives a different implementation, which this would not.

## Why separately

Nothing behavioural. A red lane costs every open PR its CI signal, and there are sixteen others. Landing this first is what makes their results mean anything.

Note also that the current `main` CI run has had `clippy`, `kernel-parity` and `bench-smoke` sitting in `in_progress` for nine hours — so `main`'s signal is currently both red *and* stalled. This PR addresses the red half only.

### DIFF
--- a/crates/consensus/src/verify_block.rs
+++ b/crates/consensus/src/verify_block.rs
@@ -470,7 +470,12 @@ mod tests {
         block
     }
     fn check_block_rules(block: &Block, context: BlockRuleContext) -> Result<(), ConsensusError> {
-        let txids: Vec<Txid> = block.0.txdata.iter().map(|tx| tx.compute_txid()).collect();
+        let txids: Vec<Txid> = block
+            .0
+            .txdata
+            .iter()
+            .map(bitcoin::Transaction::compute_txid)
+            .collect();
         let has_witness = block_has_witness(&block.0);
         verify_block_rules_precomputed(&block.0, context, &txids, has_witness)
     }
--- a/crates/consensus/src/verify_tx.rs
+++ b/crates/consensus/src/verify_tx.rs
@@ -793,7 +793,7 @@ mod tests {
 
     use super::{
         ScriptStageTimings, is_final_tx_with_locktime_cutoff, verify_coinbase_script_sig_size,
-        verify_transaction, verify_transaction_non_script,
+        verify_transaction,
     };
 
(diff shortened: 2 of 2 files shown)


gosunuts:
Closing as duplicate/already applied. The two remaining changes in this PR are already present on `main`, so merging it would not add any effective change.

PR #162, opened 2026-08-26
fix(wallet): declare the script backend psbt_roundtrip needs

### LINKED ISSUE #147: psbt_roundtrip passes only through workspace feature unification
## What happens

`crates/wallet/tests/psbt_roundtrip.rs` passes in CI and fails when the wallet crate is tested on its own:

```
$ cargo test -p bitcoin-rs-wallet --test psbt_roundtrip
Error: Script { input_index: 0, reason: "script verification failed: portable script backend cannot verify this non-taproot spend" }
```

Reproduced on `main` at `e7adc89`, so it is not a branch's doing.

## Why

The test calls `bitcoin_rs_consensus::verify_transaction`, which routes to `verify_non_taproot_portable` unless the kernel script backend is enabled. `bitcoin-rs-wallet` does not enable it and has no feature that would: its `[features]` are `rocksdb`, `fjall`, `redb`, `mdbx`.

CI runs `cargo test --workspace`, and in a workspace build Cargo unifies features across members — `bin/bitcoin-rs`'s default feature set pulls the kernel in, so `bitcoin-rs-consensus` gets it and the wallet's test passes on the way past.

So the test is green because of a dependency it does not declare, of a crate it does not name.

## Why it matters

- **The green is not evidence.** Nothing in the wallet crate says this test needs a real script backend, so nothing stops it silently becoming a no-op if the feature ever stops being unified in. It would keep passing.
- **It is not runnable where it is.** Anyone iterating on the wallet crate with `-p bitcoin-rs-wallet` sees a failure with a message about script backends, in a PSBT test, and has no reason to connect the two. I lost time to exactly that before checking whether `main` did the same.

## Options

1. **Declare it.** Give `bitcoin-rs-wallet` a `kernel` feature forwarding to `bitcoin-rs-consensus/kernel`, and gate the test on it — `#![cfg(feature = "kernel")]`. The test then runs when it can and is skipped, visibly, when it cannot.
2. **Make the requirement explicit at runtime.** Have the test assert the backend is the real one before proceeding, so the failure names the cause instead of the symptom.
3. **Move it** to a crate that does declare the kernel.

Option 1 is the smallest thing that makes the dependency honest, but any of the three beats the current state, where the test's outcome depends on how it was invoked.

## Not covered here

Whether `verify_non_taproot_portable`'s refusal is right — it is, and it says so. This is only about a test that cannot tell you which backend it exercised.

rabbitson87:
Closes #147.

`psbt_roundtrip.rs` verifies its finished transaction with `bitcoin_rs_consensus::verify_transaction`, which without the kernel routes to `verify_non_taproot_portable` — a stub accepting only bare `OP_TRUE`. The wallet crate declared no such dependency: `bitcoin-rs-consensus` is a dev-dependency taken with `default-features = false`, so `kernel` is off for a per-crate build.

It passed anyway, because `cargo test --workspace` unifies features across members and `bin/bitcoin-rs` turns the kernel on. **Green through a dependency it did not name, of a crate it did not mention** — while `cargo test -p bitcoin-rs-wallet` failed with `portable script backend cannot verify this non-taproot spend`: a message about script backends, in a PSBT test, with nothing connecting the two.

The wallet now has a `kernel` feature forwarding to `bitcoin-rs-consensus/kernel`, and the test is gated on it.

## Two things gating alone would have got wrong

**CI would have stopped running it.** Feature unification lifts a *dependency's* features, not the wallet's own — so `cargo test --workspace` would have skipped the gated file and reported the same unbroken green while checking strictly less. That is the original bug wearing a different hat. `crates/node`'s `kernel` feature now forwards to the wallet's, so anything turning the kernel on turns it on here too.

**A gated-out file is an empty binary, and an empty binary reports success** — which is the failure mode this is meant to remove. `psbt_roundtrip_gate.rs` compiles in its place when the feature is off, so the skip appears in the output as a named test rather than as nothing. It also asserts the portable backend really does refuse a real spend: if that ever stops being true, `psbt_roundtrip` is being skipped for a reason that no longer holds, and somebody finds out here instead of wondering why coverage went quiet.

## Verified in all three configurations

| invocation | roundtrip | gate |
|---|---|---|
| `-p bitcoin-rs-wallet --no-default-features` | skipped | **runs** |
| `-p bitcoin-rs-wallet --features kernel` | **runs** | skipped |
| `cargo test --workspace` (what CI runs) | **runs** | skipped |

Each was executed, not reasoned about — the middle row is the one that would have silently regressed.

## Note on CI

This branch is cut from `main` at `2cf953f`, where the `fmt` and `kernel-parity` jobs are already failing for reasons unrelated to it (`30e8430`). #161 fixes those; expect those two jobs red here until it lands.

### DIFF
--- a/crates/node/Cargo.toml
+++ b/crates/node/Cargo.toml
@@ -80,7 +80,7 @@ mdbx = [
     "bitcoin-rs-index/mdbx",
     "bitcoin-rs-storage/mdbx",
 ]
-kernel = ["bitcoin-rs-consensus/kernel"]
+kernel = ["bitcoin-rs-consensus/kernel", "bitcoin-rs-wallet/kernel"]
 checksig-census = [
     "kernel",
     "bitcoin-rs-consensus/checksig-census",
--- a/crates/wallet/Cargo.toml
+++ b/crates/wallet/Cargo.toml
@@ -37,6 +37,10 @@ tempfile = "3"
 
 [features]
 default = []
+# The real script backend, for the one test that verifies a finished
+# transaction against consensus. Off by default, as `bitcoin-rs-consensus`'s
+# own workspace declaration is, so a portable build stays portable.
+kernel = ["bitcoin-rs-consensus/kernel"]
 rocksdb = ["bitcoin-rs-storage/rocksdb"]
 fjall = ["bitcoin-rs-storage/fjall"]
 redb = ["bitcoin-rs-storage/redb"]
--- a/crates/wallet/tests/psbt_roundtrip.rs
+++ b/crates/wallet/tests/psbt_roundtrip.rs
@@ -1,4 +1,35 @@
 //! PSBT build, external signing, finalization, and consensus roundtrips.
+//!
+//! **Gated on the `kernel` feature, because it needs the real script backend.**
+//! The last step hands the finished transaction to
+//! `bitcoin_rs_consensus::verify_transaction`, which without the kernel routes
+//! to `verify_non_taproot_portable` -- a stub that accepts only bare `OP_TRUE`
+//! and refuses everything else.
+//!
+//! Before the gate this crate declared no such dependency and the test passed
+//! anyway, because `cargo test --workspace` unifies features across members and
+//! `bin/bitcoin-rs` turns the kernel on. So it was green through a dependency it
+//! did not name, of a crate it did not mention -- and `cargo test -p
+//! bitcoin-rs-wallet` failed with "portable script backend cannot verify this
+//! non-taproot spend", a message about script backends, in a PSBT test, with
+//! nothing connecting the two.
+//!
+//! Two things follow, and both matter:
+//!
+//! - `crates/node`'s `kernel` feature now forwards to this crate's, so anything
+//!   turning the kernel on turns it on here too. `cargo test --workspace` still
+//!   runs this.
+//! - A gated-out file is an empty binary, and an empty binary reports success --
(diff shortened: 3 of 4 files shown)


rabbitson87:
Closing against the new repository boundary.

`psbt_roundtrip` exercises `PsbtBuilder`, `ExternalSigner` and `finalize_signed` — PSBT construction, signing and finalisation. That is wallet functionality, and under the boundary being drawn ("up to the wallet-facing interface; wallet functionality to its own repository") it leaves this repository along with the code it covers.

Fixing the feature declaration of a test that is being deleted is work on the wrong side of the line, and there is no smaller version of this PR: the feature, the gate and the gate-visibility test all exist only to serve that one test.

## What does not go away

The defect #147 describes is real and stays open, retargeted: a test that is green **through a dependency it does not declare** is green for the wrong reason, and the resolution is now the boundary move rather than a `kernel` feature. Whoever moves the wallet out should check the migrated crate declares its own script backend rather than inheriting one by feature unification, or the same problem follows it across.

## One thing worth carrying over

The general shape is not wallet-specific and may be worth applying elsewhere: **a gated-out test file compiles to an empty binary, and an empty binary reports success.** Gating a test on a feature without something in its place converts a loud failure into a silent skip. If any other test ends up feature-gated during the boundary work, it is worth pairing with a companion that keeps the skip visible.

Reopen if the boundary move turns out to be further off than it looks — the fix is small and stops `cargo test -p bitcoin-rs-wallet` misleading anyone in the meantime.

PR #165, opened 2026-08-26
refactor: remove in-tree wallet implementation

### LINKED ISSUE #163: Move the wallet implementation out of the repository; stop at the wallet-facing interface
## Goal

Move the in-tree wallet implementation out of this repository. The repository boundary stops at the **wallet-facing interface**; wallet functionality lives in its own repository and consumes bitcoin-rs the way any external wallet does.

## Rationale

- A wallet does not run attached to a Bitcoin node's lifecycle. Nothing about block sync, validation, or chainstate needs it present.
- Wallet code raises repository complexity — key handling, PSBT, signing, coin selection, fee policy — none of which the node's own correctness depends on.
- The compatibility target for application-facing data is the **Esplora API**, and Esplora is a projection over node state. It needs `TxLookup`/`ScriptIndex`, not a wallet.

The pinned reference draws the same line: **Bitcoin Core's kernel tree has no `src/wallet/` at all.** Where Core places an RPC is therefore a usable criterion for what stays.

| method | Core's file | boundary |
| --- | --- | --- |
| `getdescriptorinfo`, `deriveaddresses` | `src/rpc/output_script.cpp` | stays |
| `scantxoutset` | `src/rpc/blockchain.cpp` | stays |
| `combinepsbt`, `finalizepsbt` | `src/rpc/rawtransaction.cpp` | stays |
| `walletcreatefundedpsbt`, `walletprocesspsbt`, `bumpfee` | absent from the kernel tree | leaves |
| the 12 disabled private-key methods | absent from the kernel tree | leaves |

Note that the `wallet` group in the RPC dispatcher is a filing convenience, not a boundary: 15 of its 20 methods are node RPCs. **Reclassify rather than move wholesale.**

## What the coupling actually is

Measured on `main`, not assumed:

- `crates/node` declares `bitcoin-rs-wallet` and forwards `rocksdb`/`fjall`/`redb`/`kernel` to it, but **does not reference a single item from it**. The dependency is feature plumbing only.
- `crates/rpc`'s only use of the crate is `bump_psbt_with_rate_sat_per_kvb` and `WalletError::Bip125`, both in `bumpfee` — a method that leaves.
- `descriptor_checksum` already lives in `crates/rpc/src/handlers/wallet.rs`, so the descriptor RPCs that stay do not depend on the wallet crate today.

So the cut is clean on `main`. #110 is the thing to watch: it introduces `crates/wallet/src/descriptor.rs`, which would put a node RPC's implementation inside the departing crate. Land the module where it belongs, or move it back as part of this work.

## Scope

- Remove `crates/wallet` from the workspace, and its entry from the root `Cargo.toml`.
- Remove the unused `bitcoin-rs-wallet` dependency and feature forwarding from `crates/node`.
- Remove `bumpfee`, `walletcreatefundedpsbt`, `walletprocesspsbt` and the disabled private-key methods from the RPC dispatcher.
- Reclassify the remaining `wallet`-group methods into the groups Core files them under.
- Update `docs/api/core-compat.toml` **in the same commit**. `the_manifest_and_the_dispatcher_name_the_same_methods` is enforced in both directions, so the manifest cannot go stale here and cannot be updated ahead of the code either.
- Keep `TxLookup`, `ScriptIndex`, Electrum, Esplora, REST, ZMQ and the descriptor/PSBT-assembly RPCs that Core files outside the wallet unchanged.

## Carried over from closed issues

- **#112** described building `bitcoin-rs-wallet` into the native Rust wallet against a wallet-facing interface. The interface half survives; the implementation half becomes the separate repository's work, with btctxbuilder as the first external consumer and reference.
- **#147** recorded that `psbt_roundtrip` was green only through workspace feature unification — the wallet crate did not declare the script backend it needed. **The defect migrates with the code:** whoever stands up the new repository must confirm it declares its own backend, because a `cargo test -p` there has no other crate to borrow the feature from. Two things learned there are worth carrying: `--workspace` unifies *dependency* features across members, not a crate's own; and a `#![cfg]`-gated-out test file compiles to an empty binary, which reports success.

## Out of scope

- Deciding the new repository's architecture, name, or wallet feature set.
- Changing what the node exposes. If a wallet needs something the interface does not provide, that is a separate request against the interface, not a reason to keep wallet code here.

gosunuts:
## Summary

The node no longer carries an attached wallet lifecycle or wallet-only key, coin-selection, and fee policy. External wallets retain the wallet-facing node interface: descriptor helpers, UTXO scanning, PSBT combination/finalization, and the existing Electrum, Esplora, REST, ZMQ, `TxLookup`, and `ScriptIndex` surfaces.

The retained RPCs now live with the Core-aligned node handlers that own their state, so the repository boundary is explicit without shrinking what external wallets can query or assemble.

## Verification

- `cargo metadata --format-version 1 --no-deps | jq -e 'all(.packages[]; .name != "bitcoin-rs-wallet")' && test ! -e crates/wallet`: passed (`true`; wallet crate absent).
- `cargo test -p bitcoin-rs-rpc --test handler_smoke removed_wallet_methods_return_method_not_found -- --exact`: passed (1 test).
- `cargo test -p bitcoin-rs-rpc --test handler_smoke all_required_handlers_return_core_shapes -- --exact`: passed (1 test).
- `cargo test -p bitcoin-rs-rpc --no-fail-fast`: passed (290 tests across 5 suites).
- `cargo fmt --all -- --check`: passed.
- `cargo test --workspace --no-fail-fast`: passed (1,615 tests; 12 ignored).
- `export FULL_NODE_FEATURES=rocksdb,fjall,redb,mdbx,kernel && cargo test -p bitcoin-rs --no-fail-fast --no-default-features --features "$FULL_NODE_FEATURES"`: passed (119 tests; 9 ignored).
- `export FULL_NODE_FEATURES=rocksdb,fjall,redb,mdbx,kernel && cargo clippy -p bitcoin-rs --all-targets --no-default-features --features "$FULL_NODE_FEATURES" -- -D warnings`: passed.
- `cargo clippy -p bitcoin-rs-node --all-targets -- -D warnings`: passed.
- `cargo clippy -p bitcoin-rs-consensus --all-targets -- -D warnings`: reproduced the same two-error failure observed on `origin/main`.

## Notes

- GitHub Actions run [#32924653546](https://github.com/gosuda/bitcoin-rs/actions/runs/32924653546) for `59c1edab9b08` passed `fmt`, `clippy`, `test`, `bench-smoke`, `portable-check`, and `deny`; `kernel-parity` failed on the two consensus clippy errors at `crates/consensus/src/verify_tx.rs:796` and `crates/consensus/src/verify_block.rs:473`.
- The latest `main` CI run [#32920013422](https://github.com/gosuda/bitcoin-rs/actions/runs/32920013422) fails the same `kernel-parity` command with the same two errors. `git diff --quiet origin/main...HEAD -- crates/consensus` also passed, proving the failure is outside this removal diff.
- `docs/api/core-compat.toml` and `the_manifest_and_the_dispatcher_name_the_same_methods` do not exist on current `main`, so this removal does not invent a new manifest schema or parity framework.
- PR #110 remains unmerged. When rebased after this boundary change, its descriptor-analysis implementation must move to a node-side module rather than restore `crates/wallet`.

Closes #163


metaphorics:
@codex review

Issue #164, opened 2026-08-26
Merge pruning and coinstats into their owning core crates


gosunuts:
## Goal

Remove `crates/pruning` and `crates/coinstats` as standalone workspace crates, while keeping the useful behavior in the core crates that already own the underlying state.

## Rationale

These two crates do not look like independent subsystems with their own durable state or lifecycle.

- **Pruning** is a retention policy over authoritative block/undo storage and chainstate safety. It is tightly coupled to storage layout, applied height, reorg requirements, and node configuration. Keeping it as a separate crate creates an orchestration layer over `storage`, `utxo`, and `chain` rather than a reusable domain boundary.
- **Coinstats** is derived computation over authoritative UTXO state. Its current dependency shape is effectively `primitives + storage + utxo`, and functionality such as UTXO statistics / MuHash belongs naturally with the UTXO owner rather than in a separate package.

This is about simplifying crate boundaries, not removing pruning or UTXO statistics functionality.

### Pruning

- Move pruning policy and implementation into the appropriate `storage` / `node` modules.
- Keep node configuration responsible for selecting pruning behavior and retention targets.
- Keep storage responsible for safe deletion / retention of block and undo data.
- Preserve reorg, restart, crash-recovery, and body-availability guarantees.
- Remove `bitcoin-rs-pruning` from workspace manifests and dependency / feature plumbing.
- Reconcile with #144 so Utreexo-specific pruning paths disappear with Utreexo instead of being preserved during the move.

### Coinstats

- Move `coinstats` functionality into `crates/utxo` (for example a `stats` / `muhash` module).
- Preserve externally visible RPC behavior such as UTXO-set statistics where still part of the compatibility surface.
- Keep any performance-sensitive implementations and benchmarks that remain useful, but colocate them with the UTXO state they operate on.
- Remove `bitcoin-rs-coinstats` from workspace manifests and dependency / feature plumbing.

## Boundary rule

A standalone crate should represent an independently owned subsystem: its own state, lifecycle, failure boundary, or a reusable dependency direction that benefits from compile-time separation.

Pruning changes the lifecycle of authoritative storage itself, while coinstats reads authoritative UTXO state. Neither currently satisfies that boundary strongly enough to justify a separate crate.

## Acceptance criteria

- [ ] `crates/pruning` is removed without changing pruning semantics or safety guarantees.
- [ ] Pruning ownership is clearly split between node policy and storage retention.
[...]

PR #167, opened 2026-08-26
refactor(utxo): fold coinstats into the crate that owns the state it reads

rabbitson87:
Closes the `coinstats` half of #164. The `pruning` half is deliberately not here — see the last section.

## The dependency edge disappears rather than moving

#164's argument is that `coinstats` is derived computation over authoritative UTXO state, not an independently owned subsystem. Its manifest already said so: `primitives + storage + utxo`, where the `utxo` edge existed only so it could read the set it summarises.

So the three source files move into `crates/utxo/src/stats/`, and every `bitcoin_rs_utxo::` path inside them collapses to `crate::`. That is the whole point of the change — the edge does not get relocated, it stops existing.

| | |
| --- | --- |
| `crates/coinstats/src/stats.rs` | → `crates/utxo/src/stats/coin_stats.rs` |
| `crates/coinstats/src/muhash3072.rs` | → `crates/utxo/src/stats/muhash3072.rs` |
| `crates/coinstats/src/persist.rs` | → `crates/utxo/src/stats/persist.rs` |
| 3 test files, 1 hot-path bench | → `crates/utxo/tests/`, `crates/utxo/benches/` |

`crates/utxo` gains **exactly one** dependency: `ruint`, used by the `MuHash3072` test oracle. It already had the other eleven, and the two crates' feature tables were already character-identical — `bench-mimalloc` plus the four backend forwards — so nothing needed reconciling.

One inconsistency fell out on the way. `crates/rpc` was forwarding storage backends through `bitcoin-rs-coinstats/{rocksdb,fjall,redb}` while depending on `bitcoin-rs-utxo` and forwarding nothing to it. Those now point at `bitcoin-rs-utxo` directly: same reachable feature set, one hop shorter, and it no longer depends on a crate it names for no other reason.

## The one thing that could have broken a datadir

```rust
const COINSTATS_CODEC: &str = "bitcoin-rs-coinstats";
```

This spells a crate name, but it is not a crate reference. It is written into `manifest-v1.json` and matched on load, so it is an **on-disk value**. Renaming it to match the new module would leave any single binary perfectly self-consistent — writer and reader use the same constant — while failing the codec check on every checkpoint already written, falling back to a full resync (`docs/policies/db-migration.md` §2.4).

**No existing test could catch that**, precisely because the constant is only ever compared to itself. So this PR adds one test:

```rust
fn manifest_codec_identifiers_are_frozen() {
    assert_eq!(super::HEADER_CODEC, "bitcoin-rs-canonical-headers");
    assert_eq!(super::UTXO_CODEC, "bitcoin-rs-utxo-spendable-v1");
    assert_eq!(super::COINSTATS_CODEC, "bitcoin-rs-coinstats");
    ...
}
```

Literals, not the constants — comparing a constant to itself proves nothing. The constant, `db-migration.md` §2.4, and `crates/utxo/README.md` all now say why the name outlives the crate, because the next person to see it will read it as leftovers and tidy it.

This is the only test added. It is the kind #122 asks for: an on-disk format identifier is externally meaningful and must survive a different implementation. Everything else here is a move, and the tests that came with `crates/coinstats` are its evidence.

## Verified

Lanes CI actually runs, on this branch:

```
  --features "rocksdb,fjall,redb,mdbx,kernel" -- -D warnings            ok
cargo clippy -p bitcoin-rs-node --all-targets -- -D warnings            ok
cargo test -p bitcoin-rs-utxo                                           154 passed
cargo test --workspace --no-fail-fast                        113 targets ok
```

**Two `--workspace` targets fail, and they fail on `main` too**: `g14_guarded_run` (8 of 11) and `g2_bitcoind_muhash_script` (3 of 3). Both are the known CRLF-checkout failure on a Windows working tree — the shell scripts they drive get `\r\n` — which is what #142 fixes. Baseline captured on a clean `origin/main` worktree rather than assumed; neither target touches UTXO statistics.

Also worth recording, since I hit it while checking the above: `cargo clippy -p bitcoin-rs-rpc --all-targets -- -D warnings` fails on clean `main` with 28 errors (11 `Result::expect`, 8 `Option::expect`, 5 `as` conversions, 2 disallowed `std::sync::Mutex`, …). CI lints only `-p bitcoin-rs` and `-p bitcoin-rs-node`, so the rpc crate's own lib-test, bench and `handler_smoke` targets are unlinted. Unrelated to this change and not fixed here.

## Not in this PR: `crates/pruning`

#148 is currently deleting `utreexo_only.rs` and `utreexo_no_blocks.rs` from that crate. Moving it now would make that PR unmergeable, and #164 asks for the reconciliation in that order.

Worth flagging one finding for whoever picks it up, because it changes the shape of the work. #164 describes pruning as "an orchestration layer over `storage`, `utxo`, and `chain`". The manifest supports that reading — five `bitcoin-rs-*` dependencies. The source does not:

```
bitcoin_rs_storage      11 references
bitcoin_rs_primitives    5 references   (all of them `Hash256`)
bitcoin_rs_utxo          0
bitcoin_rs_chain         0
bitcoin_rs_utreexo       0
bitcoin (rust-bitcoin)   0
```

**That boundary is an artefact of unused manifest entries.** Pruning never touches `utxo` or `chain`; its real shape is `storage` plus one type. So #164's acceptance criterion "pruning ownership is clearly split between node policy and storage retention" may not need a split at all — the mechanics are a pure storage concern, and `PrunePolicy` is already a plain value the node hands in.

There is a stronger argument for the move than the issue makes, too: `bitcoin_rs_pruning::block_body_key` and `BLOCK_DATA_CF` are used by `apply.rs`'s **ordinary block-read path**. The pruning crate already owns storage's block-body key schema, and code with nothing to do with pruning borrows it from there.


rabbitson87:
`kernel-parity` is red here and it is not this branch. The lane's third step fails with:

```
error: unused import: `verify_transaction_non_script`
error: redundant closure
error: could not compile `bitcoin-rs-consensus` (lib test) due to 2 previous errors
```

Those are the two errors #161 fixes, inherited from `main` — this branch does not touch `crates/consensus` at all. **Every PR branched off current `main` will show the same failure**, which makes this more than a queue nuisance: a new PR's kernel-parity result is currently uninformative, because it is red before the change is even applied.

The other eleven checks pass, including `clippy`, `test`, `portable-check` and `deny`.

Worth noting the failure shape, since it is why nobody caught this earlier: because the clippy errors are in a `lib test` target, the crate never compiles, so the lane's `cargo test -p bitcoin-rs-consensus -- --include-ignored` step never runs. The consensus parity vectors have not executed for as long as this has been broken. That is the actual cost, not the red badge.

Nothing to do on this branch — it rebases onto a fixed `main` cleanly.

PR #168, opened 2026-08-26
ci: run on stacked pull requests, and bound the runs that follow

rabbitson87:
## Five open PRs have never been built

`pull_request: branches: [main]` filters on the pull request's **base**, so a stacked PR — one targeting the branch below it — never triggers CI:

```
main
 └ #101 feat/mempool-acceptance          → main      ✅ CI runs
    └ #102 mining-package-selection      → #101      ❌ never built
       └ #103 rpc-getblocktemplate       → #102      ❌ never built
          └ #105 rpc-gbt-proposal        → #103      ❌ never built
             └ #109 rpc-mininginfo-next  → #105      ❌ never built

 └ #126 feat/core-compat-manifest        → main      ✅ CI runs
    └ #150 core-rpc-schema-oracle        → #126      ❌ never built
```

**This is worse than a red lane, because it is silent.** Those PRs report no failing checks; `gh pr checks` lists CodeRabbit and nothing else. A PR with no failing checks reads as a PR that passed. It is not a skipped run or a cancelled one — `gh run list --commit <sha>` returns nothing at all for every one of those five head commits.

So five changes to mempool acceptance, mining package selection, `getblocktemplate`, block-proposal handling and the RPC schema oracle are sitting in the queue having never been compiled, linted, or tested by CI, and looking fine.

## The change

`branches: ['**']` covers every base. A stacked PR is then built as its head merged into its own base, which is the right thing to test — the parent's content is already in that merge, so each PR is checked against the stack it actually sits on.

The `push` trigger is untouched; this only changes which pull requests are built.

## Why `concurrency` is in the same commit

Because this change makes runs more numerous, and adding load without bounding it would be leaving half the job done. A superseded pull-request run is now cancelled when a newer commit lands on the same ref, which is roughly the load the wider trigger adds back.

```yaml
concurrency:
  group: ci-${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

`cancel-in-progress` is conditional on purpose. **A push to `main` must never be cancelled** — that run is the record for that commit, and unlike a pull request there is no later run coming to replace it. An unconditional `cancel-in-progress: true` would silently lose `main`'s history whenever two commits landed close together.

Happy to split this out if it should be argued separately.

## Verified

The workflow parses to:

```
triggers:    {'push': {'branches': ['main']}, 'pull_request': {'branches': ['**']}}
concurrency: {'group': 'ci-${{ github.workflow }}-${{ github.ref }}',
              'cancel-in-progress': "${{ github.event_name == 'pull_request' }}"}
jobs: 8      bench-smoke clippy deny fmt kernel-parity portable-check test wallet-no-seckey
```

All eight jobs still present, expressions well-formed.

The real proof is what happens to #102/#103/#105/#109/#150 after this merges: they should each pick up a full run for the first time. Worth expecting some of them to come back red — that is the point of the change, and a red result on work that was never checked is information, not a regression.

## Related

A second blind spot found while checking this, filed here rather than fixed: `cargo clippy -p bitcoin-rs-rpc --all-targets -- -D warnings` fails on clean `main` with 28 errors (11 `Result::expect`, 8 `Option::expect`, 5 `as` conversions, 2 disallowed `std::sync::Mutex`, …). CI lints only `-p bitcoin-rs` and `-p bitcoin-rs-node`, so the rpc crate's own lib-test, bench and `handler_smoke` targets are unlinted. Different problem, separate fix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,24 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main]
+    # Every pull request, not only those based on `main`. A stacked PR targets
+    # the branch below it, and `branches: [main]` excluded those by
+    # construction -- five open PRs had never been built or tested, while
+    # reporting no failing checks, which reads exactly like green.
+    branches: ['**']
+
+# Bounded because the trigger above is wider: a superseded pull-request run is
+# cancelled when a newer commit lands on the same PR.
+#
+# `main` pushes are put in a group of their own, keyed by `run_id`, rather than
+# relying on `cancel-in-progress: false`. That flag only protects a run that is
+# already executing: with a shared group, a `main` run that is still *pending*
+# behind an in-progress one is cancelled when a third push arrives, and its
+# commit then has no CI record at all. A unique group per run means `main`
+# pushes never queue against each other and so are never cancelled.
+concurrency:
+  group: ci-${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.ref || github.run_id }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 env:
   CARGO_TERM_COLOR: never


gosunuts (changes requested):
The stacked-PR trigger change is correct and worth landing, but the concurrency semantics do not quite match the stated intent.

`cancel-in-progress: false` only preserves an already-running `main` run. Workflow-level concurrency still applies to `main`, so if one `main` run is running, a second is pending, and a third push arrives, GitHub may replace/cancel the older pending run. That means `pushes to main are never cancelled` is not guaranteed by the current group.

Please make `main` pushes use a unique concurrency group (for example `github.run_id`) while PR runs continue to group by their PR ref, so only superseded PR runs are cancelled.

Also, please rebase onto current `main` before merging. #165 removed `wallet-no-seckey`, but this branch's workflow still contains that obsolete job. After rebasing, rerun CI against the actual current workflow.

Once those two points are addressed, this looks good to me.

rabbitson87:
Both points addressed, and you were right on the first one in a way I want to state precisely, because my comment claimed the opposite.

**`cancel-in-progress: false` does not mean "never cancelled".** It only protects a run that is already *executing*. With a shared group, a `main` run sitting **pending** behind an in-progress one is cancelled when a third push arrives — and that commit then has no CI record at all, which is the exact failure the block was supposed to prevent. So the code did not do what its own comment said.

Fixed as suggested, with `main` in a group of its own:

```yaml
concurrency:
  group: ci-${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.ref || github.run_id }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

`github.run_id` is unique per run, so `main` pushes never share a group and therefore never queue against each other — nothing is left for the pending-cancellation rule to act on. PR runs keep grouping by `github.ref`, so a superseded run is still cancelled.

One detail worth flagging since it is invisible in review: I first wrote this as a folded block scalar (`group: >-`) across three lines for readability. YAML folding **keeps** the newlines when continuation lines are more-indented, so the group string would have contained a literal `\n` inside `${{ }}`. It is one line now, and the check below asserts that rather than trusting it.

**Rebased onto current `main`** (`3769a03`, after #165). `wallet-no-seckey` is gone from this branch's workflow along with the rest — the job count is 7, not 8. CI is re-running against the actual current workflow.

Parse check after the change:

```
triggers: {'push': {'branches': ['main']}, 'pull_request': {'branches': ['**']}}
group:    "ci-${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.ref || github.run_id }}"
newline in group: False
jobs: 7
```

Separately, and only because it changes what this PR is worth: **#161 is closed** — those two clippy fixes reached `main` by another route, so `kernel-parity` is no longer inherited-red. That means the five stacked PRs this unblocks will get a *meaningful* first result rather than one that was red before their change was applied.

gosunuts (approved):
Reviewed current head 877a7421a8f17ffed444c238e66c8fe0167db81a. The stacked-PR trigger uses branches: ['**']; PR runs group by ref and cancel only superseded PR runs, while non-PR runs use unique github.run_id groups. The obsolete wallet-no-seckey job is absent, the branch is based on the post-#165 main commit, and all current checks pass. The requested changes are addressed.

Issue #160, opened 2026-08-25
RPC results omit fields Core declares non-optional (getblockchaininfo, getmininginfo, getmempoolinfo)


rabbitson87:
Three RPC methods omit fields Bitcoin Core declares **non-optional**, and one emits a field Core carries on a different method. Found by comparing this node's answers against Core's own `RPCResult` declarations (#150); confirmed by reading `main` at `2cf953f`.

None of these were recorded anywhere, so they are gaps rather than decisions. Filing them separately from the harness that found them, because closing them is not the harness's job.

## Why Core's declarations are a fair reference

Core declares each method's result shape beside its handler, and `RPCHelpMan::Check` asserts Core's own handlers against those declarations in debug builds. They are kept true by Core, not by a doc writer. Optionality is part of the declaration — `/*optional=*/true` — so "Core declares this non-optional" means Core always emits it.

---

## 1. `getblockchaininfo` omits `bits` and `target`

`crates/rpc/src/handlers/chain.rs`. Emitted keys:

```
chain blocks headers bestblockhash difficulty time mediantime
verificationprogress initialblockdownload chainwork size_on_disk
pruned pruneheight warnings
```

Core declares, non-optional: `bits` (the tip's compact target) and `target` (its full target).

`difficulty` is present, but it is a derived float — a client cannot recover the compact target from it, which is what anything checking or reproducing the tip's proof-of-work needs.

## 2. `getmininginfo` omits `bits`, `target`, `blockmintxfee` and `next`

`crates/rpc/src/handlers/mining.rs`. Emitted keys:

```
blocks currentblockweight currentblocktx difficulty networkhashps
pooledtx chain warnings
```

Core declares, non-optional: `bits`, `target`, `blockmintxfee`, `next`.

These are the numbers a miner works against, and `next` in particular describes the block about to be mined rather than the one just accepted. **Already being repaired in #109**; listed here for completeness of the class.

## 3. `getmempoolinfo` omits five policy fields, and emits one of another method's

`crates/rpc/src/handlers/mempool.rs`. Emitted keys:

```
loaded size bytes usage total_fee maxmempool mempoolminfee
minrelaytxfee incrementalrelayfee mempool_sequence unbroadcastcount fullrbf
```

Core declares, non-optional and absent here: `permitbaremultisig`, `maxdatacarriersize`, `limitclustercount`, `limitclustersize`, `optimal`.

[...]

PR #176, opened 2026-08-26
refactor(rpc): use corepc v31 response contracts

### LINKED ISSUE #173: Use corepc-types as the Bitcoin Core RPC wire contract
`crates/rpc` currently hand-builds many Bitcoin Core-compatible JSON responses with `json!` / `Object`, and `core_compat.rs` separately hand-maintains expected key sets. This duplicates the Core RPC schema in both implementation and tests, and can let both drift in the same direction.

## Relationships

- **Parent contract:** #78
- **Boundary:** #172 keeps bitcoin-rs protocol primitives native; `corepc-types` is only a wire-level RPC adapter here.
- **Constrained by:** #122 (permanent test surface), #179 (crate hierarchy)

Use [`corepc-types`](https://github.com/rust-bitcoin/corepc) as the wire-level response contract for the supported Bitcoin Core version instead.

Suggested direction:

- Add `corepc-types` to `crates/rpc`.
- Target a specific Core RPC version explicitly (currently v31 is the natural target).
- For RPCs with versioned response types, construct `corepc_types::v31::*` values directly from bitcoin-rs node state and serialize them at the RPC boundary.
- Keep bitcoin-rs internal chain/mempool/network types unchanged; `corepc-types` should only be the wire adapter layer.
- Keep request parsing, dispatch, authentication, JSON-RPC framing, error mapping, and method semantics in bitcoin-rs.
- Replace manual key-set compatibility tests with deserialization / schema checks against the same versioned `corepc-types` types.
- Use direct bitcoind differential tests for behavior that schema types cannot cover, such as error codes, defaults, reorg behavior, and value semantics.

This should remove a significant amount of duplicated Core response-shape code and make RPC drift visible at compile/test time when Core-compatible structs change.

gosunuts:
## Summary

RPC response objects were assembled by hand, while `core_compat.rs` maintained a second copy of the Core schema. That allowed implementation and compatibility checks to drift together. This PR makes the published `corepc_types::v31` models the Bitcoin Core v31 response contract at the RPC boundary without changing request parsing, dispatch, authentication, JSON-RPC framing, or internal node types.

- Adds `corepc-types` 0.15, explicitly targeting Core v31.
- Migrates chain, mempool, network, mining, transaction, and utility response objects to versioned `corepc_types::v31` values.
- Keeps primitive-only and unsupported response shapes manual where no matching corepc result type exists.
- Replaces hand-maintained compatibility key lists with typed deserialization checks.
- Adds the required v31 fields for `getblockchaininfo`, `getmininginfo`, and `getmempoolinfo`, and keeps optional Core fields omitted when unavailable.

The migration covers the response-contract portion of #173; RPC behavior and value semantics remain owned by bitcoin-rs.

Closes #173

## Validation

- `git diff --check`
- `cargo clippy -p bitcoin-rs-rpc --lib -- -D warnings`
- `cargo test -p bitcoin-rs-rpc --no-fail-fast` — 298 passed across 5 test targets
- Typed `corepc_types::v31` compatibility tests — 6 passed


gosunuts (commented):
The `corepc-types` dependency itself looks like the right direction here. The remaining issue is semantic mapping, not the response types.

Two mappings should be corrected before treating these endpoints as Core v31-compatible:

- `getpeerinfo.startingheight` is currently populated unconditionally. In Core 31 this is deprecated and omitted by default, so this should be `None` unless bitcoin-rs explicitly supports the deprecated RPC option.
- `getmempoolinfo.limit_cluster_count` / `limit_cluster_size` should not be populated from bitcoin-rs ancestor limits. Core 31's cluster limits describe different policy semantics, and `chunk_weight` / `fees.chunk` likewise come from Core's cluster linearization. Mapping ancestor limits into those fields, or treating each transaction as a singleton chunk, produces correctly-shaped but semantically incorrect v31 responses.

I would keep `corepc-types` as the wire contract and fix the adapter/internal semantics instead of recreating local RPC structs. Where bitcoin-rs does not implement the corresponding Core concept, prefer `None` when the type allows it rather than filling the field with a nearby-but-different value.

rabbitson87:
Both findings are addressed in #201, stacked on this branch so it reads as a delta rather than a re-post. One of them needed more than a field change, and one of your points turned out to be stronger than stated.

**`getpeerinfo.startingheight` → `None`.** Stronger than "deprecated and omitted by default": the name does not appear *anywhere* in Core 31's source. v31 does not emit it at all, and corepc keeps the field `Option` only for older versions.

**`limitclustercount` / `limitclustersize` — "prefer `None`" was not available here.** Both are `i64`, not `Option<i64>`, so the typed model forces a value. That left a choice between a number that means nothing and a policy this node actually enforces, and #201 implements the policy: `cluster_count` 64 and `cluster_size_vbytes` 101,000 (Core's `DEFAULT_CLUSTER_LIMIT` / `DEFAULT_CLUSTER_SIZE_LIMIT_KVB`), enforced at admission, with `getmempoolinfo` reading them from the pool.

Your "different policy semantics" is understated, and worth recording. Core 31's `init.cpp` calls `-limitancestorcount` / `-limitdescendantcount` *"Deprecated setting … replaced by cluster limits (see -limitclustercount) and only used by wallet for coin selection"* — Core has **retired the ancestor pair as mempool policy**. And `max_ancestor_size` happens to be `101_000`, the same number as Core's cluster size limit, which is presumably why the mapping looked plausible: a coincidence of value, not of meaning.

A cluster is also not an ancestor package in shape. It is a connected component, so two children of one parent share one although neither is an ancestor of the other — `metadata_closure` cannot express it, and #201 adds a real undirected walk.

## One thing to flag on this branch specifically

`corepc_types::v31::GetMempoolInfo::full_rbf` is a non-optional `bool`, while Core 31 emits `fullrbf` only under `-deprecatedrpc=fullrbf` (`rpc/mempool.cpp`). Serializing the struct emits it unconditionally, so this node advertises a field default Core omits.

**The compatibility test cannot catch that**, and neither can any test built the same way: it deserializes the response back into the same struct, so a field the struct declares is exactly what it expects. Same blind spot covers the placeholders — `bytes_sent: 0` deserializes exactly as a real count does, and there are 124 such assignments across the migrated handlers today.

That is not an argument against the direction; the typed contract is the right call and catches the class of drift the hand-maintained key lists could not. It just means the shape check and the value check are now different problems, and only one of them is solved.

## Two smaller ones from the same pass

`ping_time` / `minimum_ping` become `None` rather than `Some(0.0)` — nothing pings peers, and a zero asserts a round trip that took no time.

`presynced_headers` / `synced_headers` / `synced_blocks` are **left** at `Some(-1)`, which I had initially marked as placeholders to fix. Core's own declaration reads *"the current height of header pre-synchronization with this peer, or -1 if no low-work sync is in progress"* — -1 is Core's way of saying unknown, so `None` would have been the wrong change.

## Sequencing

This branch zero-fills fields that open PRs already compute, so rebasing them in this order resolves each conflict toward real data:

1. **#107** — `PeerCounters` and `time_offset`, exactly the `getpeerinfo` fields left at `0`.
2. **#109** — `getmininginfo`, removing `STUB_BITS_HEX` / `STUB_TARGET_HEX`.
3. #98, #100, #106, #108 — the rest of `chain.rs` / `mempool.rs` / `tx.rs`.
4. **#177** last: it overlaps all twelve files here and builds responses by hand with no corepc, so rebasing it onto the typed contract keeps one owner for response shape.

**#160 can close with this PR** — `getblockchaininfo` now carries real `bits` / `target` from the tip header.

gosunuts:
Current-main review: the typed `corepc-types::v31` boundary still looks like the clearest direction for RPC response ownership, but this branch now needs a fresh rebase against the mining/RPC changes that landed after it was cut.

I would keep this as an active PR, not mark it superseded. During rebase, prefer preserving current `main` value semantics and moving only response-shape ownership to corepc types. Also resolve #126/#150 around this direction so the repository has one response contract and, separately, any independent compatibility oracle we still want.

metaphorics (commented):
Gate-and-merge: superseded — recommend close, head `0888e08`.

Post-resolution this branch is byte-identical to `main` (`git diff origin/main...HEAD` empty): the parent's landing path took main's side everywhere, and the only residue was dependency-ordering churn in `crates/rpc/Cargo.toml`, which I zeroed (`0888e08`, restoring main's bytes) rather than ship as a content-free merge. There is nothing left here to land. QA: no diff, no findings, no test debt.

Recommendation: close this PR unmerged. No merge will be performed by the queue.

PR #200, opened 2026-08-27
perf(p2p): coalesce message writes into one writev

gg582:
## Summary

`wire::write_message` previously emitted one p2p message with five sequential `write_all` calls (magic, command, length, checksum, payload). It now assembles the 24-byte header on the stack and emits header + payload with a single `write_vectored` loop (`IoSlice`, no payload copy, partial-write safe). Wire format is unchanged.

Splitting one message into five small segments also triggered the Nagle/delayed-ACK deadlock on every request/response exchange (ping/pong, getdata/tx, getheaders/headers). A single segment never satisfies Nagle's unacked-small-segment condition, so the stall disappears **without touching any socket options** — this PR deliberately does not set `TCP_NODELAY` (measured to add nothing on top of write coalescing; see gosuda/bitcoin-rs#190 discussion).

## Benchmarks (fresh, this machine, criterion, loopback TcpStream)

| payload | main (5x write_all) | this PR (writev) | delta |
|---|---|---|---|
| ping 8B | 24.40 µs | 4.95 µs | **-79.7%** |
| block 285B | 23.75 µs | 5.75 µs | **-75.8%** |

Round-trip latency (ping/pong, responder + client same write path, from the #190 measurement harness):

| link | main | writev |
|---|---|---|
| loopback | ~92 ms (Nagle + delayed-ACK deadlock, ~40 ms per direction) | ~18 µs |
| emulated WAN, 20 ms one-way | ~132 ms (5 segments serialized ≈ 3+ RTT) | ~40 ms = exactly 1 RTT |

Reproduce:

```bash
cargo bench -p bitcoin-rs-p2p --bench write_message
```


gosunuts:
Current-main review: this still looks like an active, independent performance PR rather than an outdated one. The write coalescing is localized to the P2P wire path, keeps the wire format unchanged, and addresses a concrete latency problem without broadening socket-policy surface.

I would keep this in the normal review queue. Given the recent P2P lifecycle fixes, it is still worth rebasing/rerunning the benchmark and integration tests on current `main`, but I do not see a newer architectural change that supersedes the idea.

gg582:
Yes, it is a minor performance enhancement without changing a structure. I'll make sure with additional benchmarks next week.

metaphorics (commented):
Gate-and-merge hold: Gate 2, head `27dd0ce`.
QA is clean: full-diff review found the vectored loop handles partial writes and iovec limits (`IoSlice::advance_slices`), `Interrupted` retries, `Ok(0)` maps to `WriteZero`, no `unsafe`, bench additive-only. One concern.

PR #177, opened 2026-08-26
Compact Core interfaces and full-stack mining

metaphorics:
Implements compact Core-compatible RPC/REST/Esplora and node-owned
solo-mining lifecycle per the compact-core-mining DAG.

- Consensus: single contextual source for versionbits, work, finality,
  dry-run/full validation; bip9_cache remains internal.
- Mempool: immutable mining snapshot, package topology, real sigops,
  persistent fee overlay, estimator ownership.
- Mining: real candidate assembly, package ordering, coinbase, witness
  commitment, limits, and assembler state through MiningControl.
- Node: generation-keyed cache, single-flight rebuild, broadcast
  wake, and synchronous submission.
- RPC: Core projections for chain, tx, mempool, network, mining, and
  REST; 55 methods with ZMQ, 54 without, and twelve REST prefixes.

## Merged with main

`main` removed three subsystems after this branch was cut, each with a
stated no-in-tree-consumer rationale, so those deletions are carried
in and the work layered on them is dropped:

- Compact-filter indexing (#146): `getblockfilter`, both
  `/rest/blockfilter*` prefixes, the `basicblockfilterindex`
  `getindexinfo` entry, and `NODE_COMPACT_FILTERS` from the advertised
  service flags, so `localservices` is now `0x09`.
- The Utreexo node mode (#148).
- The in-tree wallet implementation (#165). `main` kept the node-side
  descriptor, UTXO-scan, and PSBT-assembly RPCs by reimplementing them
  without the crate, so dispatch routes `getdescriptorinfo` and
  `deriveaddresses` to `util`, `combinepsbt` and `finalizepsbt` to
  `tx`, and `scantxoutset` to `chain`. Wallet-only methods are absent
  rather than registered-and-disabled, which is why the method counts
  above dropped from 65/64.

Kept from `main` on top of this branch's structure: the
chain-transition mutex, now initialized in `Context::production` and
shared from the node so `scantxoutset` still reads a stable
chainstate; the coinstats relocation into `bitcoin-rs-utxo`; and the
frozen checkpoint codec test, pinned to this branch's deliberate
`bitcoin-rs-canonical-headers-v2` bump.

## verificationprogress

The tree's `chain_tx_count` was restored while the `Context` atomic
stayed zero, so `getblockchaininfo` fell back to the height ratio
(0.5) instead of Core's estimate. This branch makes the block-tree
nodes authoritative and removed the atomic that block application used
to maintain, so rather than reseed a second source of truth,
[...]

Issue #114, opened 2026-08-24
Define and differentially verify Bitcoin Core-compatible mempool policy


gosunuts:
## Goal

Define and verify the externally observable mempool and transaction-admission behavior that bitcoin-rs claims to support, using a pinned Bitcoin Core reference while keeping the internal mempool implementation independent.

This issue is about policy compatibility and wallet/miner-facing behavior. It is separate from the ZMQ `A`/`R` event work in #74.

## Scope

- transaction admission and rejection categories
- standardness policy
- minimum relay fee and fee-related rejection behavior
- ancestor/descendant limits
- RBF/replacement behavior
- eviction and removal behavior where externally observable
- reorg transaction re-admission/reconsideration
- `sendrawtransaction` / relay interaction
- fee-estimation inputs only where compatibility claims depend on them

Use deterministic mempool fixtures and differential tests against the pinned Core reference. Exact internal data structures and eviction algorithms do not need to match when observable behavior remains compatible.

## Acceptance

- [ ] A pinned Bitcoin Core reference version is declared for mempool-policy compatibility.
- [ ] Supported policy behaviors and intentional deviations are documented.
- [ ] Representative valid, non-standard, conflicting, fee-insufficient, ancestor/descendant, and replacement cases are differentially tested.
- [ ] Reorg removal and re-admission behavior is tested.
- [ ] Transactions accepted through RPC relay correctly over P2P when relay is enabled.
- [ ] Wallet-generated standard transactions can be submitted without relying on bitcoin-rs-internal wallet privileges.
- [ ] Mining transaction selection consumes the same committed mempool state without defining a second policy universe.

Related: #74, #78, #112, #113.

PR #201, opened 2026-08-27
feat(mempool): enforce Core's cluster limits, and report the ones enforced

rabbitson87:
Stacked on #176 — it targets `feat/corepc-rpc-wire-contract`, not `main`, so it can be reviewed as a delta rather than as a re-post of that branch. (Stacked PRs get CI now; #168 fixed the trigger that had been excluding them.)

Addresses the review on #176 and the mempool policy the review's second finding turned out to need.

## `None` was available for one field and not the other

The review asked for `None` where bitcoin-rs has no equivalent concept. That works for `starting_height`:

```rust
pub starting_height: Option<i64>,
```

It does not work for the cluster fields:

```rust
pub limit_cluster_count: i64,   // not Option
pub limit_cluster_size: i64,    // not Option
```

So the choice there was between a number that means nothing and a policy this node actually enforces. **This implements the policy.**

That gap is the general cost of a typed v31 model, and it is now written down rather than absorbed silently — see `DEVIATIONS.md` §10.

## Cluster limits

Core 31 replaced ancestor and descendant limits with cluster limits. Its `init.cpp` calls `-limitancestorcount` and `-limitdescendantcount` *"Deprecated setting … replaced by cluster limits (see -limitclustercount) and only used by wallet for coin selection"* — so the reviewer's "different policy semantics" is if anything understated: Core has retired the ancestor pair as mempool policy, and bitcoin-rs still enforces it.

`MempoolLimits` gains `cluster_count` (64, `DEFAULT_CLUSTER_LIMIT`) and `cluster_size_vbytes` (101,000, `DEFAULT_CLUSTER_SIZE_LIMIT_KVB` × 1000), enforced in `insert_entry` beside the existing checks. `getmempoolinfo` reads them from the pool.

**`metadata_closure` is deliberately not reused.** It is ancestors ∪ self ∪ descendants — a chain through one transaction — so it misses siblings and under-counts every fan-out. A cluster is a connected component. `cluster_ids_seeded_by` is a real undirected walk whose frontier alternates direction, because reaching a sibling takes an up-edge then a down-edge.

Seeds are taken in **both** directions, because the candidate is not in the pool yet: its in-pool parents, *and* any in-pool transaction already spending its outputs. The second half is not hypothetical — `insert_entry` documents that a transaction can arrive after something that spends it, through orphan promotion or out-of-order relay.

The ancestor and descendant limits stay. Core deprecated them, this node did not, and retiring them is a policy decision for #114 rather than for a wire-format change.

## The mutation audit found a hole before this was committed

Two killed, one invalid, and one that mattered:

| mutation | result |
| --- | --- |
| `existing_spenders_of` returns empty | **KILLED** by `a_cluster_includes_transactions_that_already_spend_the_candidate` |
| `cluster_ids_seeded_by` → `metadata_closure` at the admission site | **SURVIVED** at first |
| delete the downstream seed line | **INVALID** — makes `existing_spenders_of` dead code, and rustc 1.95.0 then ICEs in `check_mod_deathness` |

The survivor is the useful one. Every admission fixture I had written hung directly off a single root, and there the chain through the root *happens* to be the whole component — so swapping the walks changed nothing and the new walk was not load-bearing in any of them. `admission_counts_a_cousin_that_no_chain_through_the_parent_reaches` builds `A→B→D` beside `A→C`, so the candidate's cousin is unreachable through its parent. The mutation dies on that test alone.

Without the audit this would have shipped claiming a property its tests did not check.

## Core corrected one of my own judgements

`presynced_headers` / `synced_headers` / `synced_blocks` are `Some(-1)` and I had them down as placeholders to fix. Core's own declaration reads *"the current height of header pre-synchronization with this peer, or **-1** if no low-work sync is in progress"* — **-1 is Core's way of saying unknown**, so `Some(-1)` is accurate and `None` would have been the wrong change. Left alone.

`ping_time` / `minimum_ping` did become `None`: nothing pings peers, and `Some(0.0)` asserts a round trip that took no time.

## One fixture the new limit caught

`bip125_replacement_rules_are_enforced` raises the ancestor caps to build a 200-transaction chain. That chain is one cluster of 200, so it is refused at 64 before the replacement rules are reached — correct under Core 31 semantics. The fixture now lifts the cluster caps alongside the ancestor caps; it is a test about BIP125, not about cluster limits.

## Verified

```
cargo test -p bitcoin-rs-rpc                                            298 passed
cargo test -p bitcoin-rs-mempool                                        all passed
```

`cargo clippy -p bitcoin-rs-mempool --all-targets -- -D warnings` reports **31 errors before and after** this change — all pre-existing, all in test and bench targets that no CI lane lints. Same blind spot as `-p bitcoin-rs-rpc`, which fails on clean `main` with 28. Not touched here.

## After this: the queue

#176 currently zero-fills fields that open PRs already compute. Rebasing them in this order resolves each conflict toward real data rather than away from it:

1. **#107** — `PeerCounters` and `time_offset`: exactly the `getpeerinfo` fields left at `0`.
2. **#109** — `getmininginfo`, removing `STUB_BITS_HEX` / `STUB_TARGET_HEX`.
3. #98, #100, #106, #108 — the remaining `chain.rs` / `mempool.rs` / `tx.rs` fixes.
4. **#177** last. It overlaps all twelve of #176's files and builds responses by hand with no corepc; rebasing it onto the typed contract keeps one owner for response shape.

**#160** can close with #176: `getblockchaininfo` now carries real `bits` / `target` from the tip header.

Still open and not decided here: whether #126 / #150 are retired now that corepc is the contract, or kept as the value-level check corepc cannot provide. #150 extracts from Core's own `RPCResult` declarations, which is a stronger source than corepc's hand-written mirrors, but both are stale and conflicting.

## Summary by Sourcery

Enforce Core v31 cluster limits across mempool admission and RPC acceptance paths, and expose the configured limits through `getmempoolinfo`.

New Features:
- Enforce Bitcoin Core 31 cluster count and virtual-size limits during mempool admission and transaction acceptance preview.
- Report the enforced cluster limits through `getmempoolinfo`.
- Handle absent `startingheight` values according to the Core v31 RPC contract.

Bug Fixes:
- Ensure replacement transactions are evaluated against clusters after evicted entries are removed.
- Ensure cluster calculations include siblings, cousins, and transactions already spending the candidate's outputs.
- Keep mempool admission and RPC acceptance preview verdicts consistent for cluster-limit checks.

Enhancements:
- Retain ancestor and descendant policy checks while distinguishing them from Core's cluster policy.
- Update the mempool policy contracts and RPC compatibility documentation for cluster enforcement and reporting.

Documentation:
- Document cluster count and size limits, their enforcement behavior, and their interaction with replacement and acceptance preview.

Tests:
- Add unit, mempool integration, and RPC tests covering cluster connectivity, count and size limits, replacement behavior, and preview/admission agreement.


gosunuts:
Current-main review: this is not stale in the same sense as the old mining stack; it is intentionally stacked on #176 and should remain blocked on that contract decision. The cluster-limit implementation itself is still a plausible current policy change, but its RPC projection depends on the corepc response model introduced below it.

So I would keep this open but treat it as waiting on #176. Once #176 is rebased/finalized, rebase this PR onto it and re-evaluate the policy semantics against current mempool behavior rather than merging it independently.

rabbitson87:
Agreed, and the state has not changed in the direction that would unblock it: #176 has not been rebased since 2026-08-29 (head still `17b56a9`), and `main` has moved a long way underneath it — the wallet extraction (#165), the ccm mining chain, the benchmark retirement, and the test-surface cleanup have all landed since.

So this stays open and waiting, as you asked, with one thing worth flagging for when #176 *is* finalized: the cluster-limit work here is now more than an RPC projection. It changed mempool admission policy — `insert_entry` enforces `cluster_count` / `cluster_size_vbytes` beside the existing ancestor and descendant checks — and a fixture had to be adjusted because a 200-transaction chain is one cluster of 200 and is correctly refused at Core's default of 64.

That half does not depend on #176 at all. If the corepc contract decision runs long, the policy change is separable: the mempool enforcement plus its tests could land against `main` on its own, leaving only the `getmempoolinfo` field mapping — which is genuinely downstream of the response model — to follow #176. Happy either way; splitting only makes sense if #176 is expected to sit for a while, and that is your call rather than mine.

When #176 lands I will rebase this onto it and re-check the policy semantics against current mempool behavior rather than assuming the August measurements still hold.

metaphorics (commented):
Gate-and-merge hold: blocking Gate 3 + Gate 4 findings, head `55040ef` (conflict-resolved onto parent, MERGEABLE; resolution verified: native-API port, DEVIATIONS.md deletion kept).

Blocking, Gate 3 (mixed concerns): the diff still bundles the `getpeerinfo`-honesty hunks (`crates/rpc/src/handlers/network.rs` — hard-coded `last_send`/`last_received`/`last_transaction` zeros plus a justifying comment) under a cluster-limits title. That is a second concern with its own revert story (it papers over unmeasured values the reviewer already flagged as indistinguishable-from-real zeros); split it into its own PR.

Blocking, Gate 4 (two findings, both verified against the branch tree):

1. Class 1, `crates/mempool/src/pool.rs:1391` (`check_cluster_limits(&self, entry)` — takes no exclusion set): a BIP125 replacement validates with its conflicts still resident, so replacing one member of a cluster already at `cluster_count` (or near `cluster_size_vbytes`) counts both the soon-evicted original and its replacement, rejecting a replacement whose post-eviction cluster is within limits. Reachable on any replacement into a full cluster.
2. Class 5 (contract drift between two RPCs), same area: actual admission runs the new cluster check via `validate_insert`, but the `testmempoolaccept` preview path (`evaluate_one` → `check_package_limits`, pool.rs:1579) checks only ancestor/descendant limits. A candidate violating only a cluster limit previews as allowed and then fails admission — the two RPCs disagree on the same candidate.

Test debt: no replacement test holds the post-eviction cluster at either limit; no test proves preview and admission agree on a cluster-only rejection. Both fixes are author design (exclusion-set plumbing, preview parity), so holding rather than redesigning here.

Gate 2: `fmt`, `clippy`, `test`, `deny` fail on this head — the same set `main` itself produces (proven by the byte-identical-to-main PR 176 run). Diagnose outside the queue. No merge on this oid until then.

metaphorics (commented):
## Review: two blocking findings + one author action

### Finding 1 — `check_cluster_limits` ignores the eviction set (blocking)

`check_cluster_limits(&self, entry)` at pool.rs:1391 takes no `excluded` parameter. During a replacement, `validate_insert` passes `excluded` to both `check_descendant_limits_excluding` (pool.rs:1315) and the evicted-parent guard (pool.rs:347), but the cluster check counts every member of the connected component — including the very entries the replacement is about to evict. A replacement that shrinks an over-large cluster is falsely rejected by the cluster it is clearing.

This is the same class of bug the descendant guard already handles: `check_descendant_limits_excluding` filters `excluded` entries (pool.rs:1323) so a replacement that trims an over-large descendant package is not rejected by the package it is about to clear. The cluster check needs the same treatment.

**Fix:** thread `excluded` through `check_cluster_limits`, filter evicted entries from the cluster walk, and generalize the signature from `&MempoolEntry` to `(&Tx, Txid, u32, &HashSet<EntryId>)` so the preview path can call it too.

### Finding 2 — preview path lacks the cluster check (blocking)

`check_package_limits` at pool.rs:1579 calls `check_ancestor_count_and_size` and `check_descendant_limits_excluding` but not `check_cluster_limits`. The preview path (`evaluate_one` → `check_package_limits`, standardness.rs:373) is the `testmempoolaccept` dry-run; the admission path (`validate_insert`) runs all three checks. A transaction that passes ancestor/descendant limits but exceeds a cluster limit is reported `allowed` by the preview and then rejected by admission — the two paths diverge.

The doc on `check_package_limits` (pool.rs:1564) already states the contract: *"the acceptance preview and the admission gate must quote the same verdict."* The cluster check is missing from that mirror.

**Fix:** add `self.check_cluster_limits(tx, tx.txid(), vsize, excluded)?;` to `check_package_limits`, passing the same `excluded` set the descendant check already receives.

### Patch

Both fixes plus two behavior tests (both should fail pre-fix):

```diff
diff --git i/crates/mempool/src/pool.rs w/crates/mempool/src/pool.rs
index 5f86f5f..df49b06 100644
--- i/crates/mempool/src/pool.rs
+++ w/crates/mempool/src/pool.rs
@@ -346,21 +346,21 @@ impl Mempool {
             self.by_txid
                 .get(&input.previous_output.txid)
                 .is_some_and(|id| excluded.contains(id))
         }) {
             return Err(MempoolError::EvictedParent);
         }
 
         let ancestors = self.ancestor_ids_for_tx(&entry.tx);
         self.check_ancestor_limits(&ancestors, &entry)?;
         self.check_descendant_limits_excluding(&ancestors, excluded)?;
-        self.check_cluster_limits(&entry)?;
+        self.check_cluster_limits(&entry.tx, entry.txid, entry.vsize, excluded)?;
 
         if excluded.is_empty() && u32::try_from(self.entries.vacant_key()).is_err() {
             return Err(MempoolError::TooManyEntries);
         }
 
         let ancestor_size = ancestors.iter().fold(u64::from(entry.vsize), |total, id| {
             total.saturating_add(
                 self.entry(*id)
                     .map_or(0, |ancestor| u64::from(ancestor.vsize)),
             )
@@ -1381,38 +1381,56 @@ impl Mempool {
 
     /// Refuses a transaction that would join a cluster over either limit.
     ///
     /// The candidate is not in the pool yet, so its cluster is seeded from
     /// both directions: the in-pool parents it spends, and any in-pool
     /// transaction that already spends *its* outputs. The second half is not
     /// hypothetical -- `insert_entry` documents that a transaction can arrive
     /// after something spending it, through orphan promotion or out-of-order
     /// relay -- and a seed set of parents alone would miss the descendants
     /// this transaction is about to connect.
-    fn check_cluster_limits(&self, entry: &MempoolEntry) -> Result<(), PolicyError> {
-        let mut seeds = entry
-            .tx
+    ///
+    /// `excluded` is the set of entry ids a replacement will evict; those
+    /// entries are skipped so a replacement that shrinks an over-large cluster
+    /// is not falsely rejected by the cluster it is about to clear -- the same
+    /// exclusion `check_descendant_limits_excluding` applies.
+    fn check_cluster_limits(
+        &self,
+        tx: &Tx,
+        txid: Txid,
+        vsize: u32,
+        excluded: &HashSet<EntryId>,
+    ) -> Result<(), PolicyError> {
+        let mut seeds = tx
             .inputs
             .iter()
             .filter_map(|input| self.by_txid.get(&input.previous_output.txid).copied())
             .collect::<Vec<_>>();
-        seeds.extend(self.existing_spenders_of(entry.txid, entry.tx.outputs.len()));
+        seeds.extend(self.existing_spenders_of(txid, tx.outputs.len()));
         if seeds.is_empty() {
             // A cluster of one. Still checked, so a single oversized
             // transaction cannot pass a limit its cluster would fail.
-            return cluster_within_limits(1, u64::from(entry.vsize), &self.limits);
+            return cluster_within_limits(1, u64::from(vsize), &self.limits);
         }
         let cluster = self.cluster_ids_seeded_by(&seeds);
-        let count = u32::try_from(cluster.len())
+        // Drop entries a replacement will evict, mirroring
+        // `check_descendant_limits_excluding`: a replacement that shrinks an
+        // over-large cluster must not be falsely rejected by the cluster it
+        // is about to clear.
+        let live: Vec<EntryId> = cluster
+            .into_iter()
+            .filter(|id| !excluded.contains(id))
+            .collect();
+        let count = u32::try_from(live.len())
             .unwrap_or(u32::MAX)
             .saturating_add(1);
-        let vsize = cluster.iter().fold(u64::from(entry.vsize), |total, id| {
+        let vsize = live.iter().fold(u64::from(vsize), |total, id| {
             total.saturating_add(self.entry(*id).map_or(0, |member| u64::from(member.vsize)))
         });
         cluster_within_limits(count, vsize, &self.limits)
     }
 
     /// In-pool entries already spending an output of `txid`.
     ///
     /// Keyed by txid rather than [`EntryId`] because the caller's transaction
     /// has no entry id yet.
     fn existing_spenders_of(&self, txid: Txid, output_count: usize) -> Vec<EntryId> {
@@ -1578,20 +1596,21 @@ impl Mempool {
     /// transaction the admission gate then rejects on package limits.
     pub fn check_package_limits(
         &self,
         tx: &Tx,
         vsize: u32,
         excluded: &HashSet<EntryId>,
     ) -> Result<(), PolicyError> {
         let ancestors = self.ancestor_ids_for_tx(tx);
         self.check_ancestor_count_and_size(&ancestors, vsize)?;
         self.check_descendant_limits_excluding(&ancestors, excluded)?;
+        self.check_cluster_limits(tx, tx.txid(), vsize, excluded)?;
         Ok(())
     }
 
     fn entry_mut(&mut self, id: EntryId) -> Option<&mut MempoolEntry> {
         usize::try_from(id)
             .ok()
             .and_then(|index| self.entries.get_mut(index))
     }
 
     fn entry_signals_rbf(&self, id: EntryId) -> bool {
@@ -3802,11 +3821,138 @@ mod spend_index_tests {
         let (mut pool, root_txid) = graph_pool();
         let Some(root_id) = pool.entry_id_by_txid(&root_txid) else {
             panic!("root missing from the fixture pool");
         };
         let removed = pool.remove_entry_and_descendants(root_id);
         assert!(!removed.is_empty());
         // The root left with its descendants, so nothing is left to spend it.
         assert!(pool.entry_id_by_txid(&root_txid).is_none());
         assert!(pool.is_empty(), "the fixture is a single connected package");
     }
+
+    /// A replacement that evicts members of a cluster at the count limit
+    /// must not be falsely rejected: the evicted entries do not count toward
+    /// the cluster after replacement, so the replacement's cluster is smaller
+    /// than the pre-replacement count suggests.
+    ///
+    /// Pre-fix this fails because `check_cluster_limits` takes no exclusion
+    /// set and counts the very entries the replacement is about to evict.
+    #[test]
+    fn a_replacement_that_evicts_from_a_full_cluster_is_admitted() {
+        // root has three outputs. A spends root:0 and signals RBF
+        // (sequence 0xFFFF_FFFD). B spends root:1. The cluster
+        // {root, A, B} is at the limit of three.
+        let confirmed = OutPoint::new(txid_of([21_u8; 32]), 0);
+        let root = tx_with(&[confirmed], 3, 1);
+        let root_txid = root.txid();
+        let mut a = tx_with(&[OutPoint::new(root_txid, 0)], 1, 2);
+        a.inputs[0].sequence = 0xFFFF_FFFD;
+        let b = tx_with(&[OutPoint::new(root_txid, 1)], 1, 3);
+
+        let limits = MempoolLimits {
+            cluster_count: 3,
+            ..MempoolLimits::default()
+        };
+        let mut pool = Mempool::new(limits);
+        let Ok(_id) = pool.insert_entry(MempoolEntry::new(Arc::new(root), 100, 10_000, 1, 7))
+        else {
+            panic!("root must be admitted");
+        };
+        let Ok(_id) = pool.insert_entry(MempoolEntry::new(Arc::new(a), 100, 10_000, 2, 7))
+        else {
+            panic!("A must be admitted");
+        };
+        let Ok(_id) = pool.insert_entry(MempoolEntry::new(Arc::new(b), 100, 10_000, 3, 7))
+        else {
+            panic!("B must be admitted");
+        };
+
+        // Replacement spends root:0 (conflicts with A, which signals RBF)
+        // and root:2 (keeps root as an in-pool parent). Fee 12_000 beats
+        // A's 10_000 on both absolute and rate, satisfying BIP125 rules 3-6.
+        let replacement = tx_with(
+            &[OutPoint::new(root_txid, 0), OutPoint::new(root_txid, 2)],
+            1,
+            4,
+        );
+        let candidate = crate::ReplacementCandidate::new(
+            Arc::new(replacement),
+            100,
+            12_000,
+            1,
+        );
+        let result = pool.replace_transaction(candidate, 4, 7, 4);
+        assert!(
+            result.is_ok(),
+            "a replacement that evicts from a full cluster must be admitted, \
+             not rejected by the entries it removes: {result:?}"
+        );
+    }
+
+    /// The preview path (`check_package_limits`) must reject a transaction
+    /// that would exceed the cluster count limit, matching the admission
+    /// gate (`validate_insert`). Without the cluster check in
+    /// `check_package_limits`, `testmempoolaccept` reports `allowed` for a
+    /// transaction that `sendrawtransaction` then rejects.
+    ///
+    /// Pre-fix this fails because `check_package_limits` does not call
+    /// `check_cluster_limits`.
+    #[test]
+    fn check_package_limits_rejects_a_cluster_only_violation() {
+        // root -> child_a, child_b: a 3-member cluster at the limit of three.
+        // child_c spending root would be the fourth member. Ancestor and
+        // descendant limits are lifted so only the cluster check can refuse.
+        let confirmed = OutPoint::new(txid_of([23_u8; 32]), 0);
+        let root = tx_with(&[confirmed], 2, 1);
+        let root_txid = root.txid();
+        let child_a = tx_with(&[OutPoint::new(root_txid, 0)], 1, 2);
+        let child_b = tx_with(&[OutPoint::new(root_txid, 1)], 1, 3);
+
+        let limits = MempoolLimits {
+            cluster_count: 3,
+            max_ancestors: 100,
+            max_ancestor_size: 1_000_000,
+            max_descendants: 100,
+            ..MempoolLimits::default()
+        };
+        let mut pool = Mempool::new(limits);
+        for tx in [root, child_a, child_b] {
+            let Ok(_id) = pool.insert_entry(MempoolEntry::new(Arc::new(tx), 100, 10_000, 1, 7))
+            else {
+                panic!("the three-member cluster must be admitted under a limit of three");
+            };
+        }
+
+        // child_c would join the cluster as the fourth member.
+        let child_c = tx_with(&[OutPoint::new(root_txid, 0)], 1, 4);
+        let excluded: HashSet<EntryId> = HashSet::new();
+
+        // Preview must reject on cluster count.
+        let preview = pool.check_package_limits(&child_c, 100, &excluded);
+        assert!(
+            matches!(preview, Err(PolicyError::ClusterCountLimit)),
+            "the preview must reject a cluster-only violation: {preview:?}"
+        );
+
+        // Admission must reject on the same error.
+        let admission = pool.insert_entry(MempoolEntry::new(
+            Arc::new(child_c),
+            100,
+            10_000,
+            4,
+            7,
+        ));
+        assert!(
+            matches!(
+                admission,
+                Err(MempoolError::Policy(PolicyError::ClusterCountLimit))
+            ),
+            "admission must reject on cluster count: {admission:?}"
+        );
+
+        // The two verdicts must agree.
+        assert_eq!(
+            preview.map_err(MempoolError::from).err(),
+            admission.err(),
+            "preview and admission must reject on the same error"
+        );
+    }
 }
```

`git apply --check` passes against the PR head (55040ef). Both tests pass with the fix applied; all 8 existing cluster tests and 5 BIP125 tests remain green.

### Author action (not patchable): split the network.rs getpeerinfo-honesty hunks

The `crates/rpc/src/handlers/network.rs` changes (ping_time/minimum_ping `Some(0.0)` → `None`, `starting_height` `Some(...)` → `None`, and the honesty comments on unmeasured counters) are unrelated to cluster limits. They should be split into their own PR so this one stays focused on the mempool policy change. These hunks are at network.rs:176-219 in the diff.

metaphorics:
@greptileai

metaphorics:
@sourcery-ai review

metaphorics:

metaphorics:

metaphorics:

metaphorics:
@cursor Resolve, validate, and fix the pr comments and feedbacks + fix merge conflicts in the pr

Issue #215, opened 2026-08-29
Production P2P handshake publication cancels the current peer lease


gosunuts:
## Summary

Production P2P sessions cancel their own `PeerLease` during handshake publication. In the live checkout, the listener inserts the current lease into `peer_outbound` before the handshake, then the production registration callback removes and cancels the lease at the same address. The first post-handshake loop therefore exits before reading any message.

This explains the observed combination of successful version/verack exchange, no pong to an inbound local ping, repeated handshake-complete logs, `getpeerinfo=[]`, and header sync remaining at the restored 778000 header tip.

## Reproduction

Live checkout: `/home/gosunuts/workspace/bitcoin-rs`.

1. Start the node with the production listener and `BlockSync::peer_registration_handle` wiring.
2. Complete an inbound Bitcoin P2P version/verack handshake against port 8333.
3. Send a valid `ping` immediately after verack.
4. Observe no `pong`; the session closes before the first post-handshake read.

The production path is wired in `crates/node/src/run.rs:163-182`, `crates/node/src/run.rs:283-297`, and `crates/node/src/run.rs:622-623`. A full clean reproduction against a rebuilt binary was not run because the live datadir is the affected mainnet instance; the source-level path is deterministic.

## Evidence

- `crates/p2p/src/listener.rs:535-539` and `crates/p2p/src/listener.rs:661-665` register the current outbound/inbound lease before handshake completion.
- `crates/p2p/src/listener.rs:772-778` publishes handshake metadata through the production registration callback.
- `crates/node/src/sync.rs:321-329` removes the lease at the peer address, calls `prior.cancel()`, then inserts the newly supplied lease. With pre-handshake registration, `prior` is the current connection's lease, not a predecessor.
- `crates/p2p/src/connection.rs:195-215` defines the shared lease cancellation state and `is_cancelled`; cancellation is shared by all lease clones.
- `crates/p2p/src/listener.rs:842-856` checks cancellation before calling `read_message`. A self-cancelled session returns cleanly at the cancellation check and never reads the first post-handshake message.
- `crates/p2p/src/dispatch.rs:60-63` would produce a `Pong` for a valid `Ping` once the ready message loop receives it.
- `crates/p2p/src/listener.rs:801-821` removes the current peer, shuts down the socket, and records a clean disconnect after that loop exit.
[...]

PR #218, opened 2026-08-29
refactor(p2p): centralize peer lifecycle ownership

gosunuts:
## Summary

- centralize lease registration, readiness publication, replacement cancellation, and teardown in P2P PeerLifecycle
- reduce BlockSync to peer snapshots, message delivery, and identity-checked disconnect requests
- clear address-scoped sync state before a replacement becomes ready and keep handshaking replacements out of eviction decisions
- document the ownership boundary and same-address identity guardrails

## Context

This is the focused architectural follow-up to #215 and merged fix #216. The transport layer now owns peer lifecycle mutation instead of delegating it back into node sync.

## Verification

- cargo fmt --all -- --check
- cargo test -p bitcoin-rs-p2p --no-default-features --lib (52 passed)
- cargo test -p bitcoin-rs-node --no-default-features --features fjall --lib (509 passed)
- cargo clippy -p bitcoin-rs-p2p --no-default-features --lib --tests -- -D warnings
- cargo clippy -p bitcoin-rs-node --no-default-features --features fjall --lib --tests -- -D warnings

## Summary by Sourcery

Centralize P2P peer lifecycle ownership in P2pService and enforce connection-identity checks across synchronization and teardown operations.

Bug Fixes:
- Prevent stale peer lifecycle operations from publishing metadata, sending messages, or disconnecting same-address replacements.
- Clear address-scoped synchronization state before replacement peers become ready and exclude handshaking peers from ready-peer eviction decisions.

Enhancements:
- Centralize P2P peer lifecycle, connection state, worker management, and download scheduling under P2pService and PeerLifecycle.
- Reduce node synchronization responsibilities to peer snapshots, inbound event coordination, and identity-checked peer operations.
- Carry connection identity through ready-peer snapshots and lifecycle mutations instead of relying on address equality.
- Consolidate listener, outbound dialing, DNS or fixed-peer bootstrap, and teardown ownership in the P2P service.

Documentation:
- Document the P2P lifecycle ownership boundary and same-address connection identity guardrails.

Tests:
- Add coverage for stale publication, replacement-preserving disconnects, identity-bearing snapshots, and stale scheduler mutations.


gosunuts (commented):
The overall direction looks good: moving peer lifecycle mutation out of `BlockSync` and into P2P is the right architectural follow-up to #215/#216.

One race still looks worth fixing before merge, though. `publish_peer_info()` checks that `lease` is current, then invokes `peer_ready(peer_addr)`, and only afterward calls `publish_ready()`, which re-checks identity. The final metadata publication is therefore protected, but the readiness callback is address-only and not identity-checked.

A possible interleaving is:

1. connection A finishes handshake and passes `is_current(A)`
2. same-address connection B registers and replaces A
3. B runs `peer_ready(addr)`, publishes ready, and may receive sync work
4. A resumes and runs `peer_ready(addr)`
5. A's `publish_ready()` correctly fails because A is stale

Step 4 can still clear B's address-scoped `DownloadWindow` / pending `getheaders` state. This is the same-address identity class of race that this refactor is intended to eliminate.

I think the readiness transition should carry connection identity (`PeerSource` or equivalent) and be ordered atomically with the lifecycle's current-connection check, rather than exposing an address-only callback between the identity check and `publish_ready()`.

Separately, longer-term it would be cleaner to share one `Arc<PeerLifecycle>` rather than constructing multiple wrappers around the same raw registry/lease stores, but I don't consider that a blocker for this PR.

gosunuts:
One more ownership/identity issue surfaced while reviewing the refactor.

`PeerLifecycle` is now documented as the sole mutation boundary for live leases and ready-peer metadata, but that invariant is not yet true across the production control paths. `NetworkControls::disconnect_matching()` still snapshots matching addresses under a read lock, then later removes whatever lease is currently stored at each address under a write lock. A same-address replacement can therefore occur between selection and removal, causing a stale disconnect decision (especially node-id based) to disconnect the replacement rather than the connection that was originally selected. The RPC `disconnectnode` path has the same class of problem: it validates against `ctx.peers`, then separately removes `peer_outbound[addr]` and mutates the registry directly.

I think these mutation paths should go through the same shared `Arc<PeerLifecycle>` and carry connection identity (`PeerSource` / connection id) whenever the operation is intended to target a specific peer. Otherwise the new "address equality does not establish connection identity" invariant is only enforced inside the listener/sync path, not at the actual ownership boundary.

There is also a smaller remaining TOCTOU in sync message delivery. `BlockSync` snapshots `ready_peers()` by address and later calls `lease(addr)` separately. If A is selected from the ready snapshot and B replaces A at the same address before the send, A's metadata (`start_height`, services, fanout eligibility) can be used to schedule work that is then queued onto B's lease. Since B's lease is registered before handshake completion and the writer exists before readiness publication, this can even send sync work to the replacement before it is the ready peer that justified the decision.

A cleaner long-term shape would be for ready snapshots to include connection identity, e.g. `{ info, source }`, and for delivery/disconnect APIs to accept that source (`lease(source)` / `send(source, msg)` / `disconnect_source(source)`) rather than resolving an address again later.

I consider the first point a merge blocker for this PR because it contradicts the ownership invariant this refactor is introducing. The ready-snapshot/send gap is the same underlying identity issue; it would be ideal to close here, but it could also be split into a focused follow-up if keeping this PR bounded is important.

metaphorics (commented):
Gate-and-merge hold: Gate 2 blocking plus a contract-drift blocking finding, head `df20115`.

Fixed on this branch already (`df20115`, verified locally: PR files `rustfmt --check` clean, `cargo clippy -p bitcoin-rs-p2p --all-targets` reports only two pre-existing `dispatch.rs` warnings from outside this PR): five files rustfmt flagged (`crates/node/src/run.rs`, `crates/p2p/src/connection.rs`, `crates/p2p/src/download.rs`, `crates/p2p/src/service.rs`, `crates/rpc/src/handlers/network.rs`) plus the overlong first doc paragraph on the `serve_with_shutdown` compatibility shim (`crates/p2p/src/listener.rs:220`).

Blocking, Gate 2 — these four fail identically on `main` HEAD 46c1973 (run 33709944082), i.e. pre-existing and outside this PR's 22-file set:
- `fmt` — 52 main-side files (none in this PR after the fix above).
- `clippy` — `bitcoin-rs-utxo` lib (`crates/utxo/src/connect.rs:100` too-many-arguments and doc paragraph).
- `deny` — license requirements plus duplicate-crate entries; this PR adds no dependencies.
- `test` — `crates/rpc/tests/handler_smoke.rs` (`Candidate`/`TemplateId`/`MempoolEntry`) and `spentby_tests::graph_ctx` (`compute_txid`, tuple arity), same as main.

Blocking, Gate 4 class 5 (contract drift) — the refactor removes/renames `NetworkHandles::peers`, `NetworkHandles::peer_outbound` and `NodeState::peers`, but `crates/rpc/tests/core_parity.rs`, which is outside this PR's file set, still calls all three, so the `core_parity` test target does not compile (`E0560`/`E0599`). The fix needs a file outside `gh pr diff 218 --name-only`, so this is blocking by definition.

Separately, `reviewDecision: REVIEW_REQUIRED` means the base-branch policy would refuse the merge even green.

Holding without merge; re-gate once main's jobs go green and the `core_parity` caller is migrated. No merge on this oid until then.

metaphorics:
@greptileai

metaphorics:
@sourcery-ai review

metaphorics:

metaphorics:

metaphorics:

metaphorics:

metaphorics:

metaphorics:

metaphorics:

Issue #224, opened 2026-08-31
Separate benchmark and evidence tooling from the node runtime


gosunuts:
## Goal

Keep `crates/node/src` focused on production runtime responsibilities and **aggressively reduce** benchmark-, replay-, experiment-, corpus-, and evidence-specific code.

The objective is not to preserve the existing benchmark suite under a cleaner directory. The objective is to keep only a small set of benchmarks that still protect an important performance property or answer an active engineering question, and delete the rest.

This is a focused execution issue under #217 and complements the configuration cleanup in #221.

## Problem

`crates/node` currently mixes production node architecture with code that exists primarily to measure, replay, export, or record a particular experiment/evidence target.

Concrete examples include:

- `crates/node/src/g14_utxo_commit.rs` — G14-specific UTXO commit timing sample selection and evidence-file emission;
- `crates/node/src/g2_muhash.rs` — G2-specific MuHash sampling and evidence persistence;
- `crates/node/src/corpus.rs` — corpus manifest/archive generation, validation, Bitcoin Core REST fetching, and active-chain export for replay/evidence workflows;
- `crates/node/examples/export_active_chain_corpus.rs` — corpus generation driver;
- `crates/node/examples/mainnet_prefix_replay.rs` — mainnet replay/performance harness;
- `crates/node/examples/verify_replay_durability.rs` — replay-result durability/evidence verifier.

These are not stable node-domain responsibilities or normal examples of using the public node API. They are benchmark/replay/evidence infrastructure. Keeping them in `src/` or `examples/` makes historical measurement workflows look like production architecture and encourages future one-off performance work to accumulate in runtime-facing surfaces.

More importantly, merely moving all of this code into `benches/` would preserve the same maintenance burden and historical baggage under a different path.

**Default to deletion.** Existing benchmark code does not earn preservation because it already exists or once produced useful evidence.

A benchmark should survive only when there is a clear answer to all of these questions:

1. What current performance property or architectural invariant does it protect?
2. What regression would it detect that ordinary correctness tests or a smaller benchmark would not?
3. Is someone expected to run it again when making relevant changes?
4. Is its result actionable enough to influence an implementation decision?
5. Is the benchmark substantially cheaper to maintain than recreating the measurement when it is next needed?

If those answers are weak, delete it.

## Retention policy

Treat every benchmark/evidence workflow as **delete unless justified**.

### Keep only when

A benchmark should remain when it does at least one of the following:

[...]

PR #227, opened 2026-08-31
refactor(node): remove obsolete benchmark paths

gosunuts:
## Summary

- remove the obsolete node-side diagnostic benchmark and G2/G14 runtime evidence hooks
- move corpus and replay measurement support under node examples
- remove stale G2/G14 scripts and gate tests
- prune UTXO commit benchmark no-op listener, duplicate build, and manual summary cases while retaining production-shaped coverage
- update plans, README, and benchmark documentation

## Verification

- `cargo test -p bitcoin-rs-utxo --lib --tests --no-default-features`
- `cargo check -p bitcoin-rs-utxo --all-targets --no-default-features`
- `cargo metadata --no-deps --format-version 1`
- `git diff --check`

The full node check is currently blocked by the pre-existing Windows `signal_hook::iterator` conditional-compilation error.


gosunuts:
Additional benchmark-retention review: I think this PR should go one pass further rather than stopping at the current G2/G14 cleanup.

The important rule from #224 is that `benches/` and `examples/` should not become an archive of old optimization campaigns or refactor proofs. Historical before/after evidence belongs in the PR/issue/docs; executable benchmark code should survive only when it still protects a current performance property.

Strong deletion candidates:

- `tools/benchmark-campaign/**`: this is a campaign-specific subsystem, not generic benchmark support. `runner.py` hard-codes an exact seven-pair / fourteen-arm matrix and a fixed performance gate. If that campaign is over, keeping the runner plus its own tests is pure historical infrastructure.
- Remaining G14 scripts and `bin/bitcoin-rs/tests/g14_guarded_run.rs`: the latter tests the G14 systemd/resource-guard harness rather than Bitcoin node behavior. Retiring G14 should retire these together.
- `crates/utxo/examples/utxo_memory_attribution.rs` and `snapshot_memory.rs`: these are measurement-campaign tools, including explicit G14/memory-campaign assumptions and duplicated RSS reporting. They should not remain permanent examples unless there is an active recurring consumer.
- `crates/utxo/benches/coin_stats_hotpath.rs`: much of this is an allocator A/B / rolling `CoinStatsListener` benchmark. If the default node path is moving away from the rolling listener, this should be deleted or reduced to only a small direct MuHash benchmark if that primitive still needs a regression guard.

Also audit the refactor-set benchmarks. Several permanently preserve the old implementation solely to keep measuring `before_*` vs `after_*`:

- `crates/rpc/benches/blocklookup.rs`
- `crates/rpc/benches/chaininfo.rs`
- `crates/rpc/benches/txoutproof.rs`
- `crates/mempool/benches/pareto.rs`
- `crates/mempool/benches/spentby.rs`
- `crates/index/benches/history_resolve.rs`

Once the new implementation is the architecture, the old scan/sort path should generally not live forever inside a benchmark. If the current production operation still deserves protection, keep only a small current-path benchmark that detects the relevant scaling/latency regression; otherwise delete the file.

`crates/node/benches/sync_pipeline.rs` is the main case where I would retain the performance question but aggressively shrink the harness. End-to-end apply/sync throughput is durable, but the current file mixes production-shaped sync with reverse-scan overflow, unsolicited ordering, oversized bursts, many-peer ticks, backend/index permutations, and manual summaries. Several of those are correctness/adversarial scenarios rather than distinct long-lived benchmark contracts. Prefer one small canonical node sync/apply benchmark with only materially distinct production shapes.

Conversely, benchmarks such as consensus verification/Merkle, storage backend operations, UTXO record codec/commit (after pruning historical/no-op arms), and mining candidate assembly have clearer durable performance ownership and are stronger keep candidates.

So the desired outcome should not just be “remove G2/G14 hooks.” It should be: fewer executable benchmarks overall, no campaign framework, no benchmark-framework tests for deleted infrastructure, no permanent old-vs-new implementation museum, and a small set of current production-path performance contracts.

gosunuts (commented):
The cleanup direction looks good overall, especially retiring historical benchmark/evidence machinery instead of keeping it in runtime paths. I do see two things to fix before merge:

1. `PLAN.md` is internally inconsistent after retiring G14. The PR says the former G14 artifact drivers/harness were retired and current performance contracts live in retained crate benchmarks, but the final Done definition still requires IBD to complete with `G3 + G14 reporting green`. That requirement is no longer executable under the new model. Please update the Done definition accordingly. The repository tree description for `scripts/` also still says `G2/G14 evidence-collection and measurement drivers`, which is now stale.

2. I would remove `.agent-tasks/node-runtime-benchmark-cleanup/` from the repository. The PR is deleting historical benchmark scaffolding, but these new GOALS/task-proof files replace some of that with agent execution metadata (searches performed, commands run, task-specific proof). The PR description/commit history is enough for that. Keeping task-run bookkeeping in the product tree risks recreating the same accumulation problem this PR is trying to solve.

The CoinStats change itself looks reasonable to me: making checkpoint/stable scans authoritative rather than maintaining a second rolling representation and asserting equality simplifies the model. I don't consider that part a blocker.

PR #235, opened 2026-09-01
Refactor tests around supported contracts

### LINKED ISSUE #234: Remove unsupported legacy code paths retained only by tests
## Goal

Enforce the repository's clean-cutover policy in production code: once an old format, implementation, compatibility path, or fallback is no longer supported, delete it instead of keeping it alive because tests still exercise it.

The durable rule should be:

> Tests protect supported contracts. They must not grant permanence to unsupported legacy code.

## Problem

bitcoin-rs explicitly documents a clean-cutover policy that rejects backward-compatibility shims, deprecated wrappers, transitional flags, and legacy fallback paths. The tree still contains code that appears to violate that policy because historical tests continue to call it.

A concrete current example is UTXO snapshot loading.

`crates/utxo/src/snapshot.rs` still exposes a permissive `read_snapshot()` path backed by `SnapshotReadPolicy::Legacy`, while the production checkpoint path uses the strict current-format readers instead.

Current references to the permissive reader are concentrated in tests such as:

- `crates/utxo/tests/snapshot_v4_golden.rs`
- `crates/utxo/tests/snapshot_roundtrip.rs`

The repository policy meanwhile states that clean cutovers should not retain unsupported compatibility machinery merely to preserve old formats.

This creates the wrong dependency direction:

```text
unsupported legacy implementation
        ↓
historical test still imports it
        ↓
legacy implementation remains production API
```

The test has effectively become the owner of the code's lifetime.

## Principle

For every legacy, fallback, compatibility, old-format, or retired implementation path, ask:

1. Is this path still part of a supported product contract?
2. Does production code actually depend on it?
3. Is there an explicit compatibility promise requiring it to remain?

If all answers are no, delete the implementation and the tests that exist only to preserve it.

Do not interpret the existence of a regression/equivalence/roundtrip test as evidence that the underlying old behavior must remain supported.

## Initial target: UTXO snapshot legacy reader

Audit the snapshot API and reduce it to the format/read policy actually supported by the current node.

If current checkpoint/recovery semantics require only the strict v4/current reader:

- remove `SnapshotReadPolicy::Legacy`;
- remove the permissive `read_snapshot()` wrapper;
- remove v2/v3 decoding branches that are not part of the supported storage contract;
- remove tests whose only purpose is to exercise those unsupported paths;
- retain a fixed golden compatibility test only for versions that are explicitly promised to remain readable;
- keep malformed/truncated/current-format rejection tests where they protect the active persistence contract.

If an old version is intentionally supported, document that compatibility promise explicitly and keep only the minimum fixed-vector test needed to protect it.

## Broader audit

Search the repository for similar patterns:

- `legacy` readers/writers not used by production;
- old-vs-new implementations retained solely as equivalence oracles;
- fallback paths whose only callers are tests;
- deprecated wrappers and aliases imported only by tests/examples;
- format migration helpers for rebuildable derived state that is supposed to use delete-and-rebuild;
- compatibility code preserved because a historical fixture still expects it;
- `pub` / `pub(crate)` surface widened only so tests can reach retired behavior.

For each case, classify it as:

- **supported contract** -> keep and document the compatibility boundary;
- **production requirement** -> keep for the real consumer;
- **test-owned legacy residue** -> delete code and test together.

## Rules

- Do not delete Bitcoin consensus, wire, RPC, or persistence compatibility that is explicitly supported.
- Do not replace removed legacy paths with new compatibility abstractions.
- Do not retain an old implementation solely as an oracle for the current implementation; use independent vectors/specifications/reference behavior instead.
- Rebuildable derived state should prefer reset/rebuild over legacy readers when that is already the project policy.
- A same-build roundtrip test does not establish a compatibility requirement by itself.
- Removing an unsupported path should also shrink any public API, feature flag, fixture, documentation, and CI surface that existed only for that path.

## Acceptance criteria

- [ ] The supported UTXO snapshot read contract is stated explicitly.
- [ ] `SnapshotReadPolicy::Legacy` and permissive legacy readers are removed unless an explicit current compatibility requirement justifies them.
- [ ] Unsupported v2/v3/currently-retired snapshot decoding is removed together with tests that exist only to preserve it.
- [ ] Retained snapshot compatibility tests use fixed authoritative/golden bytes for explicitly supported versions rather than same-build self-roundtrips.
- [ ] Repository-wide legacy/fallback/old-format paths are audited for production callers and explicit compatibility ownership.
- [ ] No code remains solely because a historical test imports or compares against it.
- [ ] Public/crate-public API exposed only for retired test/oracle paths is reduced.
- [ ] Clean-cutover documentation and implementation behavior no longer contradict each other.

## Related

- #122 — keep the permanent test surface narrow around durable contracts
- #228 — tests should enforce authoritative contracts rather than preserve duplicated implementation state
- #232 — remove development-time artifacts from permanent repository state
- #233 — prefer simplification over permanent verification scaffolding
- `docs/policies/source-compatibility.md`
- `docs/policies/db-migration.md`

gosunuts:
## Summary

- remove unsupported UTXO snapshot v2/v3 decoding and the legacy reader API
- keep the strict native v4 reader, fixed v4 golden-load coverage, and malformed/truncated input rejection
- remove implementation-detail, orchestration-only, umbrella smoke, and historical equivalence tests
- remove source-local BIP9 cache mechanics, shutdown synchronization mechanics, TxIndex reconciliation choreography, decompression counters, and the internal UTXO reference model
- remove unused future-only script stack, taproot, and Schnorr batch helper APIs and their dependencies
- reduce resolver fallback coverage to the valid eight-byte prefix-collision contract; stale, malformed, and manually blank rows are not promoted to contracts
- remove consumer-free alternate serialized/verified ingest and verified rollback APIs with their equivalence tests
- keep observable query, lifecycle, P2P, consensus, persistence, and RPC contracts covered
- update public exports, README, migration/source-compatibility policy, project concepts, benchmark notes, and task proof
- keep .gitignore unchanged with .agent-tasks/ ignored

Closes #234

## Verification

- cargo fmt --all -- --check
- cargo metadata --no-deps --format-version 1
- cargo test -p bitcoin-rs-primitives --lib --tests (26 tests)
- cargo test -p bitcoin-rs-index --lib --tests (33 tests)
- cargo test -p bitcoin-rs-chain --lib --tests (64 tests)
- cargo test -p bitcoin-rs-script --lib --tests (7 tests)
- cargo test -p bitcoin-rs-utxo --lib --tests (102 tests)
- cargo test -p bitcoin-rs-p2p --lib (49 tests)
- cargo check -p bitcoin-rs-p2p --tests
- remaining P2P integration tests (49 unit + 39 integration tests) pass; the existing outbound ban test hangs on Windows
- cargo test -p bitcoin-rs-rpc --lib --tests (378 tests)
- cargo test -p bitcoin-rs-mempool --lib --tests (127 tests)
- repository stale-reference scan and git diff --check

The node sync_smoke no-default-features check remains blocked by existing feature-gated signal_hook/empty-storage compilation errors; the default node build also requires unavailable Windows native libbitcoinkernel/libclang dependencies.


gosunuts:
The cleanup should go further than the current integration-test removals. The same rule should be applied to source-local `#[cfg(test)]` suites: if a test would need to change after a complete internal rewrite that preserves the supported Bitcoin/node behavior, it is probably testing implementation rather than a durable contract.

Concrete candidates I would audit in this PR or immediately after it:

- `crates/chain/src/bip9_cache.rs`: the unit tests mostly exercise `HashMap`-style cache mechanics (`insert/get`, `invalidate`, `clear`, `len`). The durable contract is correct BIP9 state, not the cache representation.
- `crates/node/src/shutdown.rs`: several tests pin the private `ShutdownBroadcast`/`Condvar`/atomic sequencing. Prefer one or two observable shutdown contracts over testing the synchronization implementation itself.
- `crates/node/src/txindex_worker_catchup_tests.rs`, `txindex_worker_reconcile_tests.rs`, `txindex_worker_query_tests.rs`: these are still a very large implementation test surface. Assertions around `ReconcileAction::Buffered/Progressed/CaughtUp`, pending deadlines, exact sync counts, and specific stale-prefix repair choreography should be reduced to observable contracts such as: the index eventually matches the authoritative applied tip, stale rows are never exposed as authoritative results, and unavailable/incomplete index state returns retry/unavailable rather than stale data.
- `crates/p2p/src/inbound.rs::from_decoded_is_source_less`: trivial constructor self-test; delete.
- UTXO source-local tests that count internal work (for example decompression-call counts) or compare against a second internal model should be reconsidered unless that exact behavior is a named current contract.

Also audit code whose only meaningful consumer is its own test. `script::stack` is explicitly future hand-rolled-interpreter infrastructure and currently appears to be public surface maintained largely by its own unit test. If there is no production consumer, remove the module/API rather than retaining it because it is tested. The same question applies to thin/future helpers such as the Schnorr batch/taproot helper modules if they are not actually used by the production path.

What should remain are tests with independent authority: consensus/BIP rules, Core compatibility, real P2P protocol behavior, current persisted-format integrity, authoritative state transitions, crash/restart behavior, and observable node/API contracts.

A useful review question for every remaining unit test is:

> If this subsystem were completely rewritten while preserving every supported external/consensus contract, should this test still pass unchanged?

If the answer is no, delete it or replace it with a higher-level contract test. The goal here should not just be fewer tests; it should be a smaller permanent architecture and authority surface.

PR #237, opened 2026-09-02
Require explicit resync for incompatible datadirs

### LINKED ISSUE #236: Support only the current datadir format and require explicit resync on schema changes
## Goal

Simplify persistent-state compatibility by supporting exactly one current on-disk format.

`bitcoin-rs` should not migrate, translate, or automatically recover incompatible persistent formats across schema revisions. If an existing datadir does not match the current schema, startup should fail clearly and require the operator to remove/recreate the datadir and resync.

## Why

The project already does not support in-place database migrations, but compatibility and fallback machinery still preserves multiple historical states:

- legacy UTXO snapshot readers and policy branches,
- old checkpoint component versions/codecs,
- automatic `HeadersOnly` / `Cold` fallback for incompatible checkpoint components,
- additive manifest compatibility rules,
- historical codec identifiers retained only so old checkpoints continue to load,
- tests and fixtures for old persisted layouts.

This creates a persistent-format state machine that every future storage change has to understand even though most node state can be rebuilt from the Bitcoin network.

The desired model is simpler:

```text
empty datadir            -> initialize current schema
current schema           -> open normally
anything else            -> fail startup with an explicit wipe/resync instruction
```

## Proposed policy

- Support only the current persistent format.
- Do not add in-place migration code.
- Do not retain legacy readers solely for backward compatibility.
- Do not silently downgrade an incompatible datadir into a partially reusable state.
- Detect incompatible storage at startup and fail before normal operation begins.
- Require the operator to explicitly delete/recreate the incompatible datadir and resync.
- Keep corruption detection, checksums, magic values, and current-format validation. Corruption detection is not migration.

A persistent format should therefore have:

```text
writer = one current writer
reader = one current reader
old version = incompatible error
corrupt current data = corruption error
missing data = new node
```

## Pre-1.0 compatibility policy

Until `bitcoin-rs` reaches 1.0, persisted database formats are **not a backward-compatibility contract**.

Schema changes before 1.0 should prefer the best current layout over preserving development history. An incompatible change bumps the datadir schema epoch and requires an explicit rebuild/resync. Do not add legacy readers, migration code, fallback decoders, compatibility defaults, or dormant version branches solely to keep a pre-1.0 datadir readable.

This applies to derived node state such as chainstate, UTXO, indexes, checkpoints, manifests, and KV layouts. Reusing authoritative source data that can be consumed directly by the current implementation (for example immutable block files) is fine; rebuilding current state from that data is not a database migration.

Reaching 1.0 is the point at which on-disk compatibility may become an explicit public contract. It does not imply that every future schema must be migrated in place; migration versus rebuild should then be decided deliberately per state class.

## Datadir schema marker

The KV stores currently have no schema metadata, so an incompatible key/value layout can be opened as if it were current and fail later during reads.

Introduce one small datadir-level schema epoch/version marker. Its purpose is only compatibility detection, not migration.

For example:

```text
CURRENT_SCHEMA = N
```

Startup behavior:

1. If the datadir is genuinely new/empty, initialize the current schema marker.
2. If the marker matches `CURRENT_SCHEMA`, continue.
3. If it is missing from a non-empty legacy datadir or has another value, refuse startup with a clear error explaining that the datadir must be removed and rebuilt.

A schema-breaking storage change increments this one epoch. No converter is added.

### UTXO snapshots

- Remove v2/v3 and other unsupported legacy snapshot readers.
- Remove `SnapshotReadPolicy::Legacy` and compatibility-only branches.
- Keep only the current strict snapshot reader/writer.
- Delete old snapshot fixtures/tests unless they protect the current format.

### Checkpoints

- Remove automatic version-mismatch recovery that turns incompatible persisted state into `HeadersOnly` solely to preserve old data.
- Remove compatibility logic whose only purpose is reading older manifests/codecs.
- Treat an incompatible checkpoint/datadir as an explicit startup failure.
- Keep normal cold-start behavior for a genuinely empty/new datadir.

### KV / block / undo storage

- Use the datadir schema epoch to reject old layouts before opening them as current state.
- Do not add per-version translation layers.
- Keep current-format corruption/integrity checks.

### Documentation/tests

- Replace the migration-policy document with a concise current-format / explicit-resync policy, or reduce it accordingly.
- Remove tests that assert backward compatibility with unsupported persisted versions.
- Retain tests that prove current-format read/write integrity, corruption rejection, schema mismatch rejection, and clean initialization.

## Error behavior

The startup error should be actionable and explicit, for example:

```text
This datadir was created by an incompatible bitcoin-rs schema.
Remove or replace the datadir and restart to perform a full resync.
```

The node must not delete user data automatically.

## Non-goals

- No automatic datadir deletion.
- No in-place schema conversion.
- No compatibility matrix across historical releases.
- No keeping old codecs/readers as dormant fallback paths.
- No weakening validation of the current on-disk format.

## Acceptance criteria

- [ ] A single datadir schema epoch/version is checked before normal persistent-state use.
- [ ] A new/empty datadir initializes the current schema cleanly.
- [ ] A non-empty datadir with an old, mismatched, or absent legacy schema marker fails fast with an actionable wipe/resync message.
- [ ] Only the current UTXO snapshot format is readable/writable.
- [ ] Unsupported legacy snapshot readers/policies are removed.
- [ ] Checkpoint incompatibility no longer preserves historical formats through automatic compatibility fallback.
- [ ] No in-place migration or translation code is introduced.
- [ ] Before 1.0, no backward-compatibility code is added solely to preserve older persisted database formats.
- [ ] Backward-compatibility-only tests/fixtures are removed.
- [ ] Current-format corruption and integrity tests remain.
- [ ] Documentation states clearly that incompatible schema revisions require explicit operator resync.

Related: #234 (remove unsupported legacy code paths retained only by tests), #232 (remove development-time artifacts), #228 (retain authoritative test contracts).

gosunuts:
## Summary

- add and validate the datadir-wide `CURRENT_SCHEMA` marker before persistent state opens;
- make missing, mismatched, corrupt, or unsupported checkpoint state fail with an explicit remove/recreate/resync instruction;
- remove `HeadersOnly` recovery and legacy checkpoint compatibility state while retaining current v4 integrity checks;
- update the storage policy, project concepts, and regression coverage.

## Pre-1.0 persistence policy

Until `bitcoin-rs` reaches 1.0, persisted database formats are not a backward-compatibility contract. Schema changes should prefer the best current layout over preserving development history.

For pre-1.0 persisted state:

- an incompatible change bumps the datadir schema epoch and requires an explicit rebuild/resync;
- do not add legacy readers, migration code, fallback decoders, compatibility defaults, or dormant version branches solely to keep an older datadir readable;
- rebuilding current derived state from authoritative source data that the current implementation can still consume directly (for example immutable block files) is allowed and is not considered a database migration.

1.0 is the point at which on-disk compatibility may become an explicit public contract; it does not automatically imply that every later schema change must support in-place migration.

## Validation

- `cargo check -p bitcoin-rs-node --no-default-features --features fjall`
- `cargo test -p bitcoin-rs-node --lib --no-default-features --features fjall -- --test-threads=1` (419 passed)
- `cargo test -p bitcoin-rs-utxo --lib --tests` (101 passed)

Closes #236


gosunuts (commented):
Direction is good and this gets much closer to the clean-cutover policy in #236, especially by validating the datadir epoch before storage opens and removing `HeadersOnly` / legacy manifest compatibility. I still think a few correctness issues should be fixed before merge:

1. Please do not classify arbitrary I/O failures as an incompatible datadir that should be deleted/resynced. `PermissionDenied`, transient filesystem errors, descriptor exhaustion, mount problems, etc. are operational failures, not schema incompatibility. These should preserve the original error and must not recommend destructive remediation. Reserve the explicit remove/resync path for actual schema/version/codec mismatches or authenticated current-format corruption.

2. `CURRENT_SCHEMA` creation should be crash-atomic. The current `create_new -> write_all -> sync_all -> sync_dir` sequence can leave an empty/partial marker if the process dies after creation but before the write is durable. On the next startup that genuinely fresh datadir is treated as incompatible and requires manual deletion. Prefer temp-file + fsync + atomic rename + directory fsync (or an equivalent scheme that cannot strand a malformed committed marker).

3. The schema epoch currently has two mutable authorities: `CURRENT_SCHEMA_VERSION = 1` and `CURRENT_SCHEMA_BYTES = b"1\n"`. A future bump can update one and not the other, defeating the guard. Keep one source of truth and serialize/parse around that value.

4. The Windows signal-handling change is unrelated to the datadir-format cutover and has no Windows validation listed in this PR. Please split it unless it is required for this change.

Once the first three are addressed, the simplified `Cold | Complete | Error` startup model looks much cleaner than the old fallback state machine.

gosunuts (commented):
Additional legacy-cleanup audit:

The checkpoint/UTXO cleanup itself is mostly complete on this head. `snapshot.rs` has only the strict v4 reader/writer, the old v2/v3-specific UTXO error branches are gone, `CheckpointLoad::HeadersOnly` is gone, the old manifest `serde(default)` compatibility test is gone, and the `Rolling` / `Zero` trailer branches are gone.

There are still a few pieces of historical-format state worth addressing before calling the clean cutover complete:

- `COINSTATS_CODEC` is still literally `"bitcoin-rs-coinstats"`, with comments saying the old crate name is deliberately retained. Now that pre-marker datadirs are rejected before checkpoint load, there is no compatibility reason to make *new* schema-1 checkpoints continue writing that historical identifier. This PR is the natural cutover point to rename it to a current identifier and update the stale comments in `crates/utxo/src/stats.rs` / README.

- The transaction index still has a separate legacy-format state machine: `IndexFormat::Legacy`, `LegacyCursorlessIndex`, unsupported format detection, and `NodeState` automatically calls `IndexWriter::reset_index()` and rebuilds incompatible derived index state. `docs/getting-started.md` still explicitly documents automatic deletion/rebuild of an existing legacy index. This is inconsistent with the new top-level wording that the datadir has one current format and incompatible persisted state requires explicit operator action, unless derived indexes are intentionally carved out as an exception. Either remove this legacy path/tests/docs and let the datadir epoch own compatibility, or explicitly document derived indexes as a separate rebuildable cache contract.

- `TrailerKindV1` now has exactly one value (`Scanned`) but `trailer_kind` remains persisted and tested. Since the field only existed to distinguish the removed `Rolling` / `Zero` historical cases, it looks vestigial under a single-current-format policy. Not a blocker by itself, but this is the cheapest time to delete it instead of making the new schema preserve another historical discriminator.

One related but separate residue: `crates/utxo/src/record.rs` still contains substantial `LEGACY_INLINE_*` state and `legacy_inline_len` logic. That is not the old persisted snapshot reader, so I would not mix it into the datadir-format blocker without checking its runtime purpose, but it does mean the repository-wide legacy audit from #234 is not literally finished just because the snapshot reader was removed.

gosunuts (commented):
I would push this PR one step further and make schema epoch 1 a true clean-cutover baseline: remove all persistence compatibility state that exists only because older bitcoin-rs layouts once existed.

Concretely, I think this PR should delete rather than preserve the remaining historical state:

- Rename `COINSTATS_CODEC` to a current identifier now. Keeping `"bitcoin-rs-coinstats"` solely because old checkpoints used that spelling defeats the point of rejecting every pre-epoch datadir before checkpoint load.
- Remove the derived-index legacy compatibility state machine (`IndexFormat::Legacy`, legacy cursorless detection, old format-version handling, startup reset/rebuild paths that exist specifically to consume old index layouts) unless there is an explicit product contract that derived indexes are intentionally exempt from the datadir epoch. If the epoch covers the datadir, old index layouts should be impossible to admit as current state.
- Remove compatibility-only tests and docs for those old index states along with the code.
- `TrailerKindV1` now has only `Scanned`; if the field exists only to distinguish historical `Rolling`/`Zero` variants, remove the discriminator from the new current format instead of carrying a permanently-singleton historical field forward.
- Audit current persisted/internal record layouts for historical fields/names such as `legacy_inline_len` / `LEGACY_INLINE_*`. If they still encode a real current invariant, rename and define that invariant without legacy semantics; if they exist only to preserve an older representation, remove them while this schema cutover explicitly permits a breaking format reset.

The rule I would apply is stronger than “old readers are gone”:

> The first datadir schema epoch should not inherit compatibility identifiers, branches, fields, or tests whose only owner is a pre-epoch bitcoin-rs format.

Consensus, wire, and explicitly supported external compatibility are different and should remain. But rebuildable bitcoin-rs-local persistence has no reason to start epoch 1 with historical baggage already embedded in the new baseline.

That also makes future schema bumps much easier to reason about: one current writer, one current reader, one current semantic vocabulary, and no hidden per-subsystem legacy state machines underneath the datadir marker.

gosunuts:
Addressed the review follow-up in `e175494`:

- `CURRENT_SCHEMA` initialization now writes and fsyncs a temporary marker, atomically renames it, and syncs the containing directory where supported.
- Schema read failures preserve ordinary I/O errors; only malformed/current-epoch mismatches produce the explicit resync guidance. The schema epoch has one serialized source of truth.
- Removed the derived-index format state machine, legacy index tests, and automatic index reset/rebuild path. `CURRENT_SCHEMA` is the single compatibility boundary.
- Removed the UTXO trailer discriminator, renamed the current CoinStats codec identifier, and renamed the current UTXO inline-layout terminology.
- Reverted the unrelated Windows signal change from this PR.

Validation: formatting, index tests, UTXO tests, and the node library suite (419 tests) passed locally. The Linux cross-check could not run because this Windows host lacks the Linux C compiler/headers.

Issue #226, opened 2026-08-31
Audit and optimize ScriptIndex on-disk representation


gosunuts:
## Goal

Audit the current ScriptIndex storage layout and select the durable row formats needed by #225, especially the Live locator. Non-format #225 design work may proceed while this audit runs.

The current layout is already compact and intentionally electrs-shaped, but it has not yet been demonstrated to be the best trade-off for bitcoin-rs across disk size, IBD write cost, reorg behavior, and explorer query latency.

Because ScriptIndex is rebuildable derived state, **do not constrain this work by backward compatibility with the current on-disk format.** If the chosen representation changes, delete the old ScriptIndex database and rebuild it. Evaluate the best final format now rather than preserving an intermediate layout or designing migration machinery.

## Storage compatibility policy

Previous ScriptIndex storage formats are disposable.

If this issue selects a different row format, key encoding, value encoding, column-family layout, watermark representation, or index version, the supported transition is:

```text
stop node
  -> mark each incompatible derived capability unavailable
  -> delete only the affected capability rows when separation is safe
  -> start node
  -> rebuild affected derived state from authoritative sources
```

Do not implement:

- dual-read or dual-write compatibility paths;
- in-place migrations from the current ScriptIndex format;
- legacy key/value decoders retained solely for old index files;
- format-version bridges that preserve obsolete layouts;
- conversion passes whose purpose is to avoid a clean reindex.

Reindex cost is still important because it affects operational recovery and rebuild time, but **migration compatibility cost is not a design criterion**.

The compatibility contract that matters is query/API behavior. Existing ScriptHistory/Esplora semantics may need to remain stable, but existing index bytes do not.

This rule applies only to rebuildable derived index data. It does not weaken compatibility or durability requirements for authoritative consensus, chain, or UTXO state.

## Current layout

The main durable rows are currently:

```text
TxConfirmed
  key   = txid-prefix(8) + height(4)
  value = TxPosition[n]              # 8 bytes each

Funding
  key   = script-hash-prefix(8) + height(4)
  value = TxPosition[n]              # 8 bytes each

Spending
  key   = outpoint-prefix(8) + height(4)
  value = empty

BlockHeaders
  key   = serialized header(80)
  value = empty
```

`Funding` and `TxConfirmed` group multiple transaction positions under one `(prefix, height)` key. This avoids repeating the prefix and height for every occurrence and should remain a baseline to beat.

[...]

Issue #225, opened 2026-08-31
Add a compact live UTXO view and modes to ScriptIndex


gosunuts:
## Goal

Keep the existing historical `ScriptIndex` needed for Esplora/explorer address history, while adding a compact live UTXO view so current address/scripthash UTXO queries do not scale with the script's full historical activity.

This extends the generic `ScriptIndex` introduced by #124; it does not add another explorer database or a separate indexing subsystem.

## Problem

The current `ScriptIndex` stores historical funding and spending activity. For `/address/:address/utxo` and `/scripthash/:hash/utxo`, the query path effectively:

1. scans historical funding rows for the script;
2. resolves matching outputs from transaction/block storage;
3. checks each historical outpoint against spending rows;
4. returns only outputs without a spender.

This means current UTXO lookup cost grows with **historical script activity**, not with the number of UTXOs currently owned by the script.

That is particularly poor for high-activity wallets: a script may have hundreds of thousands of historical outputs while only a small number remain unspent.

Removing historical `ScriptIndex` data is not an option for full explorer compatibility. Esplora address history, funding/spending statistics, pagination, and spent-output history still require historical funding/spending state.

## Design

Extend the existing `ScriptIndex` with a compact live-output view:

```text
ScriptIndex
├── Live
│   └── script -> current outpoints
└── History
    ├── Funding
    └── Spending
```

`Live` is a secondary view over the authoritative node UTXO set, not another copy of chainstate.

A target representation should be close to:

```text
key   = script-hash-prefix || outpoint
value = empty
```

or another similarly compact representation.

Do not duplicate authoritative coin fields such as value, full `scriptPubKey`, height, coinbase metadata, or other `Coin` state.

UTXO queries then become:

```text
script
  -> Live outpoints
  -> authoritative UTXO set
  -> Coin data
```

If the existing lossy script-hash prefix representation is reused, the final coin's full `scriptPubKey` must be exact-checked before returning it.

## Storage compatibility policy

`ScriptIndex` is derived, rebuildable state. **Backward compatibility for previous ScriptIndex database formats is intentionally out of scope.**

If #225 or #226 changes the durable row layout, column-family contents, key encoding, value encoding, watermark representation, or index version, the expected upgrade path is:

```text
stop node
  -> mark each incompatible derived capability unavailable
  -> delete only the affected capability rows
  -> start node
  -> rebuild affected Live state from authoritative UTXO state
  -> rebuild affected History state from retained block history
```

[...]

PR #240, opened 2026-09-02
feat(index): add the ScriptLive capability with anchored, ordered mutations

rabbitson87:
The index-crate half of #225: the `ScriptLive` capability, with the row format #226 names as the baseline. Node wiring — the undo-record anchor, boot-time seeding, `scriptindex=utxo|full` modes, and query routing — is the next slice; **nothing enables the capability yet**, so existing deployments build exactly what they built before.

Sequenced the way the two issues themselves ask: #226 pins Q5 (the Live locator) as the *only* format blocker, and both issues state the full-outpoint baseline — `scan-prefix(8) || txid(32) || vout(4)`, empty value — as the reference representation. If the #226 campaign later selects a smaller locator, Live is the cheap capability to migrate: selective reset plus a re-seed from a UTXO scan, no history replay.

## What the acceptance criteria forced, and how each is held

**Anchored deletes — the unanchored path refuses to exist.** A block body carries an input's prevout but not the spent coin's script, so preparation of a live transition takes a `SpentCoinScripts` source (in production: the undo record, whose restores hold the full spent `TxOut`). `prepare_block_for` returns `MissingSpentScripts` the moment `ScriptLive` is selected; an anchor that cannot resolve an external spend fails the whole preparation with `MissingSpentCoin`. Fail-closed, because the alternative is a spent output that stays visible.

**Ordered mutations in an otherwise order-free pipeline.** Every existing row family is append-only puts, which is why `PendingRows::sort` could always sort and dedup them. Live ops mix puts and deletes, and a later block in one committed batch may delete a key an earlier block inserted — so `live_ops` is never sorted, and `apply_live_ops` coalesces last-operation-per-key in memory instead of trusting any backend's write-batch ordering. Rollback is the same walk inverted, in reverse.

**The spendability predicate is UTXO admission, not the history predicate.** `build_utxo_changes` skips `is_op_return()`, scripts over 10,000 bytes, and the genesis coinbase. Live now skips exactly those; history deliberately keeps oversized-script outputs, and a test asserts that asymmetry on purpose. The bound is duplicated as `MAX_LIVE_SCRIPT_SIZE` (this crate does not depend on consensus); the node slice adds the cross-crate equality assertion.

**Same-block create-and-spend cancels before resolution** — such an output never reaches the committed UTXO set, and its spend needs no anchor entry, because the undo record would not have one either.

**Point deletes are identity-safe by construction.** Colliding scripts share the lossy 8-byte prefix but never a whole key. Prefix scans over the family are read-only by contract.

**Independent watermarks.** `ScriptLive` gets its own cursor, reset-selector bit (mask bit 2 — absent from every pre-#225 marker, which still reads back), and column-family mapping in the reset machinery. A historical commit does not stamp the live cursor and vice versa, so adding Live cannot invalidate a ready history backfill.

**Seeding stamps last.** Rows go out in bounded deferred batches; the watermark lands in one final durable write. A crash mid-seed leaves no cursor — not ready, partial rows unqueryable, recovery is reset-and-reseed. Re-seeding over a stamped cursor is refused.

## `IndexCapabilities::ALL` changed meaning, and call sites now say what they meant

Six roundtrip tests and the `commit_rollback_one` convenience used `ALL` to mean the historical pair. With a third capability, that reading changes silently — and the prepared path rightly refused them with `MissingSpentScripts`, which is how they were found. They now use the new `HISTORICAL` constant, the semantic they always had, rather than `ALL` keeping a name that no longer covers everything.

## Mutation audit

Each mutant killed by exactly the test aimed at its property:

| mutation | killed by |
| --- | --- |
| resolve external spends before same-block cancellation | `same_block_create_and_spend_cancels_without_an_anchor` |
| drop the oversized-script bound | `live_predicate_matches_utxo_admission_not_history` |
| rollback stops inverting in `apply_live_ops` | `live_rows_follow_connect_spend_and_disconnect` |

## An incidental toolchain finding

rustc 1.95.0 ICEs in `check_match` instead of reporting E0004 when a `ColumnFamily` match goes non-exhaustive — so the compiler cannot enumerate the sites a new enum variant breaks. The affected matches were found by grepping for the previous last arm. Recorded for whoever next adds a variant here.

## Verified

```
cargo clippy -p bitcoin-rs-index --all-targets --features rocksdb \
  -- -D warnings                                                        0 errors
cargo test -p bitcoin-rs-index --features rocksdb                       62 passed
cargo check --workspace                                                 ok
```

### DIFF
--- a/crates/index/src/index.rs
+++ b/crates/index/src/index.rs
@@ -97,6 +97,29 @@ pub enum IndexError {
     /// A prepared transition cannot mix with legacy buffered rows.
     #[error("cannot write prepared TxIndex mutations with buffered legacy rows")]
     PendingLegacyRows,
+    /// A capability set containing `ScriptLive` was prepared through a path
+    /// that carries no spent-coin script source. Live deletes need the spent
+    /// coin's exact script (#225), and a prevout-only block parse cannot
+    /// produce it.
+    #[error("ScriptLive preparation requires a spent-coin script source")]
+    MissingSpentScripts,
+    /// The spent-coin source could not resolve an external input's script.
+    /// Failing closed here is deliberate: a missing anchor means the Live view
+    /// would silently keep a spent output alive.
+    #[error("no spent-coin script for outpoint {txid:02x?}:{vout} in block at height {height}")]
+    MissingSpentCoin {
+        /// Spent transaction id (little-endian bytes).
+        txid: [u8; 32],
+        /// Spent output index.
+        vout: u32,
+        /// Height of the spending block.
+        height: u32,
+    },
+    /// `seed_script_live` was asked to seed over an existing live watermark.
+    /// Seeding assumes a fresh (or reset) capability; overwriting a live view
+    /// in place is how partial states become queryable.
+    #[error("ScriptLive is already seeded; reset the capability first")]
+    LiveAlreadySeeded,
     /// `TxIndex` tables exist but the format-version key is missing.
     #[error("TxIndex tables are present without a versioned watermark")]
     LegacyCursorlessIndex,
@@ -111,6 +134,7 @@ const FORMAT_VERSION_KEY: &[u8] = &[0x00, b'V'];
 const FORMAT_VERSION_VALUE: [u8; 4] = [0x03, 0x00, 0x00, 0x00];
 const TX_LOOKUP_WATERMARK_KEY: &[u8] = &[0x00, b'T'];
 const SCRIPT_HISTORY_WATERMARK_KEY: &[u8] = &[0x00, b'S'];
+const SCRIPT_LIVE_WATERMARK_KEY: &[u8] = &[0x00, b'L'];
 const RESET_CAPABILITIES_KEY: &[u8] = &[0x00, b'R'];
 const WATERMARK_LEN: usize = crate::types::HEIGHT_SIZE + 32;
 const RESET_SCAN_LIMIT: PrefixScanLimit = PrefixScanLimit {
@@ -122,6 +146,7 @@ const fn watermark_key(capability: IndexCapability) -> &'static [u8] {
     match capability {
         IndexCapability::TxLookup => TX_LOOKUP_WATERMARK_KEY,
         IndexCapability::ScriptHistory => SCRIPT_HISTORY_WATERMARK_KEY,
+        IndexCapability::ScriptLive => SCRIPT_LIVE_WATERMARK_KEY,
     }
 }
 
@@ -157,6 +182,10 @@ pub enum IndexCapability {
     TxLookup,
     /// `ScriptIndex` scripthash funding and spending rows.
     ScriptHistory,
+    /// `ScriptIndex` live-output rows: one row per currently unspent outpoint,
+    /// filed under its script (#225). Rebuildable from the authoritative UTXO
+    /// set alone, unlike history.
+    ScriptLive,
 }
 
 /// Capabilities included in one prepared index transition.
@@ -166,6 +195,8 @@ pub struct IndexCapabilities {
     pub tx_lookup: bool,
     /// Build `ScriptIndex` funding and spending rows.
     pub script_history: bool,
+    /// Build `ScriptIndex` live-output rows.
+    pub script_live: bool,
 }
 
 impl IndexCapabilities {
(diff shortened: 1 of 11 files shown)


gosunuts:
I don't think this durable DB change should land incrementally across the PR stack.

`ScriptLive` changes the on-disk contract: it adds a new column family and row format, new watermark/reset metadata, and new recovery semantics. But the lifecycle of that format is split across multiple PRs: undo anchoring is in #241, while UTXO seeding, enablement/query routing, and the final recovery path are still deferred.

For an API refactor, stacked intermediate states are fine. For a persistent database contract, I think the bar should be different. The complete transition should land in one PR and be tested as one unit, including at least:

- physical row/CF definition across all backends;
- format/version and reset behavior;
- fresh initialization and existing-index rebuild;
- interrupted reset/seed recovery;
- connect and reorg rollback;
- restart/resume;
- pruning interaction;
- authoritative UTXO seeding;
- query resolution;
- sync-level validation against authoritative UTXO state.

Otherwise each PR can be locally correct while `main` temporarily contains an incomplete durable state. The fact that `ScriptLive` needs coordinated changes across reset, watermark, worker, undo, seeding, and query paths is exactly why I would rather review and test the DB transition atomically.

Also, #226 still leaves the durable Live locator open pending Q5. I would avoid committing one physical representation now and changing the DB contract again immediately after that benchmark decision.

My preference is: finish #226 Q5 first, then combine #240, #241, seeding/enablement/query wiring, and the DB recovery behavior into one storage-transition PR, and run sync/restart/reorg/reindex validation on that complete change before merging it.

gosunuts:
One more requirement: this should not be treated as an implementation detail local to this PR.

A durable database-format change needs an explicit design/evidence gate before it is merged. #226 is the right place to make that decision: the final `ScriptLive` row/locator format, version/reset semantics, rebuild contract, and rejected alternatives should be documented there with enough measurements and reasoning that the storage contract is reviewable independently of the implementation PR.

I would therefore treat completion of the relevant #226 decision as a prerequisite for the DB-format-changing PR, rather than landing a provisional physical format and revising it later.

The final storage-transition PR also needs an end-to-end synchronization experiment, not only unit/integration tests. We need evidence that a node using the new format can actually complete synchronization and that the resulting derived state agrees with the authoritative chain/UTXO state. At minimum, the result should record the synced height/tip, successful restart/resume, and a post-sync consistency check of `ScriptLive` against authoritative UTXO resolution. Reindex/reset should be exercised as part of the same validation where practical.

For a persistent DB contract, "the code path is tested" is not enough; we need evidence that the complete database state machine converges successfully under a real synchronization run.

rabbitson87:
Agreed on both points, and the second one corrects a mistake I made rather than a preference I disagree with.

## The stack argument is right, and this PR's own description says so

I wrote "node wiring follows in the next slice" as if that were a reviewability benefit. For an API refactor it would be. For a durable format it is exactly the failure you name: between #240 and the last slice, `main` carries a column family, a row encoding, a reset-selector bit and a watermark whose *lifecycle* — seed, enable, query, recover — does not exist yet. Each PR is locally green and the tree is collectively half-defined. Splitting bought reviewability and paid for it in a property that matters more here.

## On Q5 I conflated a baseline with a decision

My reasoning was that #225 and #226 both name the full-outpoint layout (`scan-prefix(8) || txid(32) || vout(4)`), so building on it was not pre-empting the audit. Re-reading #226, that is wrong: it lists the full-outpoint form as **candidate E's baseline** — the thing smaller locators are measured *against* — and states Q5 is the one storage-format blocker for the durable Live rows. A baseline is what a decision is compared to, not the decision. I treated the two as the same and they are not.

So committing this representation now would mean either changing the on-disk contract again right after the benchmark, or letting an unmeasured format become the default by inertia. Neither is what the audit is for.

## What I am doing

**Closing #240 and #241** rather than leaving them open as a pending stack — a half-landed durable contract should not sit in the queue looking like it is on its way in. The work is not lost:

- The branches (`feat/225-script-live`, `feat/225-live-anchor`) stay pushed, so the eventual storage-transition PR can cherry-pick or rebuild from them rather than starting cold.
- Several findings in them are independent of the row format and survive whatever Q5 selects: live mutations are ordered where every other row family is a sorted set; the spendability predicate must be UTXO admission (`is_op_return` + `MAX_SCRIPT_SIZE`) and not the history predicate; same-block create-and-spend has to cancel *before* anchor resolution or the undo record is asked for an entry it structurally lacks; the undo record's restores are the exact spent-coin script anchor, block-hash-bound so a stale record refuses the job. I will carry these into #226 as design input so the audit does not have to rediscover them.
- One toolchain note worth keeping wherever this lands: rustc 1.95.0 ICEs in `check_match` (not E0004) on a non-exhaustive `ColumnFamily` match, so adding a variant gives no compiler-enumerated list of affected sites.

**Not doing** the seeding/enable/query slices against a provisional format.

## On the sync-run evidence requirement

Taking this as a hard gate, and flagging one thing about it now rather than at the end: an end-to-end mainnet sync with a post-sync `ScriptLive` ↔ authoritative-UTXO consistency check is a multi-day run on this hardware, and #226's own campaign already carries a disk contract (one canonical corpus, one disposable fixture, declared footprint and free-space cadence). I would rather agree the corpus and the consistency-check shape as part of the Q5 decision than improvise them when the implementation PR is otherwise ready. If a pinned-height regtest/signet run plus a bounded mainnet prefix is acceptable evidence, that changes the schedule materially — worth settling in #226.

Ordering I will follow unless you say otherwise: #226 Q5 decision (documented with measurements and rejected alternatives) → one storage-transition PR covering row/CF across all backends, format-version and reset, fresh init and rebuild, interrupted reset/seed recovery, connect and reorg rollback, restart/resume, pruning interaction, UTXO seeding, query resolution → sync/restart/reorg/reindex validation on that complete change.

PR #241, opened 2026-09-02
feat(node): anchor ScriptLive transitions to the block's undo record

rabbitson87:
Stacked on #240 (base: `feat/225-script-live`) — the node half of #225's write path. **Nothing enables `ScriptLive` in production yet**; `tx_index_capabilities` still builds it `false`. This is the machinery the enable will use, landed separately so the enable itself is a small, arguable diff.

## The undo record is the anchor

A block's undo record restores are exactly the coins it spent from the committed UTXO set, each carrying its full `TxOut` — the durable spent-coin script source #225 requires:

- present for **connect** (anchoring the Live delete) and **disconnect** (anchoring the re-insert);
- absent for intra-block spends *by construction*, which matches the index's same-block cancellation;
- bound to its block hash by the codec, so a stale record from an abandoned branch **refuses to become an anchor** rather than supplying the wrong scripts.

`UndoScripts` decodes one record into a `SpentCoinScripts` map. The worker reaches records through the handle it already holds: `PruneBodyStore` gains a defaulted `undo_record`, overridden by the flat-file store against the chainstate KV index where `apply` persists the record before publishing the tip.

## Fail-closed at every live-selecting step

Both forward prepare sites, the rollback prepare, and the rollback commit branch through the anchored variants when `capabilities.script_live` is set. A missing or undecodable record is `UndoUnavailable` — a worker error the operator sees — because the alternative is a spent output that stays visible in the view.

## Live joins the existing per-capability partitioning

`forward_selection` / `rollback_selection` carry it as a third member, so a live cursor behind history advances in its own group the way a fresh capability backfill already does. An absent live cursor means *advance from genesis with undo anchors* — the cold-start and headers-only path. The upgrade path for a node with an existing full history is **boot-time seeding from the restored UTXO set** (next slice), which stamps the cursor at the restored tip and skips that walk entirely.

## The predicate duplication is now held together

`bitcoin-rs-index` carries the live script-size bound as its own literal (it cannot depend on the consensus crate). `script_live_size_bound_matches_utxo_admission` — in the node, which depends on both — pins `MAX_LIVE_SCRIPT_SIZE == MAX_SCRIPT_SIZE` beside the admission rule it mirrors, so the two cannot drift silently.

## Tests

- `anchor_serves_the_records_restored_scripts` — the adapter answers exactly what the record restores, nothing else.
- `anchor_refuses_a_record_for_another_block` — hash binding enforced at the anchor boundary.
- The cross-crate constant assertion above.

The anchored end-to-end worker run deliberately lands with the enable (next slice), where it asserts against a real configuration instead of a synthetic one.

## A toolchain note

rustc 1.95.0 ICEd in `check_mod_deathness` the moment `UndoScripts` existed without a caller — the same diagnostic crash hit twice before in this campaign — so the adapter and its call sites land in one commit rather than as reviewable halves.

## Verified

```
cargo test -p bitcoin-rs-node --no-default-features --features fjall    no failures
```

### DIFF
--- a/crates/node/src/apply.rs
+++ b/crates/node/src/apply.rs
@@ -467,6 +467,22 @@ pub(crate) trait PruneBodyStore: Send + Sync {
         Ok(Box::new(DirectPruneBodyReader { store: self }))
     }
 
+    /// The persisted undo record for `height`/`hash`, when this store can
+    /// reach one.
+    ///
+    /// The default answers nothing: only stores backed by the chainstate
+    /// key-value index hold undo rows, and a `ScriptLive`-selecting worker
+    /// step fails closed on `None` rather than indexing without its spent-coin
+    /// anchor (#225).
+    fn undo_record(
+        &self,
+        height: u32,
+        hash: bitcoin_rs_primitives::Hash256,
+    ) -> Result<Option<Vec<u8>>, StorageError> {
+        let _ = (height, hash);
+        Ok(None)
+    }
+
     /// Loads `len` body bytes starting `offset` bytes into the serialized block.
     ///
     /// Defaults to `Ok(None)`, meaning "this store cannot slice"; callers fall
@@ -645,6 +661,17 @@ impl<S: KvStore> FlatFilePruneBodyStore<S> {
 }
 
 impl<S: KvStore> PruneBodyStore for FlatFilePruneBodyStore<S> {
+    fn undo_record(
+        &self,
+        height: u32,
+        hash: bitcoin_rs_primitives::Hash256,
+    ) -> Result<Option<Vec<u8>>, StorageError> {
+        self.index.get(
+            bitcoin_rs_storage::ColumnFamily::UndoData,
+            &bitcoin_rs_storage::pruning::block_undo_key(height, hash),
+        )
+    }
+
     fn disk_usage(&self) -> Option<u64> {
         Some(self.files.disk_usage())
     }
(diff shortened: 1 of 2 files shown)


rabbitson87:
Closing alongside #240, for the reason given there: `ScriptLive` changes a durable on-disk contract, and its lifecycle must land as one storage-transition PR rather than a stack whose intermediate states leave `main` with a half-defined format. #226's Q5 decision is a prerequisite for that PR, and this branch builds on a representation Q5 has not selected — the full-outpoint layout is that audit's *baseline candidate*, not its outcome.

The branch stays pushed (`feat/225-live-anchor`), and the parts that survive any Q5 outcome are worth naming, because they are properties of the anchor rather than of the row format:

- **The undo record is the right anchor, and its block-hash binding is load-bearing.** Restores are exactly the coins a block spent from the committed set, each with its full `TxOut`; intra-block spends are absent by construction, which is why same-block cancellation must happen before resolution rather than after. A record bound to another block refuses to become an anchor, so a stale row from an abandoned branch cannot supply the wrong scripts.
- **Live-selecting steps must fail closed without their anchor.** A missing or undecodable record cannot be skipped — the alternative is a spent output that stays visible.
- **The predicate duplication needs a cross-crate assertion.** `bitcoin-rs-index` cannot depend on the consensus crate, so it carries the script-size bound as a literal; the node depends on both and is where `MAX_LIVE_SCRIPT_SIZE == MAX_SCRIPT_SIZE` can actually be held.
- **rustc 1.95.0 ICEs in `check_mod_deathness`** the moment an adapter exists without a caller, which is why the adapter and its call sites could not be split into reviewable halves here.

I will carry these into #226 as design input so the audit starts from them rather than rediscovering them.

PR #243, opened 2026-09-02
Replace the boot-time legacy-body scan with lazy fail-closed body-position validation

gosunuts:
## What

Replaces the boot-time `BLOCK_DATA` legacy-body scan with lazy fail-closed body-position validation, after review. **The `CURRENT_SCHEMA` datadir epoch marker is kept** — per #236/#237 it is the complete startup-time schema discriminator for KV layouts, cheap fail-fast metadata rather than a legacy reader, and an untagged repo does not imply no persisted datadirs exist (a real pre-#237 mainnet datadir was exercised from this branch).

## Review findings addressed

1. **Marker restored.** `CURRENT_SCHEMA` machinery, its boot-path call in `NodeState::open`, and the three epoch tests are restored unchanged from `main`.
2. **Lazy fail-closed instead of a boot-time column-family scan.** The scan walked the whole `BLOCK_DATA` CF on every startup and is removed. In its place, every body-position decode distinguishes malformed rows from missing rows: `BlockFilePosition::decode` failure returns `StorageError::IncompatibleData` from `body_position`, the direct reader path, and `prefetch_positions` — never a silent `None` masquerading as a missing body.
3. **Policy text matches reality.** `docs/policies/db-migration.md` keeps the marker contract and gains an explicit lazy-validation bullet: body-position rows validate per read, fail-closed, never reported as absence.
4. **Clippy fixed.** `FlatFilePruneBodyStore::open` is infallible (`unnecessary_wraps`); `block_body_store` follows and drops its `data_dir` plumbing. `legacy_block_body_datadir_is_refused` (open-time refusal no longer true) is replaced by `malformed_body_row_is_incompatible_not_missing`.

## Verification

- `cargo check -p bitcoin-rs-node` clean
- `cargo clippy -p bitcoin-rs-node --all-targets` zero findings
- `cargo test -p bitcoin-rs-node --lib` **431 passed / 0 failed** (3 restored epoch tests + 1 new lazy-validation test)
- `docs/policies/db-migration.md` vs `main`: exactly the one lazy-validation bullet added

### DIFF
--- a/crates/node/src/apply.rs
+++ b/crates/node/src/apply.rs
@@ -514,7 +514,7 @@ pub(crate) struct FlatFilePruneBodyStore<S: KvStore> {
 enum PositionLookup {
     Direct,
     Prefetched {
-        entries: Vec<(u32, Hash256, Option<BlockFilePosition>)>,
+        entries: Vec<(u32, Hash256, BlockFilePosition)>,
         next: usize,
     },
 }
@@ -554,13 +554,17 @@ impl PruneBodyReader for FlatFilePruneBodyReader<'_> {
             .copied()
             .zip(values)
             .map(|((height, hash), value)| {
-                (
-                    height,
-                    hash,
-                    value.as_deref().and_then(BlockFilePosition::decode),
-                )
+                let position = value
+                    .as_deref()
+                    .and_then(BlockFilePosition::decode)
+                    .ok_or_else(|| {
+                        StorageError::IncompatibleData(format!(
+                            "block-body index row for height {height} is not a 16-byte flat-file position"
+                        ))
+                    })?;
+                Ok((height, hash, position))
             })
-            .collect();
+            .collect::<Result<Vec<_>, StorageError>>()?;
         self.positions = PositionLookup::Prefetched { entries, next: 0 };
         Ok(())
     }
@@ -579,7 +583,11 @@ impl PruneBodyReader for FlatFilePruneBodyReader<'_> {
                 else {
                     return Ok(None);
                 };
-                BlockFilePosition::decode(&encoded)
+                BlockFilePosition::decode(&encoded).ok_or_else(|| {
+                    StorageError::IncompatibleData(format!(
+                        "block-body index row for height {height} is not a 16-byte flat-file position"
+                    ))
+                })?
             }
             PositionLookup::Prefetched { entries, next } => {
                 let Some(&(expected_height, expected_hash, position)) = entries.get(*next) else {
@@ -596,33 +604,19 @@ impl PruneBodyReader for FlatFilePruneBodyReader<'_> {
                 position
             }
         };
-        let Some(position) = position else {
-            return Ok(None);
-        };
         self.files.load(position, height, *hash.as_byte_array())
     }
 }
 
 impl<S: KvStore> FlatFilePruneBodyStore<S> {
-    pub(crate) fn open(
-        index: Arc<S>,
-        files: Arc<FlatFileBlockStore>,
-        data_dir: &std::path::Path,
-    ) -> Result<Self, StorageError> {
-        for row in index.iter_prefix(bitcoin_rs_storage::pruning::BLOCK_DATA_CF, b"b")? {
-            let (key, value) = row?;
-            if key.len() == 37 && value.len() != BlockFilePosition::ENCODED_LEN {
-                return Err(StorageError::IncompatibleData(format!(
-                    "datadir {} predates the flat-file block store and must be resynced",
-                    data_dir.display()
-                )));
-            }
-        }
-        Ok(Self { index, files })
+    pub(crate) fn open(index: Arc<S>, files: Arc<FlatFileBlockStore>) -> Self {
+        Self { index, files }
     }
-
     /// Resolves the flat-file position of a block body, or `None` when the
-    /// block is unknown or its index row is not a decodable position.
+    /// block is unknown. An index row that is not a decodable 16-byte
+    /// flat-file position is `IncompatibleData`, never a silent `None`: the
+    /// row's presence means the body must exist, so treating a decode failure
+    /// as absence would hide a schema mismatch behind a missing-block answer.
     ///
     /// Every read path starts here, so it is written once rather than three
     /// times: divergence between the whole-body, ranged, and metadata lookups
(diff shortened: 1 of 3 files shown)


gosunuts (commented):
I agree with removing legacy readers/migration scaffolding, but this patch also removes two pieces of incompatibility detection that are part of the current-format contract.

The main concern is that removing `CURRENT_SCHEMA` is not the same thing as removing backward-compatibility support. #236 explicitly introduced a datadir-level epoch because the KV layouts do not otherwise provide a complete startup-time schema discriminator, and #237 closed that issue by validating the marker before persistent state opens. The marker is cheap fail-fast metadata, not a legacy reader or migration layer. The fact that no tagged release exists also does not imply that no persisted datadirs exist; this PR description itself mentions exercising a real pre-#237 mainnet datadir.

More concretely, deleting the legacy flat-file scan currently makes incompatible `BLOCK_DATA` rows fail open. `body_position()` does:

```rust
let Some(encoded) = self.index.get(...) ? else {
    return Ok(None);
};
Ok(BlockFilePosition::decode(&encoded))
```

so an old/non-position value now becomes `None`, exactly like a genuinely missing body. `load_block_body`, `load_block_body_range`, and `block_body_metadata` then propagate that as absence rather than `IncompatibleData`. That conflicts with the updated policy text claiming every persistent surface is strictly validated against the current format.

I think removing the full-CF boot scan is reasonable if the cost is undesirable, but the lazy decode path should at minimum distinguish malformed/incompatible rows from missing rows, e.g. convert decode failure to `StorageError::IncompatibleData`. Separately, I would keep a single datadir schema discriminator (or replace it with an equally complete startup-time format discriminator) so future incompatible KV/key-layout changes cannot remain latent until an affected row happens to be read.

There is also a current CI failure: after deleting the work from `FlatFilePruneBodyStore::open`, it still returns `Result<Self, StorageError>` even though it is now infallible, and `clippy -D warnings` rejects it as `unnecessary_wraps`.

Requested changes conceptually:

1. Preserve a complete startup-time schema incompatibility boundary instead of removing it solely because no tagged release exists.
2. Make malformed/non-current `BLOCK_DATA` rows fail with `IncompatibleData` rather than collapse to `None`.
3. Update the new policy text to match what is actually validated at startup vs lazily.
4. Fix the clippy failure.

The deletion direction is good; I just don't think compatibility cleanup should weaken fail-closed current-format validation.

gosunuts:
Findings addressed and force-pushed (`4edb1ff`):

- `CURRENT_SCHEMA` marker system restored unchanged from `main` (system, boot-path call, three epoch tests) — agreed it is the complete startup-time discriminator for KV layouts and cheap fail-fast metadata, not compat scaffolding.
- The boot-time `BLOCK_DATA` scan is still removed, but its guarantee is preserved lazily: `BlockFilePosition::decode` failure now returns `StorageError::IncompatibleData` from `body_position`, the direct reader, and `prefetch_positions` — a malformed row can no longer collapse to a missing-body `None`.
- `db-migration.md` restored to the marker contract and gains an explicit lazy-validation bullet (startup vs per-read responsibilities now match the code).
- `open` is infallible (clippy `unnecessary_wraps`), `block_body_store` follows and drops `data_dir`.
- `legacy_block_body_datadir_is_refused` is replaced by `malformed_body_row_is_incompatible_not_missing`.

Verified: clippy `-p bitcoin-rs-node --all-targets` zero findings; lib tests 431 passed / 0 failed.

gosunuts (commented):
Re-review after the latest update: the earlier architectural blockers are fixed. Keeping `CURRENT_SCHEMA`, making `FlatFilePruneBodyStore::open()` infallible, and moving block-position validation to the read path is a much cleaner shape. I no longer think this PR is introducing unnecessary complexity.

One remaining correctness issue: the prefetched path currently collapses two distinct cases into the same `IncompatibleData` error. In `prefetch_positions()`, this chain:

```rust
value.as_deref().and_then(BlockFilePosition::decode).ok_or_else(...)?
```

means both `None` (no row / body unavailable) and `Some(invalid_bytes)` (present but malformed/incompatible row) fail as `IncompatibleData`.

That differs from the direct path, which correctly preserves:

```text
missing row        -> Ok(None)
valid row          -> Some(position)
malformed row      -> IncompatibleData
```

This matters because `txindex_worker` uses `prefetch_positions()` and intentionally treats a later `load_block_body() == None` as `ReconcileAction::Stalled`. With the current prefetch implementation, a genuinely missing body can instead terminate the index worker as a storage error.

Please keep the same semantics in the prefetched representation, e.g. retain `Option<BlockFilePosition>` and only turn `Some(bytes)` + decode failure into `IncompatibleData`. A pair of regression tests for `prefetch + missing` and `prefetch + malformed` would lock this down.

Two non-blocking cleanup points:

- The PR description is stale: it still says the `CURRENT_SCHEMA` marker is removed / markerless, but the current code deliberately keeps it.
- The policy diff replaces the explicit `utxo-v4.dat` strict-reader bullet with the block-body bullet; those are independent contracts, so I would keep the UTXO bullet and add the block-body one rather than replace it.

After the prefetch semantic mismatch is fixed, this looks like a good simplification: remove the O(N) boot scan, keep one schema epoch boundary, and validate current-format rows fail-closed when they are actually read.

gosunuts:
The current direction looks much better: keeping `CURRENT_SCHEMA` while removing the boot-time `BLOCK_DATA` scan is the right split. There is still one correctness issue in the lazy prefetch path, though.

`prefetch_positions()` currently does:

```rust
let position = value
    .as_deref()
    .and_then(BlockFilePosition::decode)
    .ok_or_else(|| StorageError::IncompatibleData(...))?;
```

That collapses two different states into the same error:

- `value == None`: the row is genuinely missing
- `Some(value)` that does not decode: the row exists but is incompatible/malformed

Those need different semantics. `KvSnapshot::get_many_sorted()` explicitly returns one `Option<Vec<u8>>` per key, and `txindex_worker` already treats `load_block_body(...) == None` as a normal stall condition. With the current prefetch implementation, a missing body becomes `IncompatibleData` before `load_block_body()` gets a chance to return `None`, turning a normal stall into a worker storage failure.

The intended contract should be:

```text
missing row       -> Ok(None)
valid row         -> Ok(Some(position))
malformed row     -> Err(IncompatibleData)
```

I would keep `PositionLookup::Prefetched` capable of representing `Option<BlockFilePosition>` and centralize this decode rule in one helper used by `body_position`, the direct reader path, and prefetch. Right now the same semantic rule is duplicated across those paths, and this regression is exactly the kind of divergence that results from that duplication.

Please also add contract coverage for at least:

1. a missing prefetched row remains `Ok(None)` / stalls normally;
2. a malformed prefetched row returns `IncompatibleData`;
3. the direct reader path also returns `IncompatibleData` for a malformed existing row.

Once that is fixed, I think the PR scope and overall approach are in good shape.

PR #244, opened 2026-09-02
ci: reduce PR validation cost

### LINKED ISSUE #171: ci: reduce PR validation cost
PR CI is doing too much exhaustive validation on every commit. The goal should be a much faster edit → push → feedback loop, while keeping full validation on `main`.

Reduce PR CI aggressively:

- keep a small required PR gate: `fmt`, `deny`, `clippy`, normal workspace tests
- remove duplicated clippy work from `kernel-parity`
- move `bench-smoke` off PRs to `main` or scheduled runs
- move portable/full-backend/kernel deep validation to `main`, or gate it only on relevant paths/features
- avoid rebuilding the same Rust/native dependency graph across multiple PR jobs where possible
- only after duplicate work is removed, evaluate workspace caching / sccache

Do not optimize for preserving every current PR lane. Optimize for the shortest useful failure signal. Exhaustive validation still belongs on `main`; it does not need to block every intermediate commit.

#168 fixes CI coverage for stacked PRs. This issue is about making the CI that now runs substantially cheaper and faster.

gosunuts:
Closes #171

## Summary

- Keep the PR gate limited to `fmt`, pure-Rust `clippy`, workspace tests, and
  `deny`.
- Exclude the kernel-default binary, consensus, and node packages from the PR
  workspace pass; check their pure-Rust feature surfaces separately.
- Move full-node, benchmark, kernel-parity, and portable validation to
  `main.yml`, triggered only by `main` pushes or manual dispatch.
- Remove duplicated clippy work from `kernel-parity`.
- Preserve stacked pull-request coverage and PR cancellation semantics.

## Validation

- Python/PyYAML workflow parse passed.
- `cargo fmt --all -- --check` passed.
- PR clippy commands passed in WSL without a C++ kernel build.
- PR test commands passed in WSL for workspace, consensus, node, binary, and
  RPC targets.
- `git diff --check` passed.
- `cargo-deny` was unavailable locally because it is not installed in the WSL
  environment.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,20 +23,19 @@ concurrency:
   group: ci-${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.ref || github.run_id }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
+# Fast PR gate: fmt, pure-Rust clippy/tests, and dependency checks. Full-node,
+# kernel, benchmark, and portable validation run in `main.yml` on main only.
 env:
   CARGO_TERM_COLOR: never
   RUST_BACKTRACE: 1
   CARGO_INCREMENTAL: 0
-  # Full-node production feature set: all four storage backends plus the
-  # bitcoinkernel script-verification engine (the kernel is the default build).
-  FULL_NODE_FEATURES: "rocksdb,fjall,redb,mdbx,kernel"
 
 jobs:
   fmt:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@1.95.0
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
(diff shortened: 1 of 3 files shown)


gosunuts (commented):
Overall this is a good implementation of #171. The split between a fast PR gate and deeper `main` validation is much cleaner, and the current PR run confirms the intended feedback loop is now substantially cheaper.

One thing I would fix before merging: the new `.github/workflows/main.yml` should set explicit token permissions, e.g.:

```yaml
permissions:
  contents: read
```

The workflow only needs repository contents, and CodeQL is already flagging the missing permissions block. This is worth addressing rather than leaving the workflow on repository-default `GITHUB_TOKEN` permissions.

One non-blocking point to make explicit: PR clippy no longer uses `--all-targets`, so some test/example/bench-target lint coverage is intentionally deferred to the deeper `main` lanes. That trade-off is consistent with #171, but it would be good to document that this is deliberate.

Aside from that, the direction looks right: this removes duplicated/native-heavy PR work instead of adding more CI lanes, while retaining exhaustive validation after merge.

PR #261, opened 2026-09-03
feat: add compact live UTXO view to ScriptIndex

### LINKED ISSUE #225: Add a compact live UTXO view and modes to ScriptIndex

gosunuts:
Closes #225

## Summary
- Add the compact `ScriptLive` column family with empty-valued `script-hash-prefix || full-outpoint` rows.
- Track `TxLookup`, `ScriptHistory`, and `ScriptLive` watermarks independently; implement `scriptindex=utxo` and `scriptindex=full` modes.
- Anchor Live deletes and rollback restores to authoritative undo scripts, handle same-block cancellation/BIP30 replacement, and match UTXO admission rules.
- Resolve Live locators against one stable authoritative UTXO view with exact script checks and rebuild/stamp the view safely after restoration.
- Add Redb support and contract/regression coverage.

## Validation
- `cargo test -p bitcoin-rs-index --tests` (67 passed)
- `cargo test -p bitcoin-rs-index --no-default-features --features redb --tests` (68 passed)
- `cargo test -p bitcoin-rs-utxo --lib` (58 passed)
- Redb storage contract/prefix tests (16 passed)
- Linux production `cargo check` for Fjall and Redb node builds

The branch also includes the latest `origin/main` merge.


metaphorics:
@greptileai

metaphorics:
@sourcery-ai review

metaphorics:

metaphorics:

metaphorics:

metaphorics:

metaphorics:

metaphorics:

metaphorics:

PR #274, opened 2026-09-04
index/storage: implement ScriptLive durable row format (#226)

### LINKED ISSUE #226: Audit and optimize ScriptIndex on-disk representation

metaphorics:
## Summary

Implements the durable `ScriptLive` row format selected by the #226 audit (Q5) so that #225 can materialize the live-output view without revisiting the on-disk layout.

The locator is `prefix(8) || txid(32, LE) || vout(4, LE)` with an empty value:
- `ScriptLiveRow` in `crates/index/src/types.rs` owns the encoding/decoding.
- `ColumnFamily::ScriptLive` is appended with discriminant `10` (additive per `docs/policies/db-migration.md` 3.1).
- `crates/storage/src/redb_impl.rs` wires the fixed-width `TXINDEX_SCRIPT_LIVE` table through `RedbTxIndexStore` (get/snapshot/prefix/scan/write/condition validation) and maps the variable-width `script_live` table for `RedbStore`.
- `NodeConfig` validation comments/error are updated to reflect that #226 Q5 has chosen the locator; `utxo` mode remains disabled until #225 wires the worker that populates the rows.

Tests: `script_live_row_roundtrips_outpoint` in `types.rs` and `txindex_script_live_roundtrips_with_empty_value` in `redb_txindex.rs` cover the key layout, outpoint roundtrip, empty-value invariant, and fixed-width rejection.

Closes #226
Unblocks #225

Link to Devin session: https://app.devin.ai/sessions/055d1325582141c58965f0399e417f0f
Open in Devin Desktop: https://app.devin.ai/desktop/session/055d1325582141c58965f0399e417f0f?variant=devin
Requested by: @metaphorics

## Summary by Sourcery

Implement the durable ScriptLive storage format and integrate it across indexing backends to unblock live-output materialization.

New Features:
- Add the durable ScriptLive row format for locating currently unspent outputs by script prefix and full outpoint.
- Register ScriptLive as a new storage column family and support it across the redb store and transaction-index store APIs.

Bug Fixes:
- Enforce ScriptLive fixed-width keys and empty-value invariants consistently across direct and batched writes.

Enhancements:
- Add round-trip, validation, and backend-equivalence coverage for ScriptLive storage behavior.

Documentation:
- Update configuration validation messaging to reflect the selected ScriptLive locator while keeping utxo mode disabled until row population is wired.

Tests:
- Cover ScriptLive key encoding, outpoint recovery, empty values, invalid writes, and backend behavior.


metaphorics:
@codex review

metaphorics:
@codex review

metaphorics:
@sourcery-ai review

metaphorics:
@greptileai

metaphorics:

metaphorics:

metaphorics:
qodo fix merge conflicts

metaphorics:

metaphorics:

PR #452, opened 2026-09-04
Store ScriptIndex heights big-endian (format 5)

metaphorics:
Closes the Q3 decision in #226: hash-prefix row heights are big-endian so a prefix scan is chronological. This is a breaking derived-index format change. Previous ScriptIndex bytes are disposable.

Restacked onto current `main` (including #502) so clippy and workspace tests inherit the combo-descriptor / `generateblock` fixes and restored policy docs.

## Chosen layout

```
TxConfirmed / Funding / Spending
  key   = prefix(8) || height_be(4)
  value = TxPosition[n]              # 8 bytes each, unchanged
ScriptLive
  key   = script-prefix(8) || txid(32) || vout_le(4)
  value = empty                      # unchanged; no height suffix
```

`spending_prefix` still uses electrs' big-endian wrapping mix. Heights are no longer little-endian, so the mixed-endian key is gone.

## Reset policy

Store format is now **5**. Opening format 3 or 4 (little-endian heights) deletes `TxLookup` and `ScriptHistory` and leaves `ScriptLive` (`IDX-04`). Foreign versions still refuse start; the worker full-resets those namespaces.

No dual-read of old keys. No in-place row rewrite.

## API order

Resolvers still sort by numeric height after exact-resolve so pagination order does not depend on raw iteration, including when an 8-byte prefix collision interleaves another script's rows.

Proof: `crates/index/tests/be_order.rs`, `format_3_open_resets_historical_keys_and_keeps_live`, `format_4_open_resets_historical_keys_and_keeps_live`.

## Follow-ups (stacked)

1. This PR — sortable keys + format 5.
2. Unify the leftover `index:format_version` row-value marker with the store format and drop `IndexFormat::Legacy` dual-read.
3. Record the rest of the #226 chosen format (TxPosition width, positioned Spending, Live locator) as the durable contract, not a keep-LE audit.

Stacked on `main`. Reindex of History/TxLookup is the supported transition.

[...]

PR #503, opened 2026-09-04
Compact ScriptIndex to store format 5

### LINKED ISSUE #226: Audit and optimize ScriptIndex on-disk representation

metaphorics:
Implements the #226 chosen ScriptIndex layout. Derived index bytes are disposable: opening format 3 or 4 resets every capability and rebuilds from chain/UTXO. There is no dual-read, dual-write, or in-place conversion.

## Chosen layout

```text
TxConfirmed / Funding / Spending
  key   = prefix(8) || height_be(4)
  value = packed TxPosition[n]       # 6 bytes each (u24 offset + u24 length)

BlockHeaders
  key   = sha256d(header)(32)
  value = empty

ScriptLive
  key   = prefix(8) || txid(32) || vout_u24(3)
  value = empty
```

Typical Funding/TxConfirmed/Spending row is **18 bytes** (was 20). Header identity is **32 bytes** (was 80). Live locator is **43 bytes** (was 44) and stays injective on `(script_prefix, outpoint)` so point deletes remain identity-safe.

## Why this mix

- **Packed positions (Q1):** consensus-serialized blocks cannot exceed 4 MiB, so the old `u32` pair wasted two bytes on every occurrence. 6-byte entries stay zerocopy. Delta/varint cannot beat that at the common `n = 1` row; merge/batch admission uses `TX_POSITION_SIZE`, so the smaller value admits more positions under the same `max_bytes` cap.
- **Spending positions (Q2):** empty values force a full-block scan per funding output and cannot disambiguate 8-byte outpoint-prefix collisions. Positions stay.
- **Big-endian height (Q3):** LE keys made lexicographic order disagree with numeric height, scattered IBD writes, and defeated LSM prefix compression. Resolvers still sort by numeric height so API pagination does not depend on raw keys.
- **Hash header keys (Q4):** the 80-byte header is identity-bearing, but storing it as the key pays 80 bytes of LSM key overhead per block. Rollback already has the header in hand to re-hash.
- **Compact Live locator (Q5):** `vout` packs to 3 bytes under the same block-size cap. Full txid stays in the key so a point delete cannot remove another script's row. Prefix-range scans stay read-only.

Rejected: keep 8-byte `TxPosition`; empty Spending values; keep LE “because electrs”; 44-byte Live locator; hashed/shorter-txid locators (not injective).

## Reset granularity

Format 5 is a store-wide epoch because it changes shared prefix-row encoding, packed positions, header keys, and the Live locator together. A later Live-only change can reset only `ScriptLive`; a packed-position change still has to reset TxLookup and ScriptHistory together.

Foreign store versions that are not 3, 4, or 5 still refuse start.

## Stacked follow-up

- #524 drops the leftover `index:format_version` dual marker and `IndexFormat::Legacy`.

## Proof

- `crates/index/tests/le_order.rs` (BE key order matches numeric; resolvers still sort)
- `iter_funding_rows_height_order_matches_numeric`
- `format_4_open_resets_all_derived_capabilities`
- packed-position / Live-injectivity / header-hash unit tests
- redb dedicated txindex widths: 12 / 32 / 43

Physical fjall tables in `docs/benchmarks/storage-footprint.md` are still the format-4 snapshot. Re-run `crates/storage/examples/storage_footprint.rs` before treating those numbers as current.

## Relation to #225

Q5 is the only storage-format blocker for Live rows. Mode parsing, capability/readiness, and spend anchors can proceed independently. Live rebuilds from one UTXO pass; History is a reindex.

Does not close #226’s full measurement campaign (pinned multi-era corpus, fjall sealed-table accounting). It lands the format that campaign selected so #225 is not blocked on an intermediate layout.

<div><a href="https://cursor.com/agents/bc-209932d8-c946-4202-b763-0127f348d23d?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-209932d8-c946-4202-b763-0127f348d23d&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


metaphorics (commented):
Held on a format-record design fork that conflict resolution cannot settle.

**The fork.** Current main publishes the live index format in the ASCII record `index:format_version` (`INDEX_FORMAT_VERSION_KEY`, value 2) — the reset-completion batch conditionally re-puts it (`resume_capability_reset`, index.rs ~826: "Resume of a claim that predates the acquire-batch marker still publishes the current row-value format"). This PR's format-5 model instead makes the binary key `[0x00, b'V']` (`FORMAT_VERSION_VALUE = [0x05,…]`) the live record and deletes the ASCII key as legacy leftover (adoption paths, index.rs:731/1769).

**Why no merge resolution is coherent.** Taking the PR's completion-batch delete breaks `reset_claim_carries_mask_epoch_and_base_version` and `union_growth_preserves_claim_identity_and_deletes_full_union_state` (they assert main's `deletes == 0` publication semantics). Restoring main's conditional put (with v5 bytes) fixes those two but breaks `format_4_open_resets_all_derived_capabilities` (:660) and `format_stays_current_after_reset_and_rebuild` (:1878) — because the PR's open/adoption then deletes the ASCII record my put just published. 63/65 tests pass in either direction; the same two fail each way from opposite ends.

**Decision needed:** which record owns the format after format 5 —
1. ASCII record stays live (main's design): renumber the PR's binary-key writes onto the ASCII record and drop the adoption-path deletes; or
2. Binary key becomes live (this PR's design): main's conditional completion put must be removed, and the two main-side tests updated to expect the legacy-cleanup delete — plus every other main code path reading the ASCII record (`stored_format`, `IndexWriter::open`) migrated in the same change.

The same ScriptIndex-format family is touched by open #462, #470, #527, #592, #463 — whichever design wins should be stated once for the family. Branch head is conflict-free, compiles green, and passes 63/65 index_roundtrip tests; the merge from current main is already on the branch.

metaphorics:
HOLD: **COMPETING**. Absorb-check 2026-09-12 (net content diff `e7dacf1a..07928bb7`, path-restricted to `crates/index`/`crates/storage` plus touched docs, checked against current `origin/main` `c5a6440c`). Every substantive hunk is live; there is nothing to close on.

Live hunks (substance absent from main):

- `crates/index/src/index.rs:1-28`: the head keeps the ~3.3k-line monolith importing `encode_height`/`header_identity_key` from `crate::types`, while main's same path is a 28-line module facade (`mod block/…/format/…/write`, 496d8833). Whole-file collision.
- `crates/index/src/index.rs:160`: `FORMAT_VERSION_VALUE = [0x05, 0x00, 0x00, 0x00]` vs main `crates/index/src/index/format.rs:82`, `INDEX_FORMAT_VERSION: u32 = 2`. Main has no format-5 compaction.
- `crates/index/src/types.rs`: `encode_height`/`decode_height` big-endian suffixes vs main `crates/index/src/types.rs:48` (`height.to_le_bytes()`; zero grep hits for `encode_height` on main); `HEADER_KEY_SIZE = 32` plus `header_identity_key` (double-SHA256 `BlockHeaders` keys) absent from main; `SCRIPT_LIVE_ROW_SIZE = HASH_PREFIX_LEN + 32 + 3` vs main `crates/index/src/types.rs:224` (`+ 4`); 6-byte packed `TxPosition`/u24 fields (`U24_MAX`) vs main `crates/index/src/types.rs:301` (8-byte `offset[4] + len[4]`).
- `crates/storage/src/redb_impl.rs`: `txindex_v2_block_headers` as `FixedTable<32>` and `txindex_v2_script_live` as `FixedTable<43>` vs main's 80-byte/44-byte v1 layout; the `column_families.rs` u24 live-key docs are likewise unlanded.

Absorbed hunks: none substantive. The `CONCEPTS.md` and `docs/contracts/indexing.md` hunks are format-5 big-endian prose that main does not carry either.

Author question: this compaction line and the #452 BE-keys line are rival format-5 designs, and main has since landed its own format-2 modular baseline (496d8833). #226 closed 2026-09-11 19:13Z without a decision comment, so no recorded authority picks the canonical line. This also overlaps the standing gatekeeper hold on #452 (`crates/index/src/index.rs` collides in both directions). No merge, no enqueue until the author decision.

metaphorics:
Superseded in the compact direction by #979, which ports this PR's format-5 delta (BE heights, u24 compact rows, marker 5) onto main's split index modules with 119 green tests, clippy-clean. Leaving this PR itself open for the maintainer to close on #979's merge. Reopen the discussion there if the direction is wrong.

metaphorics:
Superseded by #979, which ports this PR's compact format-5 delta onto main's modular index (119 tests green, clippy/fmt clean, independent review clean bill). The monolith this PR edits no longer exists on main. Closing as superseded; reopen if the direction is wrong.

metaphorics:
Superseded by #979, now landed as 1a125f24 (compact format 5 on the modular index). Closing.

PR #673, opened 2026-09-09
fix(rpc): make Esplora fees and CORS wallet-safe

### LINKED ISSUE #669: Esplora API: /fee-estimates returns 1.0000000000000002 instead of 1.0, so wallets overpay 10% at the fee floor
`GET /api/fee-estimates` returns values like `1.0000000000000002` where the answer is exactly `1.0`. The extra ULP is small, but wallets that round a fee rate *up* turn it into a visibly wrong fee.

## Reproduce

```console
$ curl -s https://<node>/api/fee-estimates | jq '{"3","6","10"}'
{
  "3": 1.0000000000000002,
  "6": 1.0000000000000002,
  "10": 1.0000000000000002
}
```

Every target sitting at the relay minimum is affected; only target `1` (a real estimate, `4.966`) looks clean.

## Cause

`crates/rpc/src/esplora/public.rs:524`

```rust
.map(|v| v * 100_000_000.0 / 1000.0)
```

`estimatesmartfee` gives BTC/kvB, so this converts to sat/vB in two steps, and the intermediate is not representable:

```python
>>> 0.00001 * 100_000_000.0      # BTC/kvB -> sat/kvB
1000.0000000000001               # already wrong
>>> 0.00001 * 100_000_000.0 / 1000.0
1.0000000000000002
>>> 0.00001 * 100_000.0          # same conversion, one step
1.0
```

The two constants are a single factor of `100_000`. Doing it as one multiply is exact for this value.

## Why it matters

A wallet cannot send a fractional-ULP fee rate, so it rounds. Rounding *up* is the safe direction — it is what you want when the estimate is the minimum you must beat — and that is exactly where this bites:

```python
>>> math.ceil(1.0000000000000002 * 10) / 10
1.1
>>> math.ceil(1.0 * 10) / 10
1.0
```

A wallet asking for the 6-block rate against this node builds at **1.1 sat/vB instead of 1.0 — a 10% overpay at the floor**, on every transaction, for as long as the network sits at the minimum. Against mempool.space and blockstream.info, which return clean values, the same wallet builds at 1.0.

## Suggested fix

```rust
.map(|v| v * 100_000.0)
```

Rounding the result to a fixed precision (Esplora publishes ~3 decimals) would also make the output independent of how the conversion is spelled.

## Environment

- Node behind `bitcoin-rs-explorer.gosunuts.xyz`, mainnet, tip 966132
- Observed 2026-09-09

### LINKED ISSUE #670: Esplora API: no CORS headers and no OPTIONS route, so browser-based wallets cannot use it
The Esplora surface sends no `Access-Control-Allow-Origin`, so a browser cannot read any response from it. Every browser-based wallet is therefore locked out, even though the API itself is complete and correct.

## Reproduce

Same page, same moment, two endpoints:

```js
await fetch("https://<bitcoin-rs-node>/api/blocks/tip/height")
// TypeError: Load failed

await fetch("https://mempool.space/api/blocks/tip/height")
// 200 -> "966132"
```

Headers:

```console
$ curl -sI https://<bitcoin-rs-node>/api/blocks/tip/height -H "Origin: https://wallet.example"
HTTP/1.1 200 OK
Content-Type: text/plain
# no access-control-* header

$ curl -sI https://mempool.space/api/blocks/tip/height -H "Origin: https://wallet.example"
access-control-allow-origin: *

$ curl -sI https://blockstream.info/api/blocks/tip/height -H "Origin: https://wallet.example"
Access-Control-Allow-Origin: *
Access-Control-Expose-Headers: x-total-results
```

Both reference implementations of this API send the header; this one does not.

## Preflight

`OPTIONS` is not routed either — it falls through to the JSON-RPC handler:

```console
$ curl -s -X OPTIONS https://<bitcoin-rs-node>/api/blocks/tip/height \
    -H "Origin: https://wallet.example" -H "Access-Control-Request-Method: GET"
{"result":null,"error":{"code":-32600,"message":"invalid request: malformed http request"},"id":null}
```

A JSON-RPC error is not a valid preflight response, so any request that needs one fails before it is sent. Returning a JSON-RPC envelope from a REST path looks like a routing gap independent of CORS.

I could not find any `Access-Control` or `OPTIONS` handling under `crates/rpc/src/`.

## Why it matters

This is the difference between "usable by wallets" and "not". A wallet running its core as WASM in a webview — which covers browser wallets and the webview-based desktop/mobile shells — calls this API with `fetch`, and the browser discards the response before the wallet sees it. The endpoint looks broken while being entirely healthy, and there is no error the wallet can report beyond a generic network failure, because the browser deliberately withholds the reason.

A native shell can work around it by routing requests through its own HTTP stack. A browser tab cannot.

## Suggested fix

Emit on the Esplora responses:

- `Access-Control-Allow-Origin: *` (public chain data; same choice as Esplora and mempool)
- `Access-Control-Expose-Headers: X-Total-Results` so paging is readable
- answer `OPTIONS` with `204` and the matching `Allow-Methods` / `Allow-Headers`

Worth applying to error responses too, not just `2xx` — otherwise a rejected broadcast reaches the wallet as an opaque failure instead of `transaction decode failed`.

If the intent is that a reverse proxy owns this, saying so in `docs/api` would save the next person the same investigation.

## Environment

- Node behind `bitcoin-rs-explorer.gosunuts.xyz`, mainnet, tip 966132
- Observed 2026-09-09, WebKit

gosunuts:
## Summary

- convert BTC/kvB fee estimates to sat/vB in one multiplication, preserving the exact 1.0 relay floor
- classify listener paths as JSON-RPC, Core REST, public Esplora, or backend Esplora surfaces with explicit CORS policies
- apply wildcard CORS consistently to every public /api success and error response while keeping /esplora, JSON-RPC, and Core REST disabled
- terminate /api preflight requests with HTTP 204 and reject every other OPTIONS request before JSON-RPC dispatch
- document the wallet-facing CORS contract and add regression coverage

## Verification

- cargo fmt --all -- --check
- cargo clippy -p bitcoin-rs-rpc --lib --tests --target x86_64-unknown-linux-gnu -- -D warnings

Closes #669
Closes #670


gosunuts (commented):
The fee conversion fix looks good, and adding CORS to the wallet-facing Esplora surface is the right direction.

I would suggest keeping the CORS design at the HTTP surface/policy layer rather than growing Esplora-specific response writers in `server.rs`.

Ideally the flow would be:

- classify the HTTP surface first (`JsonRpc`, `CoreRest`, `EsploraPublic`, `EsploraBackend`)
- each surface owns its auth / allowed-method / CORS policy
- `OPTIONS` terminates at the router instead of falling through normal request dispatch
- use one response writer that applies the selected surface policy to success and error responses alike

For `/api`, wildcard CORS is definitely part of the wallet-facing contract: wallets need UTXO/history/fee/broadcast access from browser/webview environments, and error responses must be readable too.

For `/esplora`, I would make the CORS policy explicit rather than inheriting it automatically from the public Esplora path, since `/internal/*` and `/block-template` are backend-facing. It can still be public CORS if that is intentional; the important part is that it is a deliberate property of `EsploraBackend`.

One more small routing point: now that `OPTIONS` is accepted by `read_request()`, non-Esplora `OPTIONS` requests should probably terminate as `404`/`405` instead of being able to reach JSON-RPC.

PR #740, opened 2026-09-10
fix(qa): publish golden fixtures atomically and reuse complete caches

metaphorics:
## Problem

`scripts/fetch-golden.sh` writes directly to cache destinations. A failed raw download or txid pipeline can leave a partial/empty file which the next run treats as complete. A fully populated cache also performs 15 unnecessary block-hash lookups and cannot be reused offline.

## Changes

- Stage downloads on the destination filesystem and rename each artifact only after the command/pipeline succeeds.
- Reject malformed block-hash responses, empty raw responses, and malformed/empty/non-hex txid arrays before publication.
- Treat zero-byte cache entries as missing, preserve existing nonempty artifacts, and skip all HTTP requests for complete cache entries.
- Create staging lazily and clean it on exit/HUP/INT/TERM; complete caches need no staging write.
- Emit one terminating newline for txid files, rather than an additional blank line.
- Add eight offline subprocess-level regression tests with failure injection and a small, read-only-permissions CI workflow to run them.

Publication is atomic per artifact, not across the block/txid pair. Existing nonempty cached artifacts remain trusted; this is not a consensus validator or a retrospective integrity scan.

## Verified locally

Source copy was checked against original Git blob `6073db3294ec4ce88571ca55a9dad7e18b2cd7b5` before edits.

- Baseline: seven test methods produced 12 assertion failures, including subtests.
- A second review found an eager staging-write regression; an eighth test reproduced it, and lazy staging fixed it.
- Final: `python3 -m unittest discover -s scripts/tests -v` — **8 passed**, using system Python 3.13 and an injected offline curl stub.
- `bash -n scripts/fetch-golden.sh` — passed.
- `git diff --cached --check` — passed.
- Same seeded complete cache, successful mock transport: **15 HTTP requests before, 0 after**. This is a request-count result, not a production wall-clock benchmark.

## Scope and remaining validation

One commit, three related files, based on main `d493b5dc24a1d3d10a9158caec2739e9f2741996`. No Rust source, consensus behavior, dependency lockfile, existing workflow, or operator data is changed. Scratch work and evidence logs are outside the PR.

Draft pending final-head CI/review. This environment has no Rust toolchain and terminal DNS is unavailable, so no Cargo, live Blockstream, full-repository, formal-model, or CodeRabbit success is claimed. No force-push, direct main update, merge, or auto-merge is requested.

---
## Summary by cubic
Makes `scripts/fetch-golden.sh` publish fixture downloads atomically and reuse complete caches so a failed fetch can no longer leave a partial file that later runs treat as valid. Previously each artifact was written directly to its destination and complete caches still performed block-hash lookups for every height.

- Stages downloads in a temporary directory and renames them into place only after validation succeeds.
- Treats zero-byte cache entries as missing, preserves existing nonempty artifacts, and skips all HTTP requests for complete entries.
- Rejects malformed block-hash responses, empty raw responses, and malformed or empty txid arrays before publication.
- Adds eight offline subprocess regression tests, with fixture heights derived from the script, plus a read-only CI workflow to run them.
- Publication is atomic per artifact, not across the block/txid pair, and existing nonempty cached artifacts remain trusted.

<sup>Written for commit e78753b9eedebb6989d984334634ea0ad6a811fa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gosuda/bitcoin-rs/pull/740?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/fixture-tools.yml
+++ b/.github/workflows/fixture-tools.yml
@@ -1,0 +1,24 @@
+name: fixture-tools
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: fixture-tools-${{ github.event_name == 'pull_request' && github.ref || github.run_id }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  golden-fixtures:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
+      - name: Check downloader syntax
+        run: bash -n scripts/fetch-golden.sh
+      - name: Run offline downloader regressions
+        run: python3 -m unittest discover -s scripts/tests -v
--- a/scripts/fetch-golden.sh
+++ b/scripts/fetch-golden.sh
@@ -5,20 +5,54 @@ heights=(0 1 170 91722 91812 91842 91880 173818 363731 481823 481824 624455 7096
 out_dir="crates/primitives/tests/testdata"
 mkdir -p "${out_dir}"
 
+# Stage on the destination filesystem so a failed transfer cannot become a cache hit.
+tmp_dir=""
+trap 'if [[ -n "${tmp_dir}" ]]; then rm -rf -- "${tmp_dir}"; fi' EXIT
+trap 'exit 129' HUP
+trap 'exit 130' INT
+trap 'exit 143' TERM
+
 for height in "${heights[@]}"; do
-  hash="$(curl -fsSL "https://blockstream.info/api/block-height/${height}")"
   bin_path="${out_dir}/${height}.bin"
   txids_path="${out_dir}/${height}.txids.txt"
-
-  if [[ ! -f "${bin_path}" ]]; then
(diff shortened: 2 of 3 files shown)


metaphorics:
Follow-up `e78753b9eedebb6989d984334634ea0ad6a811fa` addresses the two published test-maintenance findings:

- Fixture heights now come from the downloader's single owned array, with a fail-closed parse. First-artifact paths also derive from that inventory instead of assuming height 0.
- The test module links the acquisition contract/failure cases here, cites the immutable original script blob, and distinguishes synthetic transport-byte expectations from Bitcoin validity vectors.

No downloader behavior changed in this follow-up. Re-ran both local offline suites: **24 tests passed** (8 downloader + 16 importer tests from independent #747), and staged whitespace checks passed. The returned remote test blob `5e2fee90b7baa64a48809c87e28d16e5a03b9d43` matches the locally tested file.

The intervening branch update and ready-for-review state were preserved. Full Rust/CI validation and independent review remain unverified; earlier-head cancellation/queued runs are not treated as passing checks.

PR #744, opened 2026-09-10
fix(node): restore admission after coherent reorg failures

### LINKED ISSUE #678: Reorg path leaves the mempool generation odd on connect failure, re-wedging the node (#618 defect class outside the window path)
# Reorg path leaves the mempool generation odd on connect failure, re-wedging the node (#618 defect class outside the window path)

## Summary

PR #661 teaches `BlockSync::apply_window_followed` to finish the chain transition on clean window-apply failures (restoring the gateway's even generation so sync can retry) while keeping `UtxoCommit` failures fail-closed. The reorg path has the same drop-without-finish shape and a live retry consumer, so a clean reorg failure can still wedge the node exactly like #618: every later `begin_transition` is refused while the generation is odd, and the refusal surfaces as `ReorgError::Unavailable` / `"block apply rejected because clean shutdown has begun"`.

## Evidence

- `crates/node/src/reorg.rs:388-395` — the `G5` comment states it outright: *"An error during execution leaves the generation odd by design — admission stays closed."* `switch_to_branch` begins the transition (`begin_transition_locked`), runs `execute_streamed_plan`, and on error drops the `ChainTransition` without `finish()`.
- `crates/node/src/sync.rs:744-828` — `BlockSync::switch_branch_if_outweighed` calls `switch_to_branch` from the block-drain path on every tick and its `ConnectFailed` arm treats the failure as retryable (purges invalidated descendants, waits for re-fetch). The next tick calls `switch_to_branch` again; with the generation odd, `begin_transition_locked` fails and the reorg — and every normal window apply through `begin_transition` — is refused until restart.
- Field precedent: issue #618's post-#657 field report documents that an odd generation refuses all applies with the shutdown text and no further block-sync activity.

## Scope

- Mirror the `settle_window_failure` contract from #661 into `switch_to_branch`'s error handling: finish the transition for clean refusals whose failing block was refused before the UTXO commit-of-record; keep `UtxoCommit`-sourced failures (and `Fatal`/torn markers, which already close admission permanently) fail-closed. Copying the finish unconditionally would be wrong — a torn per-shard UTXO commit must not be retried.
- This requires `execute_streamed_plan` to distinguish `ApplyError::UtxoCommit` from other connect errors (today `ReorgError::ConnectFailed` does not carry that distinction).
- Regression test: reorg connect fails with a pre-UTXO error → generation returns to even → a retry of the same reorg (or a normal apply) succeeds without restart.

## Relations

- Follow-up to #618 / PR #661 (window path fixed there; this is the remaining same-defect-class site).
- `Chainstate::apply_block` / `apply_window` convenience entry points also drop on refusal (by design per the #661 docs — callers that need retry must finish explicitly); their production callers have no live retry loop today, so they are out of scope here.

lidarbtc:
A transient connect refusal during a reorg could leave the mempool generation odd after already disconnecting blocks. Every subsequent transition was then refused until restart. Reorg now explicitly finishes a coherent committed prefix so the same branch can be retried, while possibly torn UTXO or disconnect state remains fenced.

One reorg settlement path serves sync branch switching and RPC invalidation. It uses the existing typed `ConnectFailed.source`, keeps transaction reconsideration under the active odd generation, and skips reconsideration and reorg checkpoint publication for unsafe outcomes. RPC planning/preflight errors also reach settlement. A failed generation finish preserves the original error/progress, closes apply admission, and requests shutdown instead of returning success or retrying.

## Acceptance criteria

- [x] A fail-once block-body persistence refusal after a committed disconnect reopens admission; retry reaches the same target without restart.
- [x] Permanent rejection, partial connect progress, streamed disconnect body loss, and clean disconnect refusal preserve their committed prefix and permit appropriate later progress.
- [x] UTXO commit and fatal/stuck-marker outcomes remain fenced. The UTXO boundary test preserves real checkpoint debt and includes a positive publication control.
- [x] Finish failures after success and after a clean refusal are reported as fatal settlement failures; the original cause and progress survive.
- [x] RPC invalidation can recover from a missing-body preflight refusal without stranding generation.
- [x] Existing contracts and fault seams supply the regression coverage; no storage format, validation default, mempool policy, or direct Chainstate convenience-entry-point behavior changes.

## Validation

- Native node library: **550 passed**; kernel node library: **553 passed**.
- Journal: **6 passed in each profile**. Crash recovery: **2 passed in each profile**; one helper remains intentionally ignored and is launched by the SIGKILL parent test.
- Strict Clippy for all node targets passes with both `fjall` and `fjall,kernel`.
- Ownership, reference-set and evidence checks: **17 passed**.
- Rust 1.95.0 MSRV compilation, changed-file formatting and `git diff --check` pass.

The repository-wide formatter reports existing differences in three untouched files. `CONSTRAINTS.md` CL-18 names `overhaul_streaming_reorg`, which does not exist on the base commit; its prescribed invocation is unavailable, not a passing gate. Existing streaming reorg unit tests and both recovery suites above passed. This PR makes no full-release or power-loss durability claim.

Closes #678.

### DIFF
--- a/crates/node/src/apply.rs
+++ b/crates/node/src/apply.rs
@@ -9375,7 +9375,7 @@ mod consensus_rule_tests {
             vec![coinbase_transaction(1)],
         )?;
         let raw = bytes::Bytes::from(consensus_bytes(&block));
-        let applied = apply_block_with_serialized(&handles, &block, raw)?;
+        let applied = apply_block_with_serialized(&handles, &block, raw.clone())?;
         bodies
             .bodies
             .write()
@@ -9402,6 +9402,31 @@ mod consensus_rule_tests {
         assert_eq!(handles.chain_tip.load_full(), header_tip_before);
         assert_eq!(handles.applied_tip.load_full(), applied_tip_before);
         assert_eq!(utxo.len(), utxo_len_before);
+        // MPL-04: a read-only preflight refusal must release its generation.
+        assert!(
+            handles.mempool_gateway.stable_generation().is_some(),
+            "missing disconnect data must not leave admission closed"
+        );
+
+        bodies
+            .bodies
+            .write()
+            .insert((applied.height, applied.hash), raw.to_vec());
+        crate::reorg::invalidate_block(
+            &handles,
+            &crate::chain_effects::ChainFollowers::noop(),
+            applied.hash,
+        )?;
+        assert_eq!(
+            handles.applied_tip.load_full().map(|tip| tip.hash),
+            Some(genesis_hash),
+            "restoring the body must allow invalidation without restarting"
+        );
+        assert_eq!(
+            handles.block_tree.read().node(applied.tip_id)?.status,
+            NodeStatus::Invalid
+        );
+        assert!(handles.mempool_gateway.stable_generation().is_some());
         Ok(())
     }
 
@@ -9891,11 +9916,7 @@ mod consensus_rule_tests {
                 vec![coinbase_transaction(seed)],
             )?;
             let raw = bytes::Bytes::from(consensus_bytes(&block));
-            let tip = apply_block_with_serialized(&handles, &block, raw.clone())?;
-            bodies
-                .bodies
-                .write()
-                .insert((tip.height, tip.hash), raw.to_vec());
+            let tip = apply_block_with_serialized(&handles, &block, raw)?;
             old_tips.push(tip);
             prev = block.block_hash();
         }
@@ -9961,6 +9982,31 @@ mod consensus_rule_tests {
             Some(12),
             "applied tip must be at the height reached by the completed window"
         );
+        // MPL-04: the completed disconnect prefix is a stable chain, even
+        // when the next window cannot load its bodies.
+        assert_eq!(utxo.len(), 12);
+        assert!(
+            handles.mempool_gateway.stable_generation().is_some(),
+            "mid-rollback body loss must reopen admission at the committed prefix"
+        );
+        bodies
+            .fail_on_second_read
+            .write()
+            .remove(&(target_tip.height, target_tip.hash));
+        crate::reorg::switch_to_branch(
+            &handles,
+            &crate::chain_effects::ChainFollowers::noop(),
+            fork_target,
+            |_| None,
+            |_| {},
+        )?;
+        assert_eq!(
+            handles.applied_tip.load_full().map(|tip| tip.hash),
+            Some(Hash256::from(fork_prev)),
+            "retry must resume from the committed prefix and reach the fork tip"
+        );
+        assert_eq!(utxo.len(), 21);
+        assert!(handles.mempool_gateway.stable_generation().is_some());
         Ok(())
     }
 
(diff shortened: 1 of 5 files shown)


lidarbtc:
Execution evidence for `bdd290ae50fd661e01614343f214f251d4c73202` (base `c84a44ef`).

The initial draft regression was executed before the production change and failed at `admission must reopen after the clean connect refusal` (0 passed, 1 failed). The final implementation passes that regression, including the exact `BlockBodyPersistence` source assertion and successful retry to the same target.

| Check | Executed result |
| --- | --- |
| Native node library, fjall | 550 passed |
| Kernel node library, fjall + kernel | 553 passed |
| Journal integration | 6 passed in each profile |
| Crash recovery integration | 2 passed in each profile; one intentionally ignored subprocess helper is launched by the passing SIGKILL parent test |
| Native node Clippy, all targets, `-D warnings` | Passed |
| Kernel node Clippy, all targets, `-D warnings` | Passed |
| `overhaul_ownership` | 5 passed |
| `overhaul_reference_set` | 7 passed |
| `overhaul_evidence` | 5 passed |
| Rust 1.95.0 node/kernel library check | Passed |
| Changed-file rustfmt and `git diff --check` | Passed |

The reorg boundary coverage includes clean retry after committed disconnect, recovery from a permanently rejected branch, coherent partial connect progress, streamed body loss after eight committed disconnects, clean coinstats refusal, RPC preflight refusal, fatal marker fencing, UTXO commit fencing without checkpoint publication, and generation-finish failure after both success and refusal. The UTXO test uses a real checkpoint publisher plus a positive control proving the synthetic debt could otherwise have been cleared. It injects the typed error at settlement; it does not claim to simulate every physical torn-write pattern.

Commands (one runner, four build jobs and four test threads):

```sh
cargo test --locked --offline -p bitcoin-rs-node --no-default-features --features fjall --lib --test chainstate_journal --test crash_recovery -- --test-threads=4
cargo test --locked --offline -p bitcoin-rs-node --no-default-features --features fjall,kernel --lib --test chainstate_journal --test crash_recovery -- --test-threads=4
cargo clippy --locked --offline -p bitcoin-rs-node --no-default-features --features fjall --all-targets -- -D warnings
cargo clippy --locked --offline -p bitcoin-rs-node --no-default-features --features fjall,kernel --all-targets -- -D warnings
cargo test --locked --offline -p bitcoin-rs --no-default-features --features fjall --test overhaul_ownership --test overhaul_reference_set --test overhaul_evidence -- --test-threads=4
cargo +1.95.0 check --locked --offline -p bitcoin-rs-node --no-default-features --features fjall,kernel --lib
rustfmt --edition 2024 --check crates/node/src/reorg.rs crates/node/src/sync.rs crates/node/src/apply.rs
git diff --check
```

Stable checks used Rust 1.98.1 on x86_64 Linux. Cargo.lock SHA-256: `cf02002b43698b3318c4a3967ee59c27494adc1b292f46502a8723a3c59acf57`. The kernel profile uses bitcoinkernel 0.2.1 / libbitcoinkernel-sys 0.3.0. Missing local Boost headers were restored by extracting the distribution `libboost1.90-dev=1.90.0-6ubuntu1` package into a temporary prefix (`CMAKE_PREFIX_PATH=/tmp/bitcoin-rs-678-build-deps/usr`); no system installation or repository dependency change. Package SHA-256: `7b89698c907fd5d33ccd439674fa53803139923822181c87b910579827aca379`. The failed build-script fingerprint was invalidated and the archive rebuilt, rather than substituting an old archive.

Limits and baseline conditions:

- The first sandboxed native suite hit local socket `EPERM` in eight tests; rerunning with loopback permission passed. The final native and kernel runs above include all node library tests.
- `cargo fmt --all -- --check` reports pre-existing formatting in `bin/bitcoin-rs/tests/gates/g20_formal_models.rs`, `crates/utxo/src/set/persistent.rs`, and `crates/utxo/tests/overhaul_persistent_coins.rs`. None is modified here; changed files pass rustfmt.
- The prescribed CL-18 command, `cargo test --locked --offline -p bitcoin-rs-node --no-default-features --features fjall --test overhaul_streaming_reorg -- --nocapture`, exits 101 because that target does not exist on the base. This is unavailable evidence, not a passing CL-18 gate; current streaming coverage in `apply.rs` passed in both profiles.
- This records local execution, not a completed GitHub CI run or full release promotion. No performance or new power-loss durability claim is made.

The draft-only review findings are covered by the production settlement fix, exact error-source assertion, fail-closed tests, and explicit contract paths. The PR remains draft as requested.

PR #784, opened 2026-09-10
fix(fuzz): reject invalid disk probes and keep provenance owner-derived

metaphorics:
## Scope

Independent follow-up to #762, which merged during the deeper review. One commit based on main `ec43331f871d9457d26f1b21799d1bb6048977c4`, changing only `scripts/import-qa-assets.sh` and its existing provenance-flow test module. This does not depend on #783.

## Reproduced defects

1. A successful `df` command with a nonnumeric or out-of-range available-space field causes both shell `-lt` checks to error inside a conditional. The script treats those errors as false comparisons, logs `disk ok`, and continues through mapping, minimization, and provenance replacement. Reject missing, nondecimal, negative, and unrepresentable available-space values before the disk guard. Preserve real command/pipeline failure status explicitly.
2. The mapper derives script selectors from Rust-owned `FLAGS`, but the generated provenance still claims fixed `0x00` and `0x03`. Reordering the owner inventory makes the record false even when the imported frames are correct. Remove the duplicated numbers from shell documentation and generated provenance; describe `NONE` and `TAPROOT` through their owning `FLAGS` entries.

No new limits, target defaults, dependencies, Rust source, live corpus files, historical provenance data, or main-branch writes.

## Verification

- Original hash-verified importer suite: **47 passed**.
- Three added full-shell regressions against original code: **3 failed**, reproducing the two disk-result cases and the stale selector claims.
- This PR independently, with the merged mapper unchanged: `python3 -m unittest discover -s scripts/tests -p 'test_import_qa_assets*.py' -v` — **50 passed**.
- This PR combined locally with #783 — **59 passed**; the change sets have disjoint files and neither requires the other.
- `bash -n scripts/import-qa-assets.sh`, Python compilation, and staged whitespace checks passed.
- Remote blobs match tested files: script `9f2b6b83e40e2041e7cb3c2c54e05d3bfa56529c`, tests `60adf32392a04c632e777c5b5382f52b01e7ecd8`.

Tests use disposable directories and injected external tools; the real shell and mapper run, but no actual clone, Cargo build, or minimization was performed. The three new regressions remain in the existing QAC-03-linked test suite, which now has 15 methods.

## Limits / gates

The disk check remains a preflight free-space observation, not a reservation or a whole-import disk budget guarantee. Per-file publication is not a transaction spanning corpus and provenance and does not establish power-loss durability.

The locked CL-14 Cargo gate and `cargo clippy --locked --workspace --all-targets -- -D warnings` were attempted locally and returned **127: cargo not found**. No gate pass or performance gain is claimed. Draft pending supported CI and independent review.

---
## Summary by cubic
Rejects invalid disk probe results in the QA asset import script so a failed or malformed `df` call cannot be mistaken for enough free space. Also makes provenance and harness documentation describe selectors by their owning FLAGS entry instead of hardcoded `0x00` and `0x03` values, so the record stays accurate after flag reordering.

- Disk probe now aborts on missing, nonnumeric, negative, or overflowing available-space values.
- Provenance, the shell documentation, and the `script_eval` harness comment now reference the FLAGS entries `NONE` and `TAPROOT` instead of fixed selector bytes.
- Added regression tests for invalid disk probes and for docs not regressing to fixed selector claims.

<sup>Written for commit 19618c040412bcb9ae1be4a5ee961a47e697520e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gosuda/bitcoin-rs/pull/784?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/fuzz/fuzz_targets/script_eval.rs
+++ b/fuzz/fuzz_targets/script_eval.rs
@@ -27,9 +27,9 @@ use bitcoin_rs_script::{Interpreter, VerifyFlags};
 /// ```
 ///
 /// Scripts from the qa-assets corpus are wrapped into this framing by
-/// `scripts/import-qa-assets.sh`; seeds written by that script use selector
-/// `0x00` (NONE) for raw scripts and `0x03` (TAPROOT) for the wrapped P2TR
-/// variant. Keep those indices in step with `FLAGS` below.
+/// `scripts/import-qa-assets.sh`; seeds written by that script use the harness
+/// `FLAGS` entry `NONE` for raw scripts and, for files >= 32 bytes, a P2TR
+/// variant using its `TAPROOT` entry.
 const FLAGS: [VerifyFlags; 6] = [
     VerifyFlags::NONE,
     VerifyFlags::MANDATORY,
--- a/scripts/import-qa-assets.sh
+++ b/scripts/import-qa-assets.sh
@@ -11,8 +11,8 @@
 #   tx_decode     <- bitcoin_deserialize_transaction (raw consensus bytes, direct)
 #   script_eval   <- bitcoin_deserialize_script + bitcoin_script_bytes_to_asm_fmt
 #                    (raw script bytes, wrapped into the script_eval framing:
-#                    selector 0x00 = NONE, and for files >= 32 bytes a P2TR
-#                    variant with selector 0x03 = TAPROOT)
+#                    the harness FLAGS entry NONE, and for files >= 32 bytes
+#                    a P2TR variant using its TAPROOT entry)
 #
 # Disk discipline (repo AGENTS.md): the worst-case footprint of the clone is
 # declared below (shallow clone ~= corpus size, assumed <= 2 GiB); free space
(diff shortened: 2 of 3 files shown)


metaphorics:
## Deeper contract audit — remaining mismatch and verified limit

A separate, pre-existing standards defect remains in `CONSTRAINTS.md#qa-corpus-importer-setup-contract`: it documents injected test sentinels as unconditional tool exit codes (`git rev-parse` must exit 19, etc.). The implementation actually propagates the failing command's status, consistent with QAC-03 and the provenance test module. A disposable full-shell probe made the Git stub return 42; the importer returned **42**, did not start minimization, and preserved previous provenance. The register's fixed-number wording is therefore not an accurate implementation contract. This PR does not silently rewrite that register or claim it is correct; it needs owner-consistent documentation correction rather than remapping real command failures to test constants.

Also tested the documented atomicity boundary: with an existing direct seed `old seed`, a mapping followed by injected cmin failure returned **43**, left that seed updated to `Q`, and preserved the prior provenance file. This is the declared **per-file** guarantee, not an all-or-nothing import transaction. Consumers must not infer current-corpus identity from an unchanged provenance file after a failed import.

These observations are separate from the 50 passing tests for this PR and the 59 passing tests with #783. Neither local passes nor a future green lightweight workflow constitute the unavailable Cargo/resource-gate evidence.

PR #809, opened 2026-09-10
perf(rpc): bound mempool page-selection scratch and clone only winners

metaphorics:
## Scope

Atomic follow-up to merged #796: one commit, one source file, based directly on `main` at `5070f377ecee293b9c7739d4a59d2bba47e10543`. Preserve the concurrently merged API-09 cursor contract and tests; do not rewrite the earlier PR or its branch.

Replace capture-of-every-eligible-Arc with batched selection over borrowed entries. This closes the O(R) scratch-space limitation documented in #796, where R is the suffix after the cursor.

## Algorithm and bounds

For positive requested page size K, maintain at most 2K live borrowed entries. Each full batch uses nth-element selection, keeps K entries, and records the first excluded `(time, txid)` key. Because pool txids are unique, entries at or beyond that cutoff cannot beat the retained page. Each full batch discards K entries, so total batch-selection work is linear in visited entries. A final partial batch is trimmed before the winning transaction Arcs are cloned.

- Selection scratch: O(min(R, K)), rather than O(R), apart from ordinary Vec capacity rounding. The **2K bound is on live borrowed entries, not exact allocator bytes**.
- Snapshot Arc clones: min(R, K), rather than R; no Arc clone/drop churn for losing entries.
- Selection plus final ordering: O(N + P log P), with N visited pool entries and P returned entries. The full pool traversal remains.
- No request-sized up-front reservation: Vec grows from actual observed entries. `len / 2 >= K` avoids overflow from computing `2 * K`; zero returns before polling the iterator, and final nth selection has an explicit `K < len` guard.
- Cursor lookup, selection, and final Arc capture share one read guard, so no second lookup can observe removal/replacement between selection and capture. Final ordering and transaction projection remain outside the guard.

**Tradeoff:** partition work now runs under the existing read guard. Bounded scratch and fewer atomic reference operations do not prove better writer latency. Read-lock contention and nearly-full/unbounded requests still require native measurement. This is not an unconditional speedup claim or default-promotion result. No new cache/index, persistence, consensus, dependency, public API, or unsafe code.

## Regression and benchmark coverage

Keep all four existing API-09 pagination regressions. Add three tests: ascending/reverse traversal across repeated batch boundaries (257 entries, tied times, cursors and extreme limits); borrowed-payload/no-selection-Arc-clone behavior; zero-limit no-iterator-poll behavior.

Add an ignored paired release microbenchmark comparing the prior #796 capture/partition algorithm against the new selector. It checks equality before timing, keeps setup untimed, alternates execution order across six trials, and covers 10,000 entries with K=1/25/1000/9999/unbounded, with and without a cursor. It measures page selection/Arc capture, **not HTTP, live locks, RSS, or end-to-end node throughput**. No timing assertions in normal tests.

```
cargo +1.95.0 test --locked --release -p bitcoin-rs-rpc --lib benchmark_bounded_page_selection -- --ignored --nocapture
```

## Executed preparation evidence

- Exact merged source blob reconstructed and verified: `a74279986c63c0487cc9f2c0ac87c21cad890591`.
- Published candidate blob matches the locally reviewed source: `09991b399a471d4a5ad4edb4e2bf2f0f6cf05c54`.
- Independent Python full-sort/reference vs batched-selection model: **275,490 comparisons, no mismatches**. Every comparison also checks the live 2K bound and batch-discard work bound. Exhaustive permutations through six entries, randomized snapshots through 16,384 entries, malformed/noncanonical/missing cursors, and size boundaries.
- Five deliberately faulty controls detected, including an incorrectly tightened cutoff that drops a later winning entry.
- Whitespace checks and patch reverse-application checks passed. GitHub compare confirms one commit / one changed file against current main; no force-push or main write.

The model is a separately written Python algorithm, **not execution of the Rust selector or std's implementation**. It does not validate actual allocation behavior, locking, HTTP, or compiler semantics.

## Pending native acceptance — keep draft

Cargo/rustc are absent in the preparation environment; repository cloning also fails DNS resolution. Explicit attempts at pinned RPC tests, Clippy `--all-targets -- -D warnings`, rustfmt, CL-07 `overhaul_core_api`, CL-14 `overhaul_resource_bounds`, and the release microbenchmark all failed at command launch (`[Errno 2] No such file or directory: 'cargo'`). These are BLOCKED, not passes or code-test failures. AGENTS.md's pre-publication Clippy requirement remains unsatisfied and is disclosed here rather than suppressed.

Before ready/merge: compile and run the new and existing Rust tests, fix all changed-line Clippy/rustfmt issues, run the owning gates and real concurrent public-route tests, and collect paired native CPU/allocation/RSS/read-lock/writer-latency evidence. The inherited unlimited-response behavior is unchanged; this does not close the whole API-08/CL-14 resource-budget programme.

Library reference: https://doc.rust-lang.org/std/primitive.slice.html#method.select_nth_unstable_by_key . The API contract supports the algorithm argument; served documentation is not a substitute for the pinned compiler.

---
## Summary by cubic
Bounds Esplora API-09 mempool pagination scratch to at most 2K live borrowed entries instead of retaining the full eligible suffix, and clones `Arc` payloads only for returned transactions. Cursor lookup, selection, and capture share one read guard; ordering and cursor semantics stay unchanged, while selection work moves inside the guard. Adds `CONSTRAINTS.md` entries SEL-01 and SEL-02 documenting the borrowed-payload and zero-limit contracts, and links the selector tests to them.

**Validation**

- Adds coverage for repeated batch boundaries, reverse traversal, tied timestamps, borrowed payloads, and zero-limit requests.
- Adds an ignored paired release benchmark for bounded and previous page-selection algorithms across several page sizes and cursor states.
- Pinned Rust tests, Clippy, rustfmt, and the benchmark remain blocked because `cargo` and `rustc` are unavailable.

<sup>Written for commit 4a71ef4d2f37ab2da9d7e577ced451ba131937f3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gosuda/bitcoin-rs/pull/809?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CONSTRAINTS.md
+++ b/CONSTRAINTS.md
@@ -11,6 +11,17 @@ Every measurement cell below is `UNMEASURED`. Target values are contracts, not
 results. Unknown is not false; a missing measurement blocks the owning task.
 `BLOCKED` is a workflow state, not a verdict.
 
+## Mempool page-selection contracts (v1)
+
+These contracts own the selector's bounded-work and ownership expectations;
+benchmark results remain subject to CL-19 and CL-20 and must not be inferred
+from the unit tests.
+
+| ID | requirement |
+| --- | --- |
+| SEL-01 | Page selection borrows candidate payloads and clones payloads only for the final returned page. |
+| SEL-02 | A page with limit zero returns immediately without polling the pool iterator. |
+
 ## Formal model tool identity
 
 | Field | Value |
--- a/crates/rpc/src/esplora/backend.rs
+++ b/crates/rpc/src/esplora/backend.rs
@@ -435,6 +435,7 @@ mod pagination_tests {
         }
     }
 
+    // Contract: CONSTRAINTS.md, "Mempool page-selection contracts (v1)", SEL-01.
     #[test]
     fn selection_borrows_payloads_and_only_the_final_page_is_cloned() {
         let entries: Vec<_> = (0_u32..129)
@@ -472,6 +473,7 @@ mod pagination_tests {
         assert!(entries.iter().all(|entry| Arc::strong_count(&entry.tx) == 1));
     }
 
+    // Contract: CONSTRAINTS.md, "Mempool page-selection contracts (v1)", SEL-02.
     #[test]
     fn zero_selection_does_not_poll_the_pool_iterator() {
         let entries = std::iter::from_fn(|| -> Option<&MempoolEntry> {
--- a/crates/rpc/src/esplora/backend.rs
+++ b/crates/rpc/src/esplora/backend.rs
@@ -6,6 +6,7 @@
 use alloc::sync::Arc;
 use core::str::FromStr as _;
 
+use bitcoin_rs_mempool::MempoolEntry;
 use bitcoin_rs_primitives::{OutPoint, Txid};
 
 use super::http::{bad, dispatch_error, json_response, query_limit};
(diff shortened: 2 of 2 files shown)


metaphorics:
## Additional independent selector checks and consolidation

Re-read the published selector rather than publishing the competing `perf/mempool-page-bounded-scratch` experiment as a duplicate PR. This PR's observed-entry vector growth and overflow-safe batch boundary are preserved. The selector body at `4a71ef4d2f37ab2da9d7e577ced451ba131937f3` (backend blob `f426c4ed8a8f5f22d16bea64901476f6af9dde0b`) matches the algorithm modeled from original blob `09991b399a471d4a5ad4edb4e2bf2f0f6cf05c54`; concurrent contract/test references were not overwritten.

An additional Python full-sort-versus-batched-selection run passed **275,136 comparisons**, checking page equality, the live 2K reference bound, and a linear bound on total partition-input work. It includes repeated batches, tied keys, cursor/limit boundaries, ascending/descending/random traversals, and 100,000-entry cases. This is a separate run from the 275,490-comparison evidence already in the description, not a replacement count.

For 100,000 modeled entries with K=25, both tested large traversal orders peaked at 50 live borrowed references and returned 25 entries. Native allocator capacity, allocated bytes, RSS and Arc operations were not instrumented: this is algorithm/resource-model evidence only.

Two new deliberately faulty selector controls were detected. The first negative-control fixture initially failed to expose an incorrectly tightened cutoff; it was replaced with traversal times `[0,10,20,30,40,50,60,70,80,90,5]`, K=5, where the late `5` must displace `40`. The corrected fixture rejects the wrong cutoff. No production-model page mismatch was found. The original four independent model controls also still reject their mutations.

The Python checks do **not** execute the Rust code, std's partition implementation, locks or HTTP routes. Native tests/Clippy/rustfmt and the paired release benchmark remain unrun here because Cargo/rustc are absent. The checked head's CI is queued. Partitioning under the pool read guard still needs actual read-hold and writer-latency measurements; nearly-full and unlimited requests need regression measurements. No measured speedup, full resource-budget closure or zero-findings verdict is claimed. The externally changed ready state does not supply that missing evidence.

PR #811, opened 2026-09-10
perf(mempool): reuse witness-free txids and computed transaction weight

metaphorics:
## Scope

One independent atomic commit, one file (`crates/mempool/src/entry.rs`), based on main `5070f377ecee293b9c7739d4a59d2bba47e10543`. No dependency on the separate bounded-page follow-up #809. No changes to the active encoder rewrite in #522.

## Remove repeated work in entry construction

1. Compute txid as before. When `Tx::has_witness()` is false, reuse those digest bytes for wtxid instead of serializing and double-hashing the same transaction again. Transactions with any witness stack still call the existing `Tx::wtxid()` implementation. A stack containing an empty item counts as witness; this distinction is tested.
2. Compute `Tx::weight()` once, retain it, and derive BIP141 vsize with the existing rounded-up division and u32 clamp. Calling `tx.vsize()` and then `tx.weight()` previously recomputed the same weight.

On the inspected primitives implementation, explicit counting-writer transaction traversals in the constructor fall from **five to three**. Witness-free construction also removes one complete transaction-ID double-hash computation. These are source-level work counts, **not measured CPU or throughput percentages**. LLVM code generation, the additional witness-presence check, and end-to-end benefit remain unmeasured.

Keep `Tx::weight()` as the weight owner: no second weight formula, new persistent cache, mutable metadata, unsafe code, dependency, public API, schema, default, or consensus-validation change. Policy-adjusted vsize, fee/ancestor/descendant accounting, payload Arc identity, and sigop handling are unchanged.

## Independent references and new tests

BIP141, Specification / Transaction ID and Additional definitions / Transaction size calculations: https://bips.dev/141/ . BIP144, Serialization and Hashes: https://bips.dev/144/ .

Add three native regression functions:

- Seven independent wire vectors covering witness-free, an empty witness item, witness only in the second input, every weight-rounding residue, a 253-byte witness item, and a 253-item witness stack. Check txid, wtxid, size, weight, vsize and payload identity.
- Policy sizes 0/1/999/u32::MAX remain separate from BIP141 vsize, with fee and ancestor/descendant fields preserved.
- Existing raw zero-input/zero-output construction semantics remain intact; this is not claimed to be valid transaction admission.

Golden values come from a separate Python `struct`/`hashlib` BIP wire encoder, not the candidate or baseline Rust methods. All seven share stripped size 103 and txid `9e3f408ee773b8b6473f7cc02cbdef80644f916cee3981ee0704e279b7cdc6dc`. Full sizes are 103/108/109/110/111/365/362; weights are 412/417/418/419/420/674/671.

Reproduce the golden vectors independently:

```python
import hashlib, struct

def compact(n):
    if n < 253: return bytes([n])
    if n <= 0xffff: return b'\xfd' + struct.pack('<H', n)
    if n <= 0xffffffff: return b'\xfe' + struct.pack('<I', n)
    return b'\xff' + struct.pack('<Q', n)

def script(b): return compact(len(b)) + b

def stack(items):
    return compact(len(items)) + b''.join(script(x) for x in items)

def digest(b):
    return hashlib.sha256(hashlib.sha256(b).digest()).digest()[::-1].hex()

vin = (b'\x11' * 32 + struct.pack('<I', 0) + script(b'\x51')
       + struct.pack('<I', 0xfffffffe)
       + b'\x22' * 32 + struct.pack('<I', 7) + script(b'')
       + struct.pack('<I', 0xffffffff))
body = b'\x02' + vin + b'\x01' + struct.pack('<Q', 1000) + script(b'\x51')
prefix, suffix = struct.pack('<i', 2), struct.pack('<I', 9)
base = prefix + body + suffix
cases = [([], []), ([b''], []), ([], [b'\xaa']),
         ([], [b'\xaa' * 2]), ([], [b'\xaa' * 3]),
         ([b'', b'\xbb' * 253], [b'']), ([], [b''] * 253)]
for witnesses in cases:
    full = (prefix + b'\x00\x01' + body
            + b''.join(stack(x) for x in witnesses) + suffix
            if any(witnesses) else base)
    weight = 3 * len(base) + len(full)
    print(digest(base), digest(full), len(full), weight, (weight + 3) // 4)
```

## Executed preparation evidence

- Reconstructed current-main source verified against Git blob `b8631f813b8d201f5d1cd7c74d500f6d3804b0a2`.
- Published candidate blob exactly matches local reviewed content: `697de05e3e583b6e85e7caa1c68bc56501322e53`.
- The independent encoder above was executed and generated the seven checked-in tuples. Eighteen separate arithmetic boundary cases checked rounded vsize equivalence with saturation/clamping, including u64 limits.
- Whitespace and reverse-patch checks passed; GitHub comparison confirms one commit and only this file changed. No force-push or main mutation.

**Not executed:** the Rust tests, native constructor benchmarks, rustfmt, Clippy, full owner gates or workspace tests. The preparation environment has no Cargo/rustc. Explicit pinned test and Clippy command attempts fail at launch with `[Errno 2] No such file or directory: 'cargo'`; this is BLOCKED, not a pass. AGENTS.md's required pre-publication Clippy proof remains unsatisfied and is disclosed rather than waived.

## Before ready/merge

```
cargo +1.95.0 test --locked -p bitcoin-rs-mempool --lib wire_metadata_tests
cargo +1.95.0 test --locked --release -p bitcoin-rs-mempool --lib
cargo +1.95.0 clippy --locked -p bitcoin-rs-mempool --all-targets -- -D warnings
cargo +1.95.0 fmt --all -- --check
```

Run the applicable constraint/owner gates and paired baseline/candidate constructor measurements on witness-free, mixed-witness and large-input/large-witness workloads. Record pinned compiler, CPU, input identity, repeated timing and allocations before asserting a measured performance gain. Keep this PR draft until native acceptance evidence exists.

---
## Summary by cubic
Reduces repeated work in `MempoolEntry` construction by reusing the txid as the wtxid for witness-free transactions and computing transaction weight once instead of twice. Adds a paired benchmark example comparing the new constructor against the original.

**Refactors**
- Witness-free txs skip the second transaction-ID double-hash; any witness stack (even an empty item) still uses `Tx::wtxid()`.
- Centralizes BIP141 vsize derivation in `Tx::vsize_from_weight`, shared by `Tx::vsize` and `MempoolEntry::new`.
- Documents the raw zero-input entry contract in `CONSTRAINTS.md`.
- Adds seven BIP141/BIP144 wire vectors, policy-size preservation checks, and raw empty-entry coverage with golden values from an independent Python encoder.
- Adds `entry_construction_bench`, a release-only paired sample of the new and original constructors across witness-free and mixed-witness fixtures.

<sup>Written for commit 9a835b45f547da2fdc3d522762207740c645aa87. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gosuda/bitcoin-rs/pull/811?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CONSTRAINTS.md
+++ b/CONSTRAINTS.md
@@ -127,3 +127,11 @@ probe failure exits with status 7, and an insufficient-space check exits with
 status 1. Every setup failure removes the temporary staging directory and does
 not attempt the clone. `scripts/tests/test_import_qa_assets.py`
 `SetupFailureTests` is the regression suite for this contract.
+
+## Raw zero-input mempool entry contract
+
+`MempoolEntry::new` accepts a raw default transaction for compatibility with
+callers that construct entries before admission. For `Tx::default()` it must
+preserve the metadata values `(size, weight, bip141_vsize) = (10, 40, 10)`,
+use the txid as the wtxid, and produce a zero fee rate. The regression test is
+`crates/mempool/src/entry.rs::raw_empty_entry_keeps_zero_input_behavior`.
--- a/crates/mempool/examples/entry_construction_bench.rs
+++ b/crates/mempool/examples/entry_construction_bench.rs
@@ -1,0 +1,127 @@
+//! Paired `MempoolEntry` construction benchmark against main `ff965032`.
+//!
+//! Run with `cargo run --locked --release -p bitcoin-rs-mempool
+//! --example entry_construction_bench`. Fixtures exercise serialization cost,
+//! not admission validity. Samples are not a performance acceptance gate.
+#![allow(clippy::expect_used)]
+
+use std::sync::Arc;
+
+use bitcoin_rs_mempool::MempoolEntry;
+use bitcoin_rs_primitives::{OutPoint, Tx, TxIn, TxOut};
+use bitcoin_rs_script::count_tx_legacy;
+
+fn fixture(inputs: usize, script_len: usize, witness_len: Option<usize>) -> Tx {
+    let mut tx = Tx {
+        version: 2,
+        inputs: vec![
+            TxIn {
+                previous_output: OutPoint::default(),
+                script_sig: vec![0x51; script_len],
+                sequence: 0xffff_fffd,
+                witness: Vec::new(),
+            };
+            inputs
+        ],
+        outputs: vec![TxOut {
+            value: 1_000,
+            script_pubkey: vec![0x51; script_len],
+        }],
+        lock_time: 42,
+    };
+    if let Some(length) = witness_len {
+        // Witness only on the last input also exercises mixed transactions.
+        let input = tx.inputs.last_mut().expect("witness fixture input");
+        input.witness = vec![vec![0x55; length]];
+    }
+    tx
+}
+
+// Exact original constructor, benchmark-only; no production compatibility path.
+fn baseline_entry(tx: Arc<Tx>, vsize: u32, fee: u64, time: u64, height: u32) -> MempoolEntry {
+    let own_size = u64::from(vsize);
+    let txid = tx.txid();
+    let wtxid = tx.wtxid();
+    let bip141_vsize = u32::try_from(tx.vsize()).unwrap_or(u32::MAX);
+    let size = u32::try_from(tx.total_size()).unwrap_or(u32::MAX);
+    let weight = tx.weight();
+    let sigop_cost = count_tx_legacy(&tx);
+    MempoolEntry {
+        tx,
+        txid,
+        wtxid,
+        vsize,
+        bip141_vsize,
+        size,
+        weight,
+        sigop_cost,
+        fee,
+        fee_rate: fee_rate(fee, own_size),
+        fee_delta: 0,
+        ancestor_size: own_size,
+        ancestor_fee: fee,
+        ancestor_fee_delta: 0,
+        descendant_size: own_size,
+        descendant_fee: fee,
+        descendant_fee_delta: 0,
+        time,
+        height,
+    }
+}
+
+const fn fee_rate(fee: u64, vsize: u64) -> u64 {
+    if vsize == 0 {
+        return 0;
+    }
+    fee.saturating_mul(1_000) / vsize
+}
+
+fn main() {
+    use std::hint::black_box;
+    use std::time::Instant;
+
+    assert!(!cfg!(debug_assertions), "run this benchmark with --release");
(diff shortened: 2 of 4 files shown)


metaphorics:
## Paired constructor benchmark added; native acceptance still unverified

Pushed `5a5fe8a5ff886a67b61f3a17f181dae08163b9b7`, adding only `crates/mempool/examples/entry_construction_bench.rs`. The benchmark file remains unchanged at subsequent head `7426aebeb5d3baf5612ea9c5db3b38f2f6c3383e`: read-back blob `0180ddeacdff5f6b8ca4f152c6f727578914c9ad` matches the local prepared source. Existing metadata code, golden tests, and the concurrent `Tx::vsize_from_weight`/contract changes were preserved, not overwritten. The original description's one-commit/one-file count describes the earlier head, not the now-expanded PR.

The executable compares the original `ff965032` constructor against production `MempoolEntry::new`, with 1/8/128 inputs and absent or last-input witness. Each of the six fixture shapes checks txid, wtxid, size, weight and BIP141 vsize equality before timing. It warms both arms, alternates order across seven pairs, uses black_box, and emits raw nanoseconds and iteration counts. Fixtures exercise construction cost, not admission validity.

```
cargo +1.95.0 run --locked --release -p bitcoin-rs-mempool --example entry_construction_bench
cargo +1.95.0 clippy --locked -p bitcoin-rs-mempool --all-targets -- -D warnings
```

Additional executed evidence: all seven previously checked-in wire-vector tuples were independently reproduced with Python struct/hashlib. Local whitespace checks passed. These are not native Rust tests. Both commands above, and pinned rustfmt, fail at launch here because `cargo` is absent; no constructor timing samples or Clippy pass exist from this environment. Same-binary microbenchmark samples also do not replace allocation/RSS measurements or independently built baseline/candidate end-to-end comparisons.

At the checked head, CI remains queued/pending. The PR was marked ready outside this work; that status change is not native acceptance evidence, and no merge or readiness transition was performed here. Compiler, owner-gate, and measured performance acceptance remain outstanding.

PR #818, opened 2026-09-10
perf(consensus): remove block-fact reserialization and redundant witness hashing

metaphorics:
## Status

Draft optimization candidate. Rust compilation, Clippy and matched performance evidence are still required. **No measured speedup, full-kernel equivalence, or readiness-for-merge claim.** Follow-up to the now-merged #798; its witness-validation fixes and review changes are preserved.

## Work removed

`BlockFacts::from_parsed` currently reconstructs every transaction's stripped serialization into one growable scratch buffer, re-encodes script CompactSize lengths, and then walks the input/output metadata again to compute stripped size.

- Hash legacy transactions directly from their borrowed transaction span.
- For SegWit, stream exactly three borrowed ranges into SHA256: version, canonical input/output bytes, and lock time. Marker/flag and witness bytes are excluded without field reconstruction.
- Compute stripped size from the same ranges in constant metadata work per transaction. Hashing itself remains linear in bytes; no cryptographic work is skipped.
- Delete `txid_from_spans`, `base_size_from_spans`, `extend_span`, `push_compact`, and `span_len`, together with the per-derivation transaction-sized scratch Vec.
- On first lazy witness-ID fill, reuse cached txids for witness-free transactions instead of serializing and double-hashing them again. Iterate over the full transaction slice, not a truncating zip; missing caller-supplied identity rows retain the old hashing fallback.

This does **not** make all BlockFacts construction allocation-free: identity vectors, layout spans, and the existing Merkle implementation remain unchanged. No parser behavior, consensus rule, public API, unsafe code, dependency, lockfile, feature default, cryptographic backend, storage format, or durability change.

## Comparator structure

Two commits deliberately separate the controls from the implementation:

1. `c8b084f8b530e2f98aa6116dcc4a282857fad76f` adds tests and benchmarks while retaining the original production blob `f97d1fb6a578a65b42d1296bc7d5d1aa1afd0dd2`. This is a source control, **not a measured baseline**.
2. `d79b9abf3f5a8f269d159176ec791a292ccd727b` changes only `crates/consensus/src/block_view.rs`; the harness is byte-identical to the parent.

The new Criterion target has 19 cases: eight fixture shapes measured as derivation-only and parse-plus-derivation, plus three first-fill lazy-witness-ID cases. Shapes include small legacy, 1,024-transaction legacy/SegWit/mixed blocks, 65,536-byte scripts, and 253-input/output boundaries. Fixtures are wire/identity workloads, not signed consensus-valid blocks. Construction, independent oracle assertions and lazy-cache reset are outside the relevant timed loops.

Run the identical command in separate worktrees/target directories for those two revisions:

```sh
cargo bench --locked -p bitcoin-rs-consensus --no-default-features --bench block_facts
```

Retain raw results for at least three alternating matched runs and apply CL-19/CL-20, including stability, host-noise and whole-apply/replay non-regression requirements. The microbenchmark is not a substitute for product workloads or RSS/high-water measurements.

## Correctness coverage added

Eleven Rust regression functions use independent rust-bitcoin decoding, txid/wtxid hashing, weight and Merkle reduction. They cover SHA256 and CompactSize boundaries; nonzero transaction origins; empty outputs/final scripts; witness-only mutation; a single empty witness item; trailing-message exclusion; lazy-cache identities and reuse; empty blocks; real duplicate-tail mutation versus padding; noncanonical-length rejection; and version/lock-time inclusion.

References:
- BIP141 transaction IDs and block weight: https://bips.dev/141/
- Existing parser owner: `crates/primitives/src/layout.rs`, blob `5065f10271480d0cafb0411e5da956f5f456e9ea`.
- SHA2 streaming API: https://docs.rs/crate/digest/0.11.0

## Executed checks and limits

- Reconstructed input files match their immutable Git blob SHAs; all five uploaded candidate blobs match the locally checked bytes.
- Local `git diff --check` passes. GitHub compare confirms exactly two commits and five intended paths (three added, two modified).
- A supplementary independent Python wire/range model checked 3,084 cases with zero mismatches and five OpenSSL SHA256d cross-checks. **This did not execute the Rust candidate and is not compiler-backed or differential-kernel evidence.**
- Cargo/rustc probes, formatting, native Clippy, native/kernel-feature regressions, native library tests, benchmark compilation and applicable CL-06, CL-08/13/21, CL-14/19, CL-20/23 commands returned **127: executable absent**. Rust tests and benchmarks have therefore not run locally. Terminal cloning also failed DNS resolution; the local verification tree contains only the reconstructed change set, not a complete buildable workspace.

Base: `5070f377ecee293b9c7739d4a59d2bba47e10543`. No force-push or main-branch mutation. Keep draft until compiler-backed gates and the required matched performance evidence exist.

---
## Summary by cubic
Removes redundant block-fact reserialization and witness double-hashing when deriving transaction metadata from parsed blocks.

- Hashes legacy transactions directly from their borrowed byte span.
- For SegWit, streams version, canonical input/output range, and lock time into SHA256 without field reconstruction.
- Computes stripped sizes from the same borrowed ranges instead of walking input/output metadata.
- Reuses cached txids for witness-free transactions on lazy witness-ID fill.
- Deletes the transaction-sized scratch buffer and helper functions for CompactSize re-encoding.
- Adds regression tests using rust-bitcoin as an independent oracle.
- Adds a 19-case Criterion microbenchmark; no measured speedup is claimed, and rustc/Clippy verification is still required.
- Makes `finalize_double_sha256` public in `crates/primitives` to reuse the shared streaming finalization.

<sup>Written for commit b1f472f62fd2cd36e1267d812eae9d95f5c0e641. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gosuda/bitcoin-rs/pull/818?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/consensus/Cargo.toml
+++ b/crates/consensus/Cargo.toml
@@ -40,3 +40,7 @@ sonic-rs.workspace = true
 [[bench]]
 name = "merkle"
 harness = false
+
+[[bench]]
+name = "block_facts"
+harness = false
--- a/crates/consensus/benches/block_facts.rs
+++ b/crates/consensus/benches/block_facts.rs
@@ -1,0 +1,86 @@
+//! Parse-once fact derivation and decoded lazy witness-ID benchmarks.
+//! Run the identical harness at its introduction commit and the optimized
+//! descendant in separate target directories. No historical-chain fixture is
+//! needed. This microbenchmark is not an apply-path or full-replay verdict.
+#![allow(missing_docs)]
+
+#[path = "../tests/support/block_facts_fixture.rs"]
+mod fixtures;
+
+use std::hint::black_box;
+
+use bitcoin_rs_consensus::{BlockView, block_view::BlockFacts};
+use bitcoin_rs_primitives::layout::ParsedBlock;
+use bitcoin_rs_primitives::{Tx, consensus_bytes};
+use criterion::{BatchSize, BenchmarkId, Criterion, Throughput, criterion_group, criterion_main};
+
+fn block_facts(c: &mut Criterion) {
+    let mut group = c.benchmark_group("block_facts");
+    // name, tx count, inputs, outputs, script bytes, witness modulus.
+    // Each fixture isolates a relevant source of copying or metadata work.
+    for (name, txs, inputs, outputs, script, witness) in [
+        ("legacy_small", 1, 1, 2, 25, 0),
+        ("legacy_1024", 1024, 1, 2, 25, 0),
+        ("segwit_1024", 1024, 1, 2, 25, 1),
+        ("mixed_1024", 1024, 1, 2, 25, 2),
+        ("legacy_large_script", 1, 1, 1, 65_536, 0),
+        ("segwit_large_script", 1, 1, 1, 65_536, 1),
+        ("segwit_253_inputs", 8, 253, 2, 0, 1),
+        ("segwit_253_outputs", 8, 1, 253, 0, 1),
+    ] {
+        let block = fixtures::fixture(txs, inputs, outputs, script, witness);
+        let bytes = consensus_bytes(&block);
+        let parsed = ParsedBlock::parse_exact(&bytes)
+            .unwrap_or_else(|error| panic!("benchmark layout: {error}"));
+        let checked = BlockFacts::from_parsed(&parsed);
+        fixtures::assert_oracle(&bytes, &checked);
+        assert!(!checked.merkle_mutated());
+        group.throughput(Throughput::Bytes(
+            u64::try_from(bytes.len()).unwrap_or_else(|error| panic!("fixture size: {error}")),
+        ));
+        group.bench_function(BenchmarkId::new("derive", name), |b| {
+            b.iter(|| black_box(BlockFacts::from_parsed(black_box(&parsed))));
+        });
+        group.bench_function(BenchmarkId::new("parse_and_derive", name), |b| {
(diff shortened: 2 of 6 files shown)


metaphorics:
Verification note for unchanged head `d79b9abf3f5a8f269d159176ec791a292ccd727b`:

All five published Git blobs match the local candidate bytes. Clean patch application, exact-byte comparison, reversal and `git diff --check` passed. The harness-only control remains `c8b084f8b530e2f98aa6116dcc4a282857fad76f`; GitHub compare confirms the optimized descendant changes only `block_view.rs`.

I created this PR as a draft. It has since been marked ready for review outside this session; that state change does not supply compiler or performance evidence. The latest checks still show `ci`, `fixture-tools`, and `fuzz-corpus-tools` queued. Local Cargo/Rust/Clippy remain unavailable, the eleven Rust regressions and nineteen benchmark cases have not run locally, and matched performance/RSS/apply/replay gates remain unverified. The Python range model is supplementary only, not execution of the Rust candidate. No merge or native-default promotion was performed in this session.

PR #453, opened 2026-09-04
ci: own the PR gate in one script

metaphorics:
Closes #171. Third of three stacked PRs. Base is #447.

## Why

The PR gate commands were copied across `ci.yml`, `CONTRIBUTING.md`, and `.pre-commit-config.yaml`, and they had already drifted (local `cargo clippy --workspace` pulls kernel; pre-commit ran the full-node C++ surface). `deny-full` on main repeated the PR deny graph.

## What this PR does

- Add `scripts/ci-pr.sh` as the single owner of `fmt` / `clippy` / `test` / `deny`
- CI `fmt`, `deny`, and `rust` jobs invoke the script (deny installs `cargo-deny`, then runs the script — no second feature list)
- `CONTRIBUTING.md` and pre-commit call the same script (local hooks are now the kernel-free PR gate, not the C++ full-node lane)
- Drop `deny-full` from `main.yml`
- Point the kernel-oracle local command at the tests `main.yml` actually runs

## Cache

After the first two PRs, the PR rust graph compiles once with `Swatinem/rust-cache`. sccache is not added: a second compiler wrapper does not pay for a single remaining compile job, and main's jobs use incompatible feature sets (kernel vs not, stable vs nightly) so a shared sccache would thrash.

## Stack

1. #445 → `main` — cut expensive PR lanes
2. #447 → #445 — one Rust compile graph
3. **This PR** → #447 — script owner, drop duplicated deny

<div><a href="https://cursor.com/agents/bc-88ed6820-8862-48d2-b86e-59b67748566e?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-88ed6820-8862-48d2-b86e-59b67748566e&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #447, opened 2026-09-04
ci: compile the PR rust gate once

metaphorics:
Part of #171. Second of three stacked PRs. Base is #445.

## Why

After #445 the PR gate still compiled the kernel-free graph twice: `clippy` and `test` each spun a runner, restored rust-cache, and built the workspace. That is the remaining duplicate work the issue called out.

## What this PR does

- Merge `clippy` and `test` into one `rust` job (clippy first, then tests)
- Keep `fmt` and `deny` parallel — they do not compile the workspace
- One rust-cache restore per PR

## Required checks

Replace `clippy` and `test` with `rust`. The PR jobs are now `fmt`, `deny`, `rust`.

## Stack

1. #445 → `main` — cut expensive PR lanes
2. **This PR** → #445 — one Rust compile graph
3. #453 → this branch — script owner, drop `deny-full`

<div><a href="https://cursor.com/agents/bc-88ed6820-8862-48d2-b86e-59b67748566e?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-88ed6820-8862-48d2-b86e-59b67748566e&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #445, opened 2026-09-04
ci: cut PR gate to fmt, deny, clippy, and tests

metaphorics:
Part of #171. First of three stacked PRs.

## Why

A recent successful `main` `ci.yml` run spent **~16 minutes** on `bench-smoke` (cmake + libboost + kernel) and extra minutes compiling the full-backend test surface. That is the opposite of a fast edit → push → feedback loop.

## What this PR does

The required PR jobs are now exactly the issue’s gate:

- `fmt`
- `deny` (full graph, metadata only — no C++ toolchain)
- `clippy` (kernel-free, `--all-targets`)
- `test` (kernel-free workspace tests)

Moved off PRs onto `main.yml`:

- `bench-smoke` (already existed on main; the PR copy also compiled kernel)
- `comparator-tests`
- portable full-backend tests (`rocksdb,fjall,redb,mdbx`) and isolated `bitcoin-rs-rpc --no-default-features`

Also sets `permissions: contents: read` on both workflows (the review note from closed #244).

## Required checks

If branch protection still requires `bench-smoke` or `comparator-tests` on pull requests, drop those names. The remaining PR jobs are `fmt`, `deny`, `clippy`, `test`.

## Stack

1. **This PR** → `main` — cut expensive PR lanes
2. #447 → this branch — one Rust compile graph
3. #453 → #447 — script owner, drop duplicated `deny-full`

Does not close #171 until the stack lands.

<div><a href="https://cursor.com/agents/bc-88ed6820-8862-48d2-b86e-59b67748566e?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-88ed6820-8862-48d2-b86e-59b67748566e&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #877, opened 2026-09-10
ci: own the PR gate in one script and one compile graph

### LINKED ISSUE #171: ci: reduce PR validation cost

gosunuts:
## Summary

Consolidates the CI-gate stack **#445 → #447 → #453** (open and conflicted since
2026-09-04) into one PR re-derived against current main, and repairs the
workspace breakage the 2026-09-10 merge storm landed without a green gate.
Closes #171. Supersedes #445, #447, #453.

## Why one PR instead of rebasing the stack

The three stacked PRs went idle on 2026-09-04 with green CI but no reviewer,
while main absorbed ~400 commits. #445 is `mergeable: false / DIRTY` today, and
#447/#453 sit on top of it. Rebasing the chain serializes three conflict
resolutions against a CI surface that itself moved (+279 lines in `ci.yml`
since the stack was authored). Re-deriving the intent against current main is
one review instead of three stale ones.

### Commit 1 — repair the workspace (`fix:`)

Main's `ci` workflow was red on every job except `deny`/`comparator-tests`
(run 34469920669): a hand-edited `Cargo.lock` carried a **`bitcoin-rs-mempool`
self-dependency**, so every `cargo --locked` invocation died before compiling —
which masked everything below it. Fixed, each verified by the gate itself:

- `Cargo.lock`: drop the corrupt self-dependency entry.
- `consensus`: `block_witness_commitment_matches` referenced a `coinbase`
  binding that did not exist in scope — the crate did not compile on main.
- `p2p`: `wire_len` encoded envelopes into `std::io::sink()` via bitcoin's
  `Encodable` (wrong `Write` trait, unconvertible error); now uses the
  `encode::serialize` pattern the same file already uses.
- `node`: vestigial never-looping `loop` in `invalidate_block`; commented-out
  mutation loop in `chainstate_journal::encode_record`; unguarded Vec-write
  `expect`s; unused `serde::Deserialize` import.
- Reference identity: `version_output` pinned a line the real binary does not
  print. The pinned bitcoind 31.1 prints a trailing program name
  (`… version v31.1.0 bitcoind` — observed in run 34469920669's fixture
  provisioning). Grounded the manifest, the docs projection, the fixture
  provenance check, and the test expectation in the observed output, and
  re-audited `RELEASE_CUSTODY_SHA256` (formula verified by reproducing the
  previous digest first). Local provisioning now passes end-to-end.
- Test repairs: `corpus_rejects_stale_fixture_after_reference_changes` loaded
  the corpus with the *unedited* release (`&reference.release` instead of
  `&edited`), so it could never fail; `mempool_writer_source_scan_passes`
  needed the block-apply eviction site (`apply.rs` → `handles.mempool_gateway`)
  registered in `AUTHORIZED_GATEWAY_CALLS` per the scan's own review mechanism.
- clippy 1.98.0 conformance and rustfmt across the storm's files.

### Commit 2 — own the PR gate in one script and one compile graph (`ci:`)

- **`scripts/ci-pr.sh` is the single owner** of `fmt` / `clippy` / `test` /
  `deny` commands. CI, pre-commit, and CONTRIBUTING all invoke it. The
  pre-commit config had already drifted (it still referenced the removed
  `mdbx` backend) and its full-node hooks pulled the C++ kernel into every
  local commit.
- **PR gate = `fmt` + `deny` + `rust`** (8 jobs → 3). `rust` compiles the
  kernel-free graph once: three clippy profiles, then five test profiles.
  Profiles run independently inside the script (accumulate, report all), so
  one failure does not hide later diagnostics — preserving the per-step
  `if:` conditions the old jobs had.
- **Deep lanes move to `main.yml`**: bench smoke (kernel C++, witness
  activation, SegWit-v0 oracle — unioned with the portable bench compiles),
  MSRV, native-script evidence, comparator corpus. `deny-full` is dropped;
  `ci.yml` runs the identical full-graph deny on every main push.
- Both workflows get explicit `permissions: contents: read`.

## Verification

- `./scripts/ci-pr.sh fmt` — pass.
- `./scripts/ci-pr.sh clippy` — 3/3 profiles pass (each fix above was caught
  and re-verified by this gate locally).
- `./scripts/ci-pr.sh test` — all 5 profiles (fixtures provisioned and
  verified against the pinned bitcoind locally, including the version-line
  check this PR repairs).
- `bash scripts/provision-ci-reference-fixtures.sh` — pass (downloads,
  sha256-verify, version probe).

- The formal solver run (g20, up to 1h x 6 models) is **main-only**: the
  workspace test profile now skips `all_model_specs_check_with_apalache`
  exactly like the binary profile already did, and main.yml's unfiltered
  full-node tests own the proof. Without this, every PR spends hours in the
  model checker -- the opposite of #171's shortest-useful-signal goal.

## Reviewer notes

- The blast radius of commit 1 is large in file count but each hunk is
  mechanical; the interesting judgment calls are (a) grounding
  `version_output` in the observed binary output (custody fingerprint
  re-audit) and (b) registering the block-apply mempool eviction in the
  ownership scan's audit table. Both follow the contract's own escape hatch.
- If you would rather keep `msrv`/`native-script`/`comparator-tests` on PRs,
  that is a one-line revert per job — the script ownership and single compile
  graph are independent of lane placement.


metaphorics:
qodo resolve merge conflicts

PR #946, opened 2026-09-11
refactor(node): isolate txindex query work and unify accounting

metaphorics:
## Summary

Continue the txindex cleanup on top of the owners already merged into main (including #941), rather than introducing a competing worker layout.

- Split the 936-line query module into a small shared snapshot gate and private transaction-resolution, script-traversal, block-source, and work-budget modules. Public query entrypoints still share the existing health/revision/watermark/chain-transition checks; no alternate read path is introduced.
- Make `QueryBudget` the single owner of historical and live scan charging. Preserve both truncation diagnostics and every existing row, byte, scan, and body-read limit.
- Replace three repeated pending-batch transfer constructions with `PendingForward::take`, retaining the fence, watermarks, capabilities, durable position, and original flush deadline.
- Route failure publication through the existing generation-checked lifecycle publication function instead of duplicating its RCU logic.
- Retain existing tests, add six budget regressions and one revoked-generation failure regression, and update the indexing contract's owner/evidence map.

No persisted format, schema, backend policy, batch/query limit, reconciliation fence, or dependency changes. The store-open shutdown behavior fix is intentionally a separate follow-up PR.

## Validation

Exact source commit: `84da525ca6fa44bb0c53b895fe0340ce1c92c725`.
Base: `7f5a460de67965ff36955bab1434c75f080ad384`.

Completed in the isolated [validation run](https://github.com/gosuda/bitcoin-rs/actions/runs/34603714053):

```sh
cargo fmt --all -- --check
cargo clippy --locked -p bitcoin-rs-node --no-default-features \
  --features fjall,redb,zmq --all-targets -- -D warnings
cargo test --locked -p bitcoin-rs-node --no-default-features \
  --features fjall,redb,zmq --lib txindex_worker -- --nocapture
cargo test --locked -p bitcoin-rs-index --features fjall,redb --no-fail-fast
```

All four checks above passed before the source branch was pushed. `git diff --check` also passed. The validation workflow and transformation scripts live only on an isolated workbench branch; this PR contains no CI/workbench helpers.

### Existing ownership-audit failure, not a green gate

`overhaul_ownership` reports **16 passed / 1 failed**. The failing test is `mempool_writer_source_scan_passes`, flagging the existing `handles.mempool_gateway.remove_for_block(` call in `crates/node/src/apply/connect.rs:538`.

The unchanged source-audit executable was also run against a pristine checkout of the pinned base at the same path. It produced the same 16/1 result and identical violation diagnostic. The audit sources, manifests, lockfile, and mempool production code are untouched. The run records both logs and `ownership-status.json`; its comparison step succeeding does **not** mean the underlying audit passed.

The `txindex-final-evidence` artifact records exact patches, source, compiler version, and logs. RocksDB-specific execution and the full workspace behavioral test suite are not claimed by these targeted checks.

---
## Summary by cubic
Refactors the txindex query path by isolating work into dedicated modules and unifying work accounting under `QueryBudget`, with no change to query limits or persisted data.

- Splits the 936-line query module into private modules for transactions, scripts, block source, and budget; public entrypoints retain all existing checks.
- Replaces three repeated pending-batch transfer constructions with `PendingForward::take`, preserving fence, watermarks, capabilities, durable position, and flush deadline.
- Routes failure publication through the existing generation-checked lifecycle function.
- Adds six budget regression tests and one revoked-generation failure regression, and updates the indexing contract's owner/evidence map.

<sup>Written for commit 84da525ca6fa44bb0c53b895fe0340ce1c92c725. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gosuda/bitcoin-rs/pull/946?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


metaphorics:
The separately reviewable store-open shutdown fix is now published as #951, stacked on this branch. It caps each receive wait at 100 ms, checks shutdown/runtime stop/generation revocation, and retains namespace exclusion when the non-cancellable backend helper is abandoned.

Review/merge order: #946 first, then retarget #951 to `main`. #951 contains only the follow-up delta, not another copy of the query refactor.

The completed validation run linked in both descriptions records 71 passing txindex tests for this commit, 79 for the follow-up, 96 passing index-crate tests on this commit, strict Clippy and formatting on both, and the cancellation negative control. The existing 16-pass/1-fail ownership-audit result remains explicitly documented rather than presented as a green gate.

PR #1036, opened 2026-09-14
Add opt-in fast sync (32-peer shallow fan-out) and --validation-mode {full,assume-valid,fast}

metaphorics:
## Summary

Adds an opt-in `--fast-sync` (`BITCOIN_RS_FAST_SYNC`, TOML `fast_sync`) that relaxes block-download scheduling, and a separate `--validation-mode` (`BITCOIN_RS_VALIDATION_MODE`, TOML `validation_mode`) that selects which historical script verification may be skipped. The two are independent; both default to current behavior.

### Fast sync

Expressed entirely through the two existing knobs — the `SyncBudget` installed into `DownloadWindow`/`BlockStager`, and the P2P outbound target — so no new scheduler state or peer-eligibility predicate is introduced:

```rust
// crates/p2p/src/download_window.rs
pub const FAST_BLOCKS_IN_TRANSIT_PER_PEER: usize = 8;   // was 16 (Core cap)
pub const FAST_MIN_PEERS_FOR_FANOUT: usize = 2;         // was 8 (full outbound set)
pub const FAST_OUTBOUND_PEER_TARGET: usize = PENDING_BUDGET / FAST_BLOCKS_IN_TRANSIT_PER_PEER; // 32

pub const fn fast_sync_budget() -> SyncBudget {
    SyncBudget { fanout_peer_inflight: FAST_BLOCKS_IN_TRANSIT_PER_PEER,
                 min_peers_for_fanout: FAST_MIN_PEERS_FOR_FANOUT, ..default_sync_budget() }
}
```

"Maximal peers in estimated bandwidth" is read as: connect to exactly as many outbound peers as the download window (`PENDING_BUDGET` blocks / `PENDING_BYTE_BUDGET` estimated bytes in flight) can keep busy at the fast stripe. A `const` assert pins `FAST_OUTBOUND_PEER_TARGET * FAST_BLOCKS_IN_TRANSIT_PER_PEER <= PENDING_BUDGET`. The existing `effective_peer_inflight` still divides the 256-block window across eligible peers, so 8 is a floor reached at 32 peers (two peers get 128 each).

Node wiring (`state/open.rs`): `outbound_active_limit`/`outbound_peer_target`/`outbound_queue_limit` all follow one `outbound_target` (8 or 32); when `config.p2p.fast_sync` is set the freshly built `BlockSync` gets `install_budget(fast_sync_budget())`.

Not relaxed on purpose: inbound peers are still excluded from fan-out (counting them is the recorded under-fill regression), and the single-peer deep fallback remains for `eligible < 2`.

### Validation mode

`ValidationMode { Full, AssumeValid (default), Fast }` lives in `ValidationOverrides`/`ValidationConfig` next to `assume_valid_height` and is carried on `Chainstate` as `validation_mode`. The single decision point is unchanged in shape:

```rust
// Chainstate::scripts_verified_upstream, BlockProvenance::Network arm
ValidationMode::Full        => false,
ValidationMode::AssumeValid => assume_valid_height > 0 && height <= assume_valid_height && gate.trusted(), // as before
ValidationMode::Fast        => { let tree = block_tree.read(); let tip = tree.tip()?;              // one coherent read
                                 height < tip.height && tree.node_at_height_from(tip.tip_id, height) == tree.lookup(block_hash) }
```

`Fast` reuses the existing `block_tree` rather than adding header-tip state; the block must be the node at its height on the header tip's own chain, so blocks on competing branches and the tip block itself always run scripts, and nothing is skipped when no header tip is known. To keep that decision valid through commit, header-tip writers now join the transition protocol: `drain_inbound_headers` and `submitheader` take `Chainstate::lock_transition()` before the `block_tree` write (lock order transition → tree, same as apply), so the header tip cannot move while a connect/window transition is held. Non-script consensus checks (`run_non_script_checks_only`) still run in every mode. `LocalReplay` is unaffected.

CL-14/CL-19/CL-20 gate evidence was not collected: no throughput claim is made, defaults are not promoted, and docs state fast sync is opt-in and unmeasured.

## Verification
- `cargo build --bin bitcoin-rs --no-default-features --features fjall`
- `cargo test -p bitcoin-rs-p2p --no-default-features` (incl. `fast_sync_budget_stripes_window_across_fast_outbound_target`)
- `cargo test -p bitcoin-rs-node --no-default-features --features fjall --lib` (incl. new `scripts_verified_upstream_follows_validation_mode`, which seeds a header chain plus a competing sibling and asserts only the best-chain block is trusted) and `--test config_layered`
- `cargo test -p bitcoin-rs --no-default-features --features fjall` — new `fast_sync_*` and `validation_mode_*` layering tests pass; the two `g20_formal_models` failures are the Apalache binary being absent locally, unrelated to this diff
- `cargo clippy -p bitcoin-rs-p2p -p bitcoin-rs-node -p bitcoin-rs --no-default-features --features fjall --all-targets -- -D warnings -A clippy::duration_suboptimal_units` clean; the allowed lint is pre-existing on `main` in untouched `crates/p2p/src/compact_blocks.rs:44`
- Regtest end-to-end at `f3440465` (two Core 28.1 peers, ~300 blocks): fresh sync succeeds in fast and default modes; fast logs show `eligible=2 fanout_active=true`, default does not fan out at two peers; env/TOML enable and CLI `--fast-sync=false` overrides both. A live-tip stall after fresh sync (`coinbase witness reserved nonce is missing or the wrong size`) reproduces on `main` and is pre-existing.

---
## Summary by cubic
Adds an opt-in fast sync mode that fans block downloads out earlier and across more outbound peers, plus a separate `--validation-mode` flag for skipping historical script verification; the default sync path is unchanged.

- Fast sync enables via `--fast-sync`, `BITCOIN_RS_FAST_SYNC`, or TOML `fast_sync`: 32 outbound peers, fan-out from two eligible peers, 256-block window divided with an 8-block per-peer floor (defaults: 8 peers, 16 blocks per peer).
- `--validation-mode` defaults to `assume-valid` (pinned-anchor gate); `full` runs every script; `fast` skips scripts only for blocks on the best header chain strictly below its tip, so the tip and competing branches always run scripts.
- `fast` holds the chain-transition lock across header admission so the header tip cannot move between the script-skip decision and a block's commit; inbound header batches and `submitheader` take that lock.
- Stall/eviction timeouts and inbound peer handling are unchanged; throughput is unmeasured and docs mark both modes as opt-in.

<sup>Written for commit 5fa6d6848cb33cedf2e2a71c8252a700e72af119. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gosuda/bitcoin-rs/pull/1036?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Link to Devin session: https://app.devin.ai/sessions/1f624f66ce704471ac717cc333f7cf01
Open in Devin Desktop: https://app.devin.ai/desktop/session/1f624f66ce704471ac717cc333f7cf01?variant=devin
Requested by: @metaphorics


metaphorics:
@devin Touch validation mode. De-strictize flag also.

metaphorics:
@codex review

metaphorics:
@codex review

PR #1077, opened 2026-09-16
refactor(sync): own download window and stager together

### LINKED ISSUE #1075: refactor sync body state under a single owner
## Problem

PR #1071 exposed a structural problem in block sync: `BlockStager` and `DownloadWindow` are separate mutable states, but many of their transitions are actually one invariant.

Today callers must manually keep them synchronized, for example:

```rust
self.block_stager.lock().retire_applied(&hash);
self.download_window.lock().drop_for_retry(&hash);
```

Similar paired mutations exist for receive, retry, expiry, reorg, and malformed-body handling. This spreads ownership across `sync/receive.rs`, `sync/branches.rs`, commit/apply paths, and peer handling, and makes lock ordering part of correctness.

The goal of this issue is deliberately narrow: establish a single owner for block-body download/staging state without changing behavior.

## Proposed first step

Introduce a node-owned state object, e.g.:

```rust
struct BodySyncState {
    window: DownloadWindow,
    stager: BlockStager,
}
```

and replace the separate `BlockSync` fields:

```rust
download_window: Mutex<DownloadWindow>,
block_stager: Mutex<BlockStager>,
```

with one owner:

```rust
body_sync: Mutex<BodySyncState>,
```

At this stage it is acceptable for existing code to access `body_sync.window` and `body_sync.stager` internally. The important change is that their lifetime and lock ownership become explicit and shared.

## Scope

- Introduce `BodySyncState` owned by `BlockSync`.
- Move `DownloadWindow` and `BlockStager` under that state.
- Replace direct access to the old separate mutexes.
- Preserve existing scheduling, staging, retry, stall, peer-blame, and reorg behavior.
- Keep current tests and invariants green.

## Non-goals

Do **not** combine this with broader architectural cleanup yet:

- no `p2p -> node::sync::body` module move in this issue;
- no `InboundBlock -> BoundBlock` admission redesign;
- no `WindowApplyDisposition` redesign;
- no consensus behavior changes;
- no new retry or peer-disconnect policy.

Those should be follow-up work after ownership is centralized.

## Follow-up direction

Once the single owner exists, paired mutations can be replaced incrementally with semantic operations such as:

```rust
accept_body(...)
reject_body(...)
retire_applied(...)
expire_staged(...)
drop_for_retry(...)
```

That should let `sync/receive.rs` and reorg handling stop knowing the internal coordination rules between staging and download state.

## Why start here

The main problem exposed by #1071 is not file placement but distributed mutation. Moving files first would create import churn while preserving the same ownership problem. Establishing one state owner is the smallest behavior-preserving step that makes later cleanup safer and more mechanical.

gosunuts:
Closes #1075.

Move DownloadWindow and BlockStager under one node-owned BodySyncState mutex. Keep P2P policy in the existing components, preserve the sync flow, and update call sites and existing tests for the shared lock.

Static verification: `cargo check --locked -p bitcoin-rs-node --tests --no-default-features --features fjall`; `cargo clippy --locked -p bitcoin-rs-node --all-targets --no-default-features --features fjall -- -D warnings`.

Tests were not run locally per request.

---
## Summary by cubic
Closes #1075. `DownloadWindow` and `BlockStager` previously took separate locks; they now share one `BodySyncState` mutex so the coupled download and staged-body state is always updated together. Sync flow, behavior, and P2P ownership policy are unchanged; call sites and tests lock the shared state. The merge also kept the `BlockSync` executor in `p2p` after main moved it, leaving node as only the chain seam.

<sup>Written for commit 8a1642141bc294c25a202f09e0bb4a2a60c3b2c9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gosuda/bitcoin-rs/pull/1077?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


crates/node/src/sync/headers.rs
```diff
@@ -196,7 +197,6 @@ impl BlockSync {
         };
         let target_height = u32::try_from(target_height).unwrap_or(0);
         let now = Instant::now();
-        let _window = self.download_window.lock();
```
gosunuts, line 199: This guard was doing more than protecting `DownloadWindow`: it also serialized `send_getheaders()` against `on_peer_ready()`'s same-address scheduler reset. Without an identity-bound replacement, an old connection can send, a replacement can become ready and observe no pending entry, then the old path can publish a stale address-only `pending_getheaders` entry afterward. Please make the send + pending publication atomic with respect to connection identity, e.g. by using the lease's `PeerSource` under `PeerTable::with_current`, and add a same-address replacement regression test.

gosunuts (commented):
The ownership refactor itself looks good and matches #1075, but there is one behavior regression to fix before merge.

Removing the old `download_window` guard from `send_getheaders()` also removed the serialization around sending the request and publishing `pending_getheaders`. A same-address replacement can become ready after the old connection sends but before the pending entry is stored; `on_peer_ready()` then sees nothing to clear, and the stale address-scoped pending entry is published afterward. The replacement can be suppressed until `HEADER_REQUEST_TIMEOUT` even though it never received that request.

Please make the send + pending-state publication connection-identity-safe, preferably by carrying the lease/source and performing both under `PeerTable::with_current`, and add a regression test for replacement between send and pending publication.

I don't consider the broader shared-mutex contention concern merge-blocking for this narrow ownership refactor.

PR #1088, opened 2026-09-16
docs(node): add the #1085 Phase-1 ownership inventory

gosunuts:
Part 1 of #1085 (the Phase-1 inventory). This PR does not close #1085; the campaign continues through Phases 2-5. No structural implementation in this PR, per the issue's execution order.

## What this is

`docs/node-ownership-inventory.md` classifies every substantial `crates/node` responsibility as `DELETE / COLLAPSE / KEEP / MOVE`, derived from the current code and call graph at `c0195fbf`, not from the stale #1038/#1073/#1075 plans. `docs/contracts/architecture.md` live gaps now points at it.

Each classification is based on state/invariant/lifecycle ownership and carries `file:line` evidence. Items are marked `[verified]` (re-checked by direct grep/read) or `[hypothesis]` (prediction the subtraction phases must confirm before acting).

## Headline classifications

- **KEEP (genuinely node-owned):** apply core (sole ARCH-07 realisation), sync conductor, state `restore`/`events`, `storage_backend` (enforced ARCH-03 owner), the whole chainstate_journal cluster (no existing owner; distinct codec/replay/writer invariants), checkpoint load/publish core, reorg root, recovery_evidence (witness + marker write + WarningStore are live end-to-end), storage_footprint (FP-01..04), config, embed (EMB-01..08), lifecycle, mining coordinator, tx_ingress, process singles.
- **COLLAPSE:** reorg subtree into one `reorg.rs`; `ChainEffects` into `ChainFollowers`; checkpoint publisher + `state/checkpoint.rs`; `state/{open,index,maintenance}` into lifecycle/index-crate; `NodeState`'s 11 duplicate `Chainstate` fields + test-only getters; `DeferredChainstateServices` seam; `sync/commit.rs` into apply.
- **DELETE (confirmed):** `import.rs` (zero production callers), `embed/testing.rs`, `metrics/uptime.rs`, journal `BlockMeta`, `recovery_evidence::read_marker`, apply `cfg(test)` duplicate hash helpers + wrapper chain, `checkpoint/io.rs` plumbing.
- **MOVE (last resort, each justified in the MOVE ledger):** `checkpoint/headers.rs` to chain; storage-domain constructors in `state/storage.rs` to storage; `apply/contextual.rs` rule set to consensus/chain.

## Verification corrections recorded for #1076/#1084

Direct grep against `c0195fbf` refuted three assumptions, now documented in the inventory:

[...]

Issue #1085, opened 2026-09-16
Reset `crates/node` architecture from the current codebase


gosunuts:
## Context

The current `crates/node` tree has reached a point where continuing the existing refactor sequence is more likely to preserve accidental structure than improve ownership.

The main problem is no longer simply that `node` owns too much. It is that earlier decomposition work has often kept the original glue/coordinator in place while also adding a large hierarchy of submodules. We now have several `foo.rs + foo/` structures where the top-level file is still substantial and the extracted submodules are also substantial.

That means the repository has often moved from:

```text
large node module
```

to:

```text
large node module
+ many node submodules
+ additional adapters/glue between them
```

without reducing the architectural knowledge held by `node` enough.

Issues #1038, #1073, and #1075 are being treated as stale because they continue from intermediate ownership/decomposition assumptions that should no longer be considered the baseline.

#1076 remains relevant and should continue as the deletion-first cleanup pass before or alongside this reset.

## Goal

Re-evaluate `crates/node/src` from the **current behavior and current call graph**, not from the current file/module layout.

The first deliverable is an ownership inventory. Every meaningful node module/responsibility should be classified as one of:

```text
DELETE    obsolete, duplicate, transitional, compatibility, or unnecessary glue
COLLAPSE  unnecessary internal layering that should become one simpler unit
KEEP      genuinely node-owned composition/orchestration/lifecycle
MOVE      necessary domain behavior that demonstrably belongs to an existing owner
```

The order above is intentional. **DELETE is the default hypothesis. MOVE is the last resort.**

Do not start by designing a new directory tree.

Do not assume any current `node/*` subdirectory deserves to survive simply because it already exists.

## Core principle

This is **not another module-splitting pass** and **not a relocation pass**.

The reset should primarily make code disappear.

During this cleanup, do not create a new `node` module merely to make an existing file smaller, and do not treat moving the same implementation to another crate as architectural progress by itself.

A structural change should do at least one of the following:

- delete behavior/glue that no longer needs to exist;
- collapse multiple ownership surfaces into one simpler path;
- remove behavior from `node` entirely because an existing owner already provides the capability;
- only when deletion/collapse is not sufficient, move a complete state/invariant/lifecycle slice to its real existing owner;
- leave behind a demonstrably thinner cross-component coordinator.

If roughly the same behavior remains after being redistributed across more files or crates, the change has not improved the architecture.

## Hard deletion-first rules

[...]

Issue #1084, opened 2026-09-16
test: remove stale refactor/evidence gates that freeze implementation shape


gosunuts:
## Problem

The test suite currently mixes two very different responsibilities:

1. tests that protect Bitcoin behavior and compatibility, and
2. tests that preserve historical refactor plans, benchmark bookkeeping, source-file layout, or current implementation shape.

The second category has grown large enough to make architectural cleanup harder. Several tests now fail when code is reorganized even if observable Bitcoin behavior remains unchanged.

The goal of this issue is to aggressively remove or demote tests whose primary purpose is to preserve a past implementation plan rather than a durable product/consensus contract.

### `bin/bitcoin-rs/tests/overhaul_ownership.rs`

Remove this file.

It duplicates the dependency/ownership gate machinery, includes many synthetic validator-self-tests, and repeatedly scans the source tree to enforce the current ownership layout.

### `bin/bitcoin-rs/tests/support/ownership_scan.rs`

Remove the lexical source scanner and its allowlists/pattern inventories such as:

- `AUTHORIZED_GATEWAY_CALLS`
- `MUTATING_METHODS`
- `POOL_WRITE_PATTERNS`
- `INDEX_CAPABILITY_PATTERNS`
- `PEER_MUTATION_METHODS`
- `TRANSITION_PROMOTION_PATTERNS`
- pressure-floor owner/path inventories

These checks freeze receiver expressions, file locations, method names, and package ownership through string matching. Architectural ownership should primarily be expressed by Rust visibility/types/crate boundaries, not a second ad-hoc parser for the source tree.

### `bin/bitcoin-rs/tests/gates/g17_dependency_direction.rs`

Keep the `cargo metadata`-based crate dependency-direction check if it remains useful, but remove `workspace_single_writer_boundaries_are_respected` and the dependency on `ownership_scan`.

### `bin/bitcoin-rs/tests/gates/g18_hot_path_ledger.rs`

Remove this file from the normal Rust test suite.

The 36-cell benchmark matrix, required path inventory, architecture/backend/corpus combinations, and benchmark bookkeeping are performance/evidence tooling concerns, not product correctness.

### `bin/bitcoin-rs/tests/overhaul_evidence.rs`

Remove this file.

Tests that require the checked-in benchmark ledger to contain an exact matrix of empty cells or preserve measurement-history structure should live in benchmark tooling/CI if still needed, not in the normal product test suite.

[...]

Issue #1076, opened 2026-09-16
cleanup: delete legacy and dead code before further ownership refactors


gosunuts:
## Context

Before continuing structural work such as #1075, do a deletion-first cleanup pass over the current tree.

Recent refactors have accumulated transitional code, duplicated implementations, stale task scaffolding, and lint suppressions. Refactoring ownership on top of that risks preserving accidental structure and turning temporary compatibility paths into permanent architecture.

The goal of this issue is not to find a better home for existing code. It is to reduce the amount of code that needs a home at all.

## Principle

For this cleanup phase, production code should have to justify its existence through a real caller or a current protocol/runtime requirement.

Prefer deletion over movement or abstraction.

The repository should become **physically smaller**, not merely differently organized.

A cleanup PR is only successful when it removes code and, where redundant modules/files exist, removes files as well. Moving the same implementation between modules, splitting one file into several, renaming legacy code, or wrapping duplicate paths behind a new facade does not count as cleanup.

The reason for making this strict is deliberate: the current problem is accumulated duplication and transitional residue. If success is defined only in terms of ownership or layout, AI-assisted refactors can repeatedly move the same code around while preserving the underlying duplication. This phase must force actual subtraction.

### Hard success conditions

For deletion-first cleanup PRs:

- production LOC must decrease materially;
- the number of production source files should decrease whenever redundant/transitional files are removed;
- duplicated implementations, wrappers, compatibility layers, and stale modules must be deleted rather than relocated;
- exported/internal symbol surface should shrink where obsolete APIs are removed;
- additions must be clearly smaller than deletions;
- a PR that mostly moves, renames, splits, wraps, or re-exports existing code is **not** considered successful cleanup;
- if the same behavior still exists in roughly the same amount of code after the PR, the cleanup has failed even if the directory structure looks better.

File-count reduction is not an arbitrary aesthetic goal. It is a guardrail against replacing one duplicated path with the same path under a different filename. When a legacy/transitional implementation is no longer needed, its file should disappear rather than survive as `*_impl.rs`, `legacy.rs`, `compat.rs`, or another helper module with the same code.

### Rules for cleanup PRs

- no file/module moves unless strictly required to delete code;
- no file splitting as a substitute for deletion;
- no new facade, trait, wrapper, adapter, or state object;
- no behavior changes;
- no speculative APIs kept for future work;
- remove obsolete compatibility/transition paths instead of renaming them;
- keep consensus/regression/e2e tests that protect behavior;
[...]

PR #1090, opened 2026-09-16
refactor(node): delete verified-dead surface from the #1085 Phase-2 list

gosunuts:
Part of #1085 (Phase 2: subtraction only, from the #1088 inventory). Continues #1076's deletion-first campaign. No moves, no new modules, no collapses beyond what deleting dead code forced.

Every candidate below was re-verified against current main (grep for production callers, then compile) before deletion. Items that failed re-verification were kept and are listed as inventory corrections.

## What was deleted

**Commit 1 — verified-dead surface:**
- `crates/node/src/import.rs` + `import/tests/` (5 files): the module docstring called it a skeleton; `import_block` had zero production callers, only its own tests.
- `metrics/uptime.rs` + re-exports + its test: `record_process_start` was never called in production, so the uptime clock was never set in a real process.
- Journal `BlockMeta` + `JournalRecord::block_meta()` + root re-export + the test assertion that pinned it: a projection with no production reader.
- `recovery_evidence::read_marker` + the stale `#[cfg_attr(not(test), allow(dead_code))]` on `WarningSnapshot::with_index`: the runtime never reads the marker (write-only audit evidence); the RCV-12 marker tests keep their oracle as a test-local `read_marker`, per the same pattern #1076 prescribed for the mempool `EntryHandle`.
- `apply.rs` `cfg(test)` `sha256d`/`merkle_root_bytes`: the single fixture caller now uses `bitcoin_rs_consensus::verify_block::compute_merkle_root` (identical signature).
- `apply/scratch.rs` `cfg(test)` `is_coinbase`: both callers now use `bitcoin_rs_utxo::is_coinbase_tx` (byte-identical semantics).
- `embed/testing.rs`: `pub(crate)` test-only `block_on`; `embed/tests.rs` gets the same local helper the external embed test already carries.
- `NodeState::chain_event_publisher`/`chain_event_hints` getters: only `state/tests/events.rs` called them.

**Commit 2 — the cascade the first deletion proved:**
- The entire hint channel end-to-end: deleting the getters exposed that the `chain_event_hints_rx` field had no production reader, which proves the channel itself is production write-only (no runtime consumer ever drained it; the index worker consumes `ChainCursorSource` snapshots + its own wake channel). Deleted: the publisher's `hints: Sender` field, the bounded construction and `Receiver` returns in `new()`/`detached()`, the `try_send` in `record()`, `CHAIN_HINT_CHANNEL_LIMIT`, the open.rs wiring, and the `record_drops_hints_when_channel_full` test (it pinned the drop behavior of a channel nobody reads). `record()` keeps returning the committed `ChainEventHint` (tests bind it; production discards), and `HintKind`/`ChainEventHint` remain the receipt types. `detached()` survives with a `Self`-only signature (7 test/bench callers updated).
- `apply::check_coinbase_maturity` cfg(test) wrapper + `NodeState::check_coinbase_maturity` cfg(test) method: compiler-proven dead once import went. Attribution proven, not inferred: `cargo check -p bitcoin-rs-node --all-targets` on the pristine pre-PR tree reports zero dead-code warnings, so these were live only through the deleted import tests.
- `chainstate_journal.rs` root residue: the three `#[allow(unused_imports)]` attributes and the stale "Task 2/4/5" comments are gone. Removing the allows let the compiler name the re-exports nothing imports through the root path (`JournalEmit`, `JournalRecordError`, `Mutation`, `decode_record`, `encode_record`, `HeadMarker`, `JournalWriterFailpoint`); those names are deleted from the re-export lists while the submodules keep them (consumers import directly). `Coin`, `JournalRecord`, `SharedJournalWriter`, `shared_journal_writer`, and the writer/delta/replay sets stay — the compiler confirms they are used through the root.

## What was kept after re-verification, and why

- `apply.rs::block_txids` (cfg(test)): ~40 live call sites across `consensus_rule_tests`; the fixture path for `BlockView` construction. Not dead; correction recorded below.
- The marker WRITE path (`write_marker`, `ChainRollbackEvent`, `reporter.rs`): confirmed live end-to-end (index rollback reconcile invokes `report_index_ahead`, RPC-visible via `getblockchaininfo.warnings`).
- `JournalEmit`/`HeadMarker`/`JournalWriterFailpoint` as items: live inside the journal cluster (writer/replay submodules); only the unused root re-export path was deleted.
- `metrics/evidence` and the warnings write-side: untouched here (gates still exist under #1084; warnings write-side needs the audit call).
- `docs/contracts/chain-events.md`: no clause names the in-process hints receiver (EVT-01 still names `active_chain_snapshot()`, which stays; EVT-02's "index wake hints" is the separate wake channel; EVT-03 consumers use the persisted cursor), so no contract drift.

## Numbers

```text
Total:            29 files, +73 / -1113 (net -1040 lines)
Files removed:    7 (import.rs, import/tests.rs + 3 test files, embed/testing.rs, metrics/uptime.rs)
Production LOC:   ~-215 non-test-gated (import 49, uptime 38, state getters/field/method/const ~35,
                   events.rs channel ~30, BlockMeta ~30, read_marker 14, journal re-exports ~10, wiring ~9)
cfg(test)/test:   ~-825 (import tests 669, drop-behavior test, uptime test, merkle/coinbase helpers, journal test bits)
Public API:       `pub mod import`, `pub use uptime::{...}`, `pub` NodeState hint getters,
                   `detached()`/`new()` Receiver returns removed
Tests:            469 lib tests (one dead drop-behavior test removed), all integration suites green
```

Verification: `cargo check --workspace --all-targets` clean, `cargo test -p bitcoin-rs-node` all suites green, `cargo clippy -p bitcoin-rs-node --all-targets -- -D warnings` clean, `cargo fmt` applied, plus the pristine-tree dead-code discriminator for attribution.

## Corrections to the #1088 inventory

1. `apply.rs::block_txids` is **not** a deletion candidate (cfg(test) but ~40 live fixture callers); reclassify DELETE -> KEEP.
2. Items the inventory missed, now deleted: the `chain_event_hints_rx` field **and the entire write-only hint channel beneath it** (StateCluster flagged only the getters; the channel was fully dead in production), plus the two `check_coinbase_maturity` test entrypoints (import-test scaffolding, proven by the pristine-tree discriminator).
3. #1076's journal re-export list was one quarter right, not three quarters wrong: `BlockMeta` and the seven names deleted here had no root-path callers; `JournalEmit`/`HeadMarker`/`JournalWriterFailpoint` were wrongly listed as fully dead — the items are live, only the root re-export path was unused.
4. The events hint tests moved from `NodeState`-hosted to `detached()`-hosted; no contract page needs updating.

Sync ownership, journal removal, checkpoint ownership, `wake_tx_index`, and the metrics/warnings audit remain untouched for later phases, as instructed.


gosunuts (commented):
The deletion direction looks good overall, including removing the write-only chain-event hint channel and the dead import/uptime surfaces. I found one blocking issue before merge: this PR removes/changes public `bitcoin-rs-node` APIs (`pub mod import`, uptime exports, and `ChainEventPublisher::detached()`'s return type) while the workspace remains at `0.5.0`. `docs/policies/source-compatibility.md` explicitly requires a minor version bump for breaking public API changes during `0.x`. Please keep the clean cutover (no compatibility shims), but bump the workspace to the next minor release and update the matching internal workspace dependency versions / lockfile. After that, this is approveable from my side.

gosunuts (commented):
Reviewed the latest head. The deletion-first cleanup is coherent, the public API cutover is correctly versioned as 0.6.0, the RPC parity fixture has been repinned, and current CI / fixture / fuzz-corpus checks are green. I don't see a runtime correctness blocker in the retained paths. One follow-up remains: regenerate `fuzz/Cargo.lock` so the separate fuzz workspace records the internal path crates as 0.6.0. Treating that as non-blocking for this PR and handling it in the next cleanup step.

Issue #1091, opened 2026-09-16
Node wedges at tip without exiting: staged-but-unapplied body suppresses stall recovery (gap=1 freeze), then the whole runtime stalls (RPC 503 busy, telemetry and index heartbeat dead)


gosunuts:
## Summary

A live mainnet node (explorer compose, image built 2026-09-16T01:38Z, binary sha256 `686f030afb047dce…`, i.e. before #1071/#1072) wedged in two stages after catching up to the tip:

1. **Silent tip-follow freeze**: header acceptance and block apply both stop at `applied 967,271 / header 967,272` (`gap=1`), while the public tip advances. Three peers stay connected, zero error/warn lines, mempool unknown (RPC later unusable). The window's stall detection never fires because the next-expected block's body is staged but never consumed, which suppresses the staller logic entirely (`apply_side_busy` short-circuit).
2. **Full runtime stall ~12h later**: at `2026-09-16T23:00:37Z` the 60s sync telemetry, the 30s txindex heartbeat, and RPC all stop mid-cadence at the same timestamp. RPC answers `503 busy` for every request — including `getblockchaininfo` — because all worker slots are occupied by handlers that never finish. The process keeps running (no exit, no panic); the container only reports `unhealthy`.

No periodic chainstate checkpoint was published during the entire 13h run, so the durable state remained at the boot generation.

## Reproduction

Not reproducible on demand; this is a live occurrence on the explorer node (`tools/btc-mempool-explorer` compose, default prod feature set). Timeline (all timestamps UTC, from `docker logs` and docker inspect, captured before any restart):

- `10:03:12` container start (restored from checkpoint gen-475)
- `~10:11–10:20` catch-up to tip (967,26x), txindex consistency gate recovered
- `11:04:03` last `header_height` change ever: 967,272 (applied frozen at 967,271) — ~1h after tip-catch
- `11:04–23:00` telemetry repeats `applied_height:967271, header_height:967272, gap:1, peers:3, ibd:true` every 60s; nothing else logs
- `23:00:37` last log line of any kind (txindex heartbeat); RPC `getblockchaininfo`/`getmempoolinfo` return `busy` with HTTP 503; docker healthcheck fails with 503 → `unhealthy`
- public tip at observation time: **967,333** (blockstream.info and mempool.space agree)

## Evidence

**Stage 1 mechanism — staged-but-unconsumed body suppresses all stall recovery:**

[...]

PR #1092, opened 2026-09-17
fix(#1091): time-bound apply-side stall suppression with blameless evict-and-refetch

gosunuts:
## Summary by cubic
Time-bounds the apply-side stall suppression (issue #1091): a well-formed body staged but never applied used to hold the window frozen indefinitely (observed live as 13h at `gap=1`). Past twice `received_timeout`, the stuck body is now evicted for refetch without blaming any peer.

**Bug Fixes**
- The stuck clock keys on the apply-front `(height, hash)` and runs only while a body is staged, so prune/refetch seams don't reset it, real progress or a same-height branch replacement does, and a slow first delivery is never penalized.
- On fire the sync layer evicts the stuck staged body for refetch; a final frontier check prevents evicting a body that no longer stalled.
- The clock re-arms on fire, so a persistently stuck frontier escalates at most once per bound.
- Records `node.sync.apply_side_stall_escalations` and logs `suppressed_secs` on escalation.

<sup>Written for commit 726ab79fcdb0118e5cff5b952f0ffda558a7e395. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gosuda/bitcoin-rs/pull/1092?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


gosunuts (commented):
The overall direction looks right: bounding `apply_side_busy` suppression and recovering via blameless evict-and-refetch is a good fit for the stage-1 `gap=1` freeze in #1091. I think there are still two correctness issues to fix before merge, though.

1. `observe_apply_side_bound()` currently creates the stuck episode before checking `apply_side_busy`. That means the timer can accumulate while no staged body exists at all. If the frontier sits undelivered for >= 2x `received_timeout`, its first normal delivery can inherit that idle time and be evicted immediately. The timer should begin only once the expected frontier body is actually staged / `apply_side_busy == true`, while still surviving the intended brief prune/refetch seam for that same frontier.

2. `ApplySideStuck` is keyed only by height. The expected block hash at `applied_tip + 1` can change at the same height during a branch replacement/reorg, so a fresh body from the replacement branch can inherit the previous branch's elapsed stuck time. Please key the episode by the actual expected frontier hash (height + hash is fine) and reset it when that hash changes. The escalation path should also verify that the frontier being drained is the same one that triggered the bound.

I would not cherry-pick #1093 as-is. Its fix calls `next_expected_block_hash()` from inside the `select_and_evict_window_peer()` callback while `download_window` is already locked; `next_expected_block_hash()` reads `block_tree`, and this codebase explicitly avoids window -> tree ordering because tree -> window is used elsewhere. A cleaner shape is to snapshot `frontier_hash` once before entering `select_and_evict_window_peer()`, derive `apply_side_busy` from that same snapshot, pass `(height, hash, busy, now)` into `observe_apply_side_bound()`, and pass the captured hash to escalation for a final equality check.

Please also add regressions for (a) a frontier that remains idle longer than the bound before its first body arrives, and (b) a same-height frontier-hash replacement resetting the timer.

One scope note: this addresses stage 1 of #1091. The later process-wide/RPC/telemetry stall in that issue still has no identified lock holder, so I would avoid treating this PR alone as closing the whole issue unless that stage is split into a follow-up.

gosunuts:
Scope note for reviewers: this PR addresses stage 1 of #1091 only (time-bound suppression + blameless evict-and-refetch, now with busy-gated episodes and frontier-hash keying). Stage 2 (whole-runtime stall, uncaptured lock holder) is tracked separately in #1094 — see also the boundary note on #1091.

gosunuts (commented):
Re-reviewed at `726ab79`.

The previous correctness blockers are addressed: the apply frontier is snapshotted before taking the download-window lock, the stuck episode is keyed by `(height, hash)`, idle time no longer starts the timer, same-height branch replacement gets a fresh episode, and the regression now verifies that escalation actually requeues the frontier for `getdata`.

The remaining frontier-check/drain TOCTOU is not a merge blocker in the current shape: the eviction still targets the captured old hash, so a concurrent branch replacement cannot drain the new branch's fresh body. At worst it retires stale staged state early and causes extra refetch churn. Avoid adding broader tree/stager synchronization here unless live evidence shows that churn matters.

CI is green on the current head. This is good to merge for #1091 stage 1; the separate process-wide/RPC stall described in the issue should remain tracked independently.

PR #1102, opened 2026-09-20
Keep external evidence out of correctness tests

metaphorics:
## Summary

Normal Cargo correctness tests no longer launch an external Java/Apalache solver or freeze historical refactor layouts, benchmark matrices, and a hardcoded validation-default verdict. Independent consensus results, custody failures, public behavior and durability evidence remain the reason to keep a test.

Formal checks now have one Python/manual-workflow owner with pinned model/tool identities, the same six properties at K=128, explicit incomplete/counterexample outcomes, process-group timeouts and retained run artifacts. Core-only test jobs no longer provision Java. Useful benchmark accounting checks run with the benchmark tooling; cargo-deny alone owns resolved-graph uniqueness. No compatibility aliases retain the retired Cargo targets.

Fixes #1084. This is the first contract-first cut for #1099, not a claim that its full test-function audit is complete. The reviewed decisions and remaining coverage gaps are recorded in `docs/contracts/test-evidence.md`.

## Validation

- Consensus native suite: 117 tests passed, including the retained independent parse/identity and input-refusal suites.
- Reference custody: 12 tests passed. Benchmark evidence: 4 tests passed.
- Python fixture tools: 87 tests passed, including six new formal-runner custody/outcome/timeout regressions.
- Clippy passed with warnings denied for all targets of the binary (fjall), node (fjall), and native consensus. Rust formatting, shell syntax and diff whitespace checks passed.

Commands: `cargo test --locked -p bitcoin-rs-consensus --no-default-features`; `cargo test --locked -p bitcoin-rs --no-default-features --features fjall --test overhaul_reference_set`; `cargo test --locked -p bitcoin-rs-node --no-default-features --features fjall --bench evidence`; `python3 -m unittest discover -s scripts/tests -v`.

## Known residuals

The solver itself, pinned Core process campaigns, dependency-range endpoints, and performance campaigns were not run in this change. Existing BLOCKED/UNMEASURED evidence remains unchanged. Runner tests use a synthetic executable and do not establish model properties. A complete individual review of the remaining node/P2P state-oriented tests and clause-to-scenario gaps remains under #1099.

Manual diff review completed. An independent automated review could not run because the local coding helpers were unavailable.

## Post-Deploy Monitoring & Validation

[...]

Issue #1099, opened 2026-09-20
Rebuild the test suite from current contracts


gosunuts:
## Context

The repository already defines `docs/contracts/` as the owner of current normative behavior. Tests should therefore act as executable proof of current contracts, not as an accumulated record of historical implementation shapes.

The current test suite has grown through many intermediate architectures and refactors. A significant portion of it now protects:

- old ownership boundaries;
- wrapper/helper structure;
- historical migration seams;
- duplicated fixtures and near-identical scenarios;
- test-only production APIs that otherwise have no runtime consumer.

That creates the wrong dependency:

> production structure survives because old tests call it

instead of:

> tests survive because they prove a current behavioral contract

Recent deletion-first work has already shown this effect: removing dead paths exposed additional test-only shims and production surfaces that could disappear with them.

## Goal

Rebuild the test suite from the current contracts instead of incrementally preserving the existing suite.

This is intentionally a destructive reset.

The target model is:

```text
docs/contracts/* = normative behavior
tests            = minimal executable proof of those contracts
bug regressions  = first promoted into a contract/invariant, then tested
implementation shape = not a reason for a test to exist
```

The purpose is not simply to reduce test LOC. The purpose is to remove historical tests that constrain obsolete architecture, expose test-only production surface, and leave a smaller suite whose reason for existence is explicit.

## Principle

Every surviving test should answer:

> Which current contract clause or real bug invariant does this prove?

If there is no good answer, the default is deletion.

A test should not survive merely because it currently passes, has existed for a long time, or exercises an internal helper.

## Before deleting tests

First bring `docs/contracts/` in sync with the current code.

Some contract documentation still contains stale ownership/current-state descriptions after the recent node, P2P, index, storage, UTXO, and mining refactors. A destructive test reset is only safe if the normative baseline is current.

For each contract clause, identify the minimum executable proof required to catch a meaningful violation.

Examples:

- `ARCH-*`: dependency direction, ownership boundaries, composition invariants;
- `RCV-*`: crash recovery, torn writes, corruption, checkpoint/journal fallback, disconnect debt;
- `MPL-*`: mempool mutation ordering and observable sequencing;
- `API-*`: public RPC behavior and Core-compatible projections;
- `P2P-*`: peer-visible wire behavior and lifecycle invariants;
- consensus/DAA rules: Core/vector-backed acceptance and rejection boundaries.

## Existing-test classification

Do not mechanically port every current test.

[...]

Issue #1094, opened 2026-09-17
follow-up(#1091): capture the stage-2 runtime-stall lock holder (telemetry-cadence watchdog)


gosunuts:
Split from #1091 stage 2, which PR #1092 explicitly does not address.

## What remains open

In the #1091 live run, ~12h after the stage-1 gap=1 freeze, one timestamp (2026-09-16T23:00:37Z) killed the 60s sync telemetry, the 30s txindex heartbeat, and RPC together: every RPC request (incl. getblockchaininfo) answered 503 busy (all worker slots held by never-returning handlers, crates/rpc/src/server.rs:120-131), logging stopped entirely, process stayed alive with no exit or panic, and no periodic chainstate checkpoint had published during the 13h run. The exact lock holder was never captured.

## Proposed deliverable

A watchdog that dumps blocked-mutex/thread state when the telemetry cadence misses (e.g. no sync-progress emission for N ticks while the process is alive), so the next occurrence names the holder instead of only showing the signature. Telemetry is emitted from the single event loop (crates/node/src/event_loop.rs) and the stall path takes stager/window mutexes (crates/node/src/sync/peers.rs, telemetry.rs:43-44) — the watchdog must not itself depend on those locks.

## Scope note

#1092 (stage-1 bound + blameless evict-and-refetch) covers the gap=1 freeze path only; this issue tracks the process-wide stall diagnosis. Evidence log preserved at /tmp/wedge-evidence/2026-09-17-both-freeze-rpc-busy.log on the operator workstation.

gosunuts:
Field evidence, 2026-09-17: TWO clean-stop publish hangs on the pre-fix image, same signature both times. (1) 10:29Z SIGTERM after the 08:46-10:28 gap=1 freeze: rpc+p2p listeners exited cleanly, then no `published clean chainstate checkpoint` for 15+ min, process alive at ~0% CPU with only txindex heartbeat lines; SIGKILL (exit 137), journal replay 173 records, zero loss. (2) 11:34Z SIGTERM after the third wedge (frozen 967399 vs tip 967402 from ~11:11Z, peers 8->4): identical hang, kill at 11:49Z. In both cases telemetry had been alive pre-stop and CURRENT stayed at gen-475 (no periodic checkpoint published in either run). Contrast: post-swap stop on the healthy node today published gen-476 cleanly in ~7min. The hang correlates with the wedged state, supporting the stuck-lock-holder hypothesis for stage 2.

PR #1103, opened 2026-09-20
Recover idle sync without discarding live requests

### LINKED ISSUE #1098: Witness-stripped deliveries of a valid best-chain block exhaust requeue and leave tip-follow idle with all recovery paths disarmed
## Summary

Post-#1091/#1092 field evidence from the explorer compose (binary sha256 `7dc10378…`, built from `84ae08db`): the valid best-chain block 967,772 was delivered with a coinbase witness that fails the size/shape check, the requeue loop ran for ~3 minutes and then went idle, and the node froze at `gap:1` for 47+ minutes with every recovery path disarmed. `getblockhash` for the announced height returns `NotFound` while sync telemetry keeps reporting `header_height: 967772` — telemetry and block tree disagree.

This is the #1070 trigger class (witness-stripped peer delivery) resurfacing downstream of the staging-gate fix: rejections are retried now, but retry exhaustion has no owner — nothing staged (so the #1092 apply-side clock is disarmed by design), nothing pending (so request timeouts never fire).

## Reproduction

Not reproduced on demand — live occurrence during mainnet tip-follow; full container log preserved before any restart (see Evidence).

## Evidence

All timestamps UTC 2026-09-20, from the container log saved at kill time (91,414 bytes):

- `01:04:53.979` — INFO `block sync: requested compact blocks near tip` (log target `bitcoin_rs_p2p::sync::requests`)
- `01:04:54.236` — INFO `p2p compact reconstruction fallback` (crates/p2p/src/listener.rs:1186); the fallback fetches `Inventory::WitnessBlock` (crates/p2p/src/listener.rs:1188-1193)
- `01:04:54.738` — WARN `block sync: body/header binding failed; rejecting delivery`, hash `000000000000000000013e22ca5bf2bed2ce65964c1313c851be23418d3734b7` (the public explorer indexes this hash at height 967,771), error `coinbase witness reserved nonce is missing or the wrong size` (crates/p2p/src/sync/receive.rs:204; check at crates/consensus/src/verify_block.rs:393-400; error at crates/consensus/src/lib.rs:161)
- `01:07:58.303` and `01:07:59.210` — the same WARN for `0000000000000000000213bc47970cf674b607527d441e8df9f3c72300990b5c` — this is the real best-chain 967,772 (public explorers: 1,597,442 bytes, 6,680 tx, coinbase witness = the standard 32-byte zero nonce, identical in shape to neighboring blocks). The delivered bytes were not faithful to the block.
- `01:09:57`–`01:10:58` — re-request from a third peer, then `peer missed the block request timeout; disconnecting and re-queueing` — the last non-telemetry line of the run.
- `01:10:58`–`02:00:34` (SIGKILL) — 47 minutes of silence: zero `stuck past the apply-side bound` escalations (bound = 2 × `RECEIVED_BLOCK_TIMEOUT` = 2 min, crates/p2p/src/download_window.rs:33 and :990 — it would have fired ~23× had the body been staged; the clock only runs while a body for the frontier is actually staged, crates/p2p/src/download_window.rs:970-972), zero request timeouts, `applied_height: 967771, header_height: 967772, gap: 1` repeating every 60 s.
- Mid-freeze (`01:52`) — `getblockhash(967772)` → `not found: block height not found` (crates/rpc/src/handlers/chain.rs:524) while telemetry still reported `header_height: 967772`. Public tip was 967,776.

Open questions (stated, not asserted):

- Where the witness bytes are lost. The fallback requests the correct inventory type (`WitnessBlock`), and two different peers produced the same defective delivery — so either responders strip witness for this fetch shape, or bitcoin-rs's raw-block decode drops it.
- Why the 967,772 header never resolves in the tree (`getblockhash` NotFound) while telemetry counts it. The subtree handling around the rejected delivery was not captured before the restart.

## Expected vs actual

Expected: a binding-failed delivery of a valid block is retried until a faithful body arrives, and if requeue exhausts, some bounded path re-arms the fetch; `getblockhash` answers for any height telemetry reports as accepted.

Actual: ~3 minutes of retries, then permanent idle — no staged body (#1092 escalation disarmed), no pending request (timeouts idle), headers beyond 967,772 never accepted (public tip 967,776), node frozen until a manual SIGKILL (cure; loss bounded to 7 blocks by the preceding clean checkpoint gen-477).

## Scope

- Mechanism stage is downstream of #1070's fix: rejections are retryable now; the gap is that retry exhaustion has no bounded escalation, and the sync/frontier state can go fully idle.
- Related but distinct: #1091/#1092 (staged-but-unapplied body — the apply-side clock needs a staged body, which this state never has), #1094 (stage-2 process-wide stall — RPC and telemetry stayed alive here).
- Candidate follow-ups: (1) log delivered-vs-header coinbase witness shape (item count, byte length) at binding failure to pin down where the bytes are lost; (2) a bounded frontier re-fetch when the tree cannot resolve the next-expected height — today no clock covers the idle-frontier state; (3) reconcile telemetry `header_height` with the block tree.

metaphorics:
## Summary

A node with known headers but no remaining body work now relearns peer capability and refetches the missing prefix without restart. Previously, losing the last credited peer after rejecting a body could leave a permanent idle gap; an applied rollback with unchanged headers could also leave the forward cursor beyond the required block.

The applied chain owns the frontier. Recovery rewinds only an unowned gap and preserves live requests and staged bodies. Idle discovery reuses the existing header request and deadline, probes from the applied chain, and rotates after empty responses time out. Accepted active-branch headers, not an invented height, grant body capability.

Peer-ready handling now uses the same table-before-download lock order as request publication. This removes a concrete potential lock cycle involving a queued peer-table writer. It is not proof that this cycle caused the historical runtime freeze. Header-only peers can no longer leak into the body fallback, and rejected-body logs expose witness shape and source without dumping transaction payloads.

Fixes #1098. Related: #1091, #1094, #1099. Stacked on #1102; review against `fix/contract-test-cut`.

## Validation

Three new regressions failed before the fixes: refetch after applied rollback, recovery after the last credited peer is rejected, and cancelled readiness while a body writer is held. They now pass. A fourth regression checks empty-response pacing and rotation. The rejected-witness scenario proceeds through a new header probe, a witness-body request and successful application.

- `cargo test --locked -p bitcoin-rs-p2p --lib`: 349 passed.
- `cargo clippy --locked -p bitcoin-rs-p2p --all-targets -- -D warnings`: passed.
- Formatting and diff whitespace checks passed.

Four previous test cases simulated application by erasing download slots without advancing the chain. Their fixtures now deliver and apply real mined blocks. The retired shortcut is private to window unit tests, not a production API.

## Known residuals

The historical mainnet stall and source of stripped witness bytes have not been reproduced here. Live Core interoperability and the full node process campaign were not run in this layer. This fixes deterministic recovery and a demonstrable lock-order hazard; independent whole-runtime stall diagnostics remain #1094 work. The complete individual test audit in #1099 is still open.

Manual diff review completed. Independent automated review was unavailable because the local coding helpers could not run.

## Post-Deploy Monitoring & Validation

During the first maintainer-operated at-tip/reconnect rehearsal, watch `node.sync.frontier_rewinds`, `node.sync.idle_frontier_probes`, applied/header height, and pending/staged counts. A rejected delivery should lead to replacement work and applied progress; repeated empty probes must remain paced by the header deadline. Correlate `body/header binding failed` with `p2p compact block reconstructed` using the block hash and witness-shape fields.

Capture diagnostics and revert this layer if requests duplicate live ownership, wrong-branch bodies are selected, or an idle known-header gap remains after an honest peer supplies headers. Preserve the datadir and existing logs; no reset or validation-default change is part of mitigation. The maintainer owns this operational rehearsal; it was not executed by this PR.

---
## Summary by cubic
Fixes #1098 by recovering an idle sync frontier without restarting or discarding live requests. Previously, losing the last credited peer after rejecting a body could leave a permanent idle gap, and an applied rollback with unchanged headers could leave the forward cursor past the required block.

- Treats the download cursor as a scan hint: only an unowned frontier is rewound, and pending and staged bodies keep their ownership.
- Fires an idle header probe from the applied chain only when the apply-frontier block itself is not pending or staged; staged successors behind a rejected frontier are stuck inventory and don't suppress recovery.
- Reuses the existing getheaders request and deadline, rotating after empty responses time out. Only accepted active-branch headers grant body capability, and header-only peers no longer fall back into body selection.
- A failed probe send falls back to normal header selection within the same tick, excluding that source so a dead highest-height peer can't mask a live lower peer.
- `send_getheaders` carries the inspected `PeerSource`, so a superseded same-address lease cannot receive a request.
- Peer-ready handling locks the peer table before the download window, removing a potential lock cycle and keeping a cancelled ready event from waiting behind a body writer.
- Rejected-body logs add source, byte counts, and witness shape; compact reconstruction logs include the block hash.
- Adds regressions for applied rollback, rejected last credited peer, cancelled readiness, empty-response pacing, staged successors, superseded sessions, and failed-probe fallback.
- The historical mainnet stall and stripped-witness source are still not reproduced; this fixes deterministic recovery and a demonstrable lock-order hazard.

<sup>Written for commit cbe128c7e7b68d3acfaec335c2b6a01c01469359. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gosuda/bitcoin-rs/pull/1103?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


metaphorics:
@codex review

PR #1115, opened 2026-09-21
fix(#627): thread submitblock wire bytes into the serialized apply path

### LINKED ISSUE #627: Parse each block once in Rust before changing the script backend
## Goal

Replace the production `KernelBlock` parsing dependency with one checked Rust block layout. Keep the script backend unchanged during this step so parser costs and consensus changes can be measured separately.

This is the parsing step of #166, not the native-validation promotion decision.

## Work

Start in [primitives](https://github.com/gosuda/bitcoin-rs/tree/fe2c132933c132d42da0be95074ae06924c5c14c/crates/primitives/src), `crates/consensus/src/block_view.rs`, and `crates/node/src/apply.rs`.

Keep immutable received bytes with checked transaction, input, output, and witness ranges. Use distinct types for byte ranges and metadata indexes. Reuse transaction IDs, sizes, and layout metadata in validation and apply. Keep an explicit adapter for the optional kernel oracle.

One structural parse does not mean that hashing or script execution may never reread bytes.

## Done when

- [ ] Native parsing matches the pinned reference for transaction boundaries, txid, wtxid, stripped size, total size, weight, and consumed bytes.
- [ ] Truncation, noncanonical lengths, overflow, impossible counts, and trailing bytes return defined errors without unchecked allocation or access.
- [ ] Merkle roots and mutation results match. The witness Merkle coinbase leaf is zero without changing the coinbase's actual wtxid.
- [ ] The normal apply path does not build a second transaction tree or serialize transactions merely to parse them again.
- [ ] Matched apply benchmarks meet the existing nonregression gate with the same script backend, corpus, and durability settings.

## Checks

Extend `kernel_block_parity`, primitive differential tests, and parser fuzzing. Include legacy and witness blocks, every supplied historical golden block, malformed lengths, and batch tails. Compare end-to-end apply time, CPU, allocations, and peak memory.

Out of scope: changing the default script verifier, replacing the project's primitive vocabulary, or changing storage formats.

Depends on #625 and the dependency contract in #179. Blueprint: #624, T05–T06. Related: #172, #166, #213.

## Diagram

One parse followed by ordered preparation and parallel verification

DPS0340:
Closes #627 (slice A: RPC/mining submission lane).

## Summary

`submitblock` now keeps the exact decoded wire image through the serialized apply path without adding a second parser or a parallel byte-validation owner.

- RPC decodes once, drops API-15 trailing bytes, and forwards only the consumed block image.
- Mining reuses the apply layer's exact-image gate, fills an omitted coinbase witness nonce when required, and re-serializes only when that fill changes the block.
- The existing submit body selects `connect_serialized` when preserved bytes are available.

## Validation

- `cargo test -p bitcoin-rs-node --no-default-features --features fjall --test mining`: 54 passed.
- `cargo test -p bitcoin-rs-rpc --lib`: 469 passed.
- `cargo clippy` with `-D warnings` is clean for `bitcoin-rs-mining`, `bitcoin-rs-node`, and `bitcoin-rs-rpc`.
- `cargo fmt --all -- --check` and `git diff --check`: clean.
- `python3 scripts/check_models.py --check-only`: unavailable locally because the pinned Apalache executable is not installed.

---

<a href="https://app.devin.ai/review/gosuda/bitcoin-rs/pull/1115" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>

---
## Summary by cubic
Threads `submitblock`'s exact wire bytes through the apply path so a submitted block is validated and applied from a single parse instead of being re-serialized and parsed a second time.

- `submitblock` submits the decoded hex via the new `MiningControl::submit_block_with_bytes`; the default implementation delegates to `submit_block`, so existing fakes keep working.
- The node checks the bytes encode the submitted block before the uncommitted-coinbase-witness fill, then routes them through `connect_serialized`. When the fill inserts the reserved nonce, the filled block is re-serialized and threaded instead.
- The byte check is a gate, not a count check: a witness-only mutation is refused even though every txid matches.
- API-15 trailing-byte tolerance is preserved: the accepted prefix is resolved through the checked layout and the tail dropped before the exact-image gate.
- Tests cover the serialized accept path, foreign wire bytes (with the gate's rejection reason pinned), trailing-tail dropping, and pre-fill images.

<sup>Written for commit abc14d400dfc4e0cf5a4aca30c6b7cd1d3e59d2d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gosuda/bitcoin-rs/pull/1115?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


metaphorics:
@codex review

DPS0340:
Follow-up addressing the two Devin review bugs (plus doc ownership):

- Trailing tail: the accepted prefix is now threaded into
  `connect_serialized`; the tail is dropped, not forwarded. Without this,
  the apply path's exact-image gate rejected valid `ffff`-suffixed
  submissions. New test `submit_block_with_bytes_drops_trailing_tail`.
- Pre-fill image: the byte check now runs before
  `fill_uncommitted_witness`. The old order always refused the valid
  pre-fill image on SegWit-active heights (fill changes the
  serialization). No-op fill threads the verified prefix into the
  serialized path; an applied fill re-serializes (same single serialize
  the decode-only path always paid). New test
  `submit_block_with_bytes_accepts_prefill_image_when_fill_applies`
  (real witness commitment, nonce cleared).
- Doc ownership: `submitted_prefix` is a free function with the
  API-15/exact-image rationale; the trait method documents its contract.

Validation: node mining suite 56 passed; rpc lib 470 passed; node clippy
`-D warnings` clean; fmt clean.

DPS0340:
Follow-up: the foreign-bytes refusal test now pins the gate reason (`Rejected("submitted bytes are not the serialization of the submitted block")`), so a downstream consensus refusal (`unexpected-witness`, etc.) fails the test instead of passing it. The test block carries no witness commitment, so the consensus layer would flag the mutated witness on its own — the exact-reason assertion isolates the bytes gate. Covers the cubic P2 review point.

metaphorics:
@codex review

